PR #77, opened 2017-12-11
aligned allocation simplified

JohanMabille:


wolfv:
Relevant benchmarks can be found in https://github.com/wolfv/xtensor-benchmark

PR #80, opened 2018-02-11
Docs: Spack Install

ax3l:
Adding [spack](https://spack.io/) to one of the install methods.

Spack can install various versions of xsimd alongside, taking the latest via
```bash
spack install xsimd
```

If you want a specific (older or development) version, run
```bash
spack install xsimd@3.1.0
```
or
```bash
spack install xsimd@develop
```

An installed package can then be activated via `spack load` with the same syntax:
```bash
spack load xsimd
```

The package management file is described in the few lines here:
https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/xsimd/package.py


ax3l:
Also adds svg's for conda and cmake as in xtl and xtensor docs :)

PR #81, opened 2018-03-12
neon support

wolfv:
Adding support for ARM NEON instruction set!


wolfv:
reminder to self to add the andnot from Eigen here.

JohanMabille:
@wolfv is that ready for a merge?

wolfv:
I can do feature detection only on two devices at the moment (arm64 and
armv7 vs armv8) maybe you should have a quick look if you like the feature
detection? I need to test on the older raspberry again, and then I think
yes, it's ready!

include/xsimd/config/xsimd_instruction_set.hpp
```diff
@@ -165,6 +165,9 @@
 
 #if !defined(XSIMD_ARM_INSTR_SET) && (defined(__ARM_NEON__) || defined(__aarch64__) || defined(_M_ARM))
     #define XSIMD_ARM_INSTR_SET XSIMD_ARM_NEON_VERSION
+    #if defined(__aarch64__)
+        #define XSIMD_ARM_64 1
+    #endif
```
JohanMabille: I think these macro could be mixed in a single one to keep conditional compilation based on instruction set simple. Even if `armv8` and `armv7` actually refer to different architectures (ie 32 bits vs 64 bits), they are not totally different instructions sets and it's unlikely that a new set comes with 32 bits mode support only. So the following should be ok:

```cpp
#define XSIMD_ARM7_NEON_VERSION XSIMD_VERSION_NUMBER(1, 0, 0)
#define XSIMD_ARM8_NEON_VERSION XSIMD_VERSION_NUMBER(2, 0, 0)

#if !defined(XSIMD_ARM_INSTR_SET) && (defined(__ARM_NEON__) || defined(__aarch64__) || defined(_M_ARM))
    #if defined(__aarch64__)
        #define XSIMD_ARM_INSTR_SET XSIMD_ARM8_NEON_VERSION 
    #else
        #define XSIMD_ARM_INSTR_SET XSIMD_ARM7_NEON_VERSION
    #endif
#endif
```

PR #82, opened 2018-03-18
added missing operator<<(batch, batch) and operator>>(batch, batch)

ukoethe:
These functions are called in the corresponding xexpressions when xtensor is compiled with `XTENSOR_USE_SIMD=ON`. Specifically, tests `left_shift` and `right_shift` in `test_xoperation.cpp` fail to compile. The PR is not pretty, but seems to be the simplest solution by far.


JohanMabille:
Hi,

Thanks for implementing this. Let's merge it for now, so it's not blocking for you.
I'll open an issue so we don't forget to reimplement it with specifics intrinsics when we have time to do it.

ukoethe:
> reimplement it with specifics intrinsics

I didn't find an intrinsic for this case. Is there any?

JohanMabille:
There are some for integers (`_mm_sll_epi64` / `_mm_srl_epi64` for instance), for floating point types we'll have to mix them with conversion / selection intrinsics.

PR #88, opened 2018-03-22
WIP Adding a shim type (batch<T, N>)

wolfv:
The shim type can be used for all batch sizes and types, enabling also a scalar version of our algorithms.

### DIFF
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -21,9 +21,10 @@ set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build." FORCE)
 include(CheckCXXCompilerFlag)
 
 string(TOUPPER "${CMAKE_BUILD_TYPE}" U_CMAKE_BUILD_TYPE)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Ofast -march=native -Wunused-parameter -Wextra -Wreorder -Wconversion")
 
 if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES "Intel")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=native -Wunused-parameter -Wextra -Wreorder -Wconversion")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Ofast -march=native -Wunused-parameter -Wextra -Wreorder -Wconversion")
     CHECK_CXX_COMPILER_FLAG("-std=c++11" HAS_CPP11_FLAG)
 
     if (HAS_CPP11_FLAG)
--- a/include/xsimd/math/xsimd_rounding.hpp
+++ b/include/xsimd/math/xsimd_rounding.hpp
@@ -157,10 +157,10 @@ namespace xsimd
         return s ^ select(v < t2n, d0 - t2n, v);
     }
 
-    template <>
-    inline batch<float, 4> trunc(const batch<float, 4>& x)
+    template <class T, std::size_t N>
+    inline batch<T, N> trunc(const batch<T, N>& x)
     {
-        using btype = batch<float, 4>;
+        using btype = batch<T, N>;
         return select(abs(x) < maxflint<btype>(), to_float(to_int(x)), x);
     }
 
--- a/include/xsimd/types/xsimd_shim.hpp
+++ b/include/xsimd/types/xsimd_shim.hpp
@@ -1,0 +1,749 @@
+/***************************************************************************
+* Copyright (c) 2016, Wolf Vollprecht, Johan Mabille and Sylvain Corlay    *
+*                                                                          *
+* Distributed under the terms of the BSD 3-Clause License.                 *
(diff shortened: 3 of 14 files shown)


wolfv:
@serge-sans-paille that's what i was talking about.

include/xsimd/types/xsimd_sse_int64.hpp
```diff
@@ -127,6 +127,8 @@ namespace xsimd
     batch_bool<int64_t, 2> operator!=(const batch<int64_t, 2>& lhs, const batch<int64_t, 2>& rhs);
     batch_bool<int64_t, 2> operator<(const batch<int64_t, 2>& lhs, const batch<int64_t, 2>& rhs);
     batch_bool<int64_t, 2> operator<=(const batch<int64_t, 2>& lhs, const batch<int64_t, 2>& rhs);
+    batch_bool<int64_t, 2> operator>(const batch<int64_t, 2>& lhs, const batch<int64_t, 2>& rhs);
+    batch_bool<int64_t, 2> operator>=(const batch<int64_t, 2>& lhs, const batch<int64_t, 2>& rhs);
```
JohanMabille: These are not needed, they are already defined for the base class.

include/xsimd/math/xsimd_rounding.hpp
```diff
@@ -157,10 +157,10 @@ namespace xsimd
         return s ^ select(v < t2n, d0 - t2n, v);
     }
 
-    template <>
-    inline batch<float, 4> trunc(const batch<float, 4>& x)
+    template <class T, std::size_t N>
+    inline batch<T, N> trunc(const batch<T, N>& x)
```
JohanMabille: Why this change? This function is the specialization of `trunc` for float, then the specialization for double is just below (if it exists).

include/xsimd/types/xsimd_shim.hpp
```diff
@@ -0,0 +1,749 @@
...
+    template <class T, std::size_t N>
+    struct simd_batch_traits<batch_bool<T, N>>
+    {
+        using value_type = typename detail::get_int_type<T>::unsigned_type;
+        static constexpr std::size_t size = N;
+        using batch_bool_type = batch_bool<T, N>;
+    };
+
+    template <class T, std::size_t N>
+    class batch_bool : public simd_batch_bool<batch_bool<T, N>>
```
JohanMabille: This should be a specialization of `batch_bool` for `N=1`, we should not provide generic implementaion of `batch_bool` for arbitrary `N`.

include/xsimd/types/xsimd_shim.hpp
```diff
@@ -0,0 +1,749 @@
...
+    template <class T, std::size_t N>
+    struct simd_batch_traits<batch<T, N>>
+    {
+        using value_type = T;
+        static constexpr std::size_t size = N;
+        using batch_bool_type = batch_bool<T, N>;
+    };
+
+    template <class T, std::size_t N>
+    class batch : public simd_batch<batch<T, N>>
```
JohanMabille: Same remark as for `batch_bool`.

include/xsimd/types/xsimd_shim.hpp
wolfv: I was thinking the following: first, we could easily test int64_t on NEON 32bit devices, where double is NOT available. 

Also, we can easily prevent some implementations for e.g. batch<uint8_t, 16> and hope (or actually measure!) that compilers are able to autovectorize the code (especially if we provide some hints, such as `__assume_aligned` etc. -- and then later we can swap out/specialize the implementations with the real stuff.

HadrienG2:
Having a scalar fallback looks very convenient! The only risk that I see is for people to accidentally activate it and not understand why their code is slow. Ideally, it should be possible to disable the scalar fallback using a preprocessor flag in order to check whether one's code uses the vector implementation or not.

HadrienG2:
Boost.SIMD also had a "vector of vectors" fallback, where if you, say, ask for a 512-bit batch<double, 8> on AVX (which is 256-bit), it emulates it as an array of two 256-bit vectors.

benchmark/CMakeLists.txt
```diff
@@ -21,9 +21,10 @@ set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build." FORCE)
 include(CheckCXXCompilerFlag)
 
 string(TOUPPER "${CMAKE_BUILD_TYPE}" U_CMAKE_BUILD_TYPE)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Ofast -march=native -Wunused-parameter -Wextra -Wreorder -Wconversion")
```
HadrienG2: You should not do this.

- It is redundant with the other directive below.
- It will likely break the build on compilers which do not use GCC's flag syntax (like MSVC).

serge-sans-paille: Use target_compile_options and generator to handle that properly

```c++
    target_compile_options(frozen.tests.html_entities_map PUBLIC
        $<$<CXX_COMPILER_ID:Clang>:     -fconstexpr-steps=123456789>
        $<$<CXX_COMPILER_ID:AppleClang>:-fconstexpr-steps=123456789>
    )
```

Oh, and ``-0fast`` is also called ``-Ocheating`` :-)

JohanMabille:
Superseded by #100

PR #94, opened 2018-04-25
Tentative fix for compiler warning

bluescarni:


bluescarni:
There's an unused variable in a function in ``xsimd_avx_int64.hpp``. I just removed the unused variable and renamed the other vars accordingly, but I have no idea if that's correct :)

There are also some ``static inline`` functions in ``xsimd_rem_pio2.hpp`` clang is complaining about, with messages such as:

```
/home/yardbird/.local/include/xsimd/math/xsimd_rem_pio2.hpp:449:29: warning: 'static' function '__ieee754_rem_pio2' declared in header file should be declared
      'static inline' [-Wunneeded-internal-declaration]
        static std::int32_t __ieee754_rem_pio2(double x, double* y)
```

I can apply the suggested change in the PR if it makes sense.

include/xsimd/types/xsimd_avx_int64.hpp
```diff
@@ -648,10 +648,9 @@ namespace xsimd
 #else
         XSIMD_SPLIT_AVX(rhs);
         __m128i tmp1 = _mm_shuffle_epi32(rhs_low, 0x0E);
-        __m128i tmp2 = _mm_add_epi64(tmp1, rhs_low);
-        __m128i tmp3 = _mm_shuffle_epi32(rhs_high, 0x0E);
-        __m128i tmp4 = _mm_add_epi64(tmp3, rhs_high);
-        __m128i res = _mm_add_epi64(tmp3, tmp4);
```
JohanMabille: I think the problem is the last line, the fix should be

```cpp
__m128i res = _mm_add_epi64(tmp2, tmp4);
```

JohanMabille:
Thanks for catching this! See my review for the fix of the unused variable, about the warnings regarding the static functions, I think we can drop the `static` specifier and use `inline` only.

bluescarni:
No problem, thanks for the great library. I updated the PR.

PR #101, opened 2018-05-03
Add AVX 512 support

wolfv:
very very initial commit for AVX 512 support.


wolfv:
@JohanMabille It seems like AVX doesn't have the same comparison operations as AVX2 anymore. All of them return a mask (integer) instead of a vector type.

The question is: Should we implement the `batch_bool` for AVX512 types with an internal value of a `__mask8/16/32`? That would remove a couple of conversions, but I am unsure if we have assumptions on this in the code base.

JohanMabille:
Except if we try to instantiate a `batch` from a `batch_bool`, the code base should be independant from the internal storage type.

So yes we should implement `batch_bool` with `__mask8/16/32`, and then add constructor / conversion operator from `__m512` in the `batch_bool` implementation if that's required.

HadrienG2:
The fallback implementation certainly does not use the same storage type for batch and batch_bool :)

wolfv:
This is how the BLIS folks test different intel archs on travis with the SDE: https://github.com/flame/blis/commit/d7f531b6999c30b721e4ad47859583779a1fb657

bluescarni:
Hello, as mentioned in #116 I switched my code to use the AVX512 implementation from this branch of xsimd and found no performance differences with my own (limited) implementation. So it's a :+1: from me :)

PR #104, opened 2018-05-04
reducers added to the doc

### LINKED ISSUE #95: Could not find batch reducers on ReadTheDocs
Maybe that's because the search box is currently broken. But I did not find documentation for hadd and other batch reducers on [ReadTheDocs](https://xsimd.readthedocs.io/en/latest/api/xsimd_batch.html), even though I expected it to be generated from [this Doxygen](https://github.com/QuantStack/xsimd/blob/2f2e8dff874776a58c4621e894895fa3349db33b/docs/source/api/xsimd_batch.hpp#L533).

JohanMabille:
Fixes #95

### DIFF
--- a/docs/source/api/xsimd_batch.rst
+++ b/docs/source/api/xsimd_batch.rst
@@ -34,6 +34,13 @@ Logical operators
    :project: xsimd
    :content-only:
 
+Reducers
+--------
+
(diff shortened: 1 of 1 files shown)


JohanMabille:
Only documentation id changed, can be merged.

PR #109, opened 2018-05-07
WIP shuffling interleave etc

wolfv:
@HadrienG2 I opened this PR because I thought you might be interested.

This adds the select stuff we've been talking about and also an initial interleave.
I really like the explicitness of the interleave with the AbCc... syntax, although it might be quite a lot of typing for some.

### DIFF
--- a/include/xsimd/types/xsimd_avx_shuffling.hpp
+++ b/include/xsimd/types/xsimd_avx_shuffling.hpp
@@ -75,8 +75,8 @@ namespace xsimd
             std::pair<batch<T, N>, batch<T, N>> operator()(const batch<T, N>& a, const batch<T, N>& b)
             {
                 // {a, b, c, d} Δ {A, B, C, D} → std::pair{{a, A, b, B}, {c, C, d, D}}
-                auto lo = interleave_impl<0b0101, 64, int64_t, 4>{}(bitwise_cast<batch<int64_t, 4>>(a), bitwise_cast<batch<int64_t, 4>>(b));
-                auto hi = interleave_impl<0b1010, 64, int64_t, 4>{}(bitwise_cast<batch<int64_t, 4>>(a), bitwise_cast<batch<int64_t, 4>>(b));
+                auto lo = interleave_impl<AaCc, 64, int64_t, 4>{}(bitwise_cast<batch<int64_t, 4>>(a), bitwise_cast<batch<int64_t, 4>>(b));
+                auto hi = interleave_impl<BbDd, 64, int64_t, 4>{}(bitwise_cast<batch<int64_t, 4>>(a), bitwise_cast<batch<int64_t, 4>>(b));
                 batch<int64_t, 4> tmp_up = _mm256_permute2f128_si256(lo, hi, _MM_SHUFFLE(0, 2, 0, 0));
                 batch<int64_t, 4> tmp_bt = _mm256_permute2f128_si256(lo, hi, _MM_SHUFFLE(0, 3, 0, 1));
                 return {
--- a/include/xsimd/types/xsimd_avx_shuffling.hpp
+++ b/include/xsimd/types/xsimd_avx_shuffling.hpp
@@ -1,0 +1,148 @@
+/***************************************************************************
+* Copyright (c) 2016, Johan Mabille and Sylvain Corlay                     *
+*                                                                          *
+* Distributed under the terms of the BSD 3-Clause License.                 *
+*                                                                          *
+* The full license is in the file LICENSE, distributed with this software. *
+****************************************************************************/
+
+#ifndef XSIMD_AVX_SHUFFLING_HPP
+#define XSIMD_AVX_SHUFFLING_HPP
+
+#include "xsimd/types/xsimd_avx_conversion.hpp"
+
+namespace xsimd 
+{
+    // Version of select(a, b) with compile time mask
+    template <class B>
+    struct selector_immediate;
+
+    template <class T, std::size_t N>
+    struct selector_immediate<batch<T, N>>
+    {
+        using type = int32_t;
+    };
+
+    template <typename selector_immediate<batch<float, 8>>::type P, class T>
+    batch<T, 8> select(const batch<T, 8>& a, const batch<T, 8>& b);
+
+    template <typename selector_immediate<batch<double, 4>>::type P, class T>
+    batch<T, 4> select(const batch<T, 4>& a, const batch<T, 4>& b);
+
+    constexpr static int AaBbCcDd = 1;
+    constexpr static int AaCc = 2;
+    constexpr static int BbDd = 3;
+    constexpr static int AaBbEeFf = 4;
+    constexpr static int ABabEFef = 5;
+    constexpr static int CcDdGgHh = 6;
+    constexpr static int CDcdGHgh = 7;
+
+    template <uint64_t P, int B = -1, class T, size_t N>
+    auto interleave(const batch<T, N>& a, const batch<T, N>& b);
+
+    namespace detail
+    {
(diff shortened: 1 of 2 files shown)


include/xsimd/types/xsimd_avx_shuffling.hpp
```diff
@@ -0,0 +1,148 @@
...
+    template <typename selector_immediate<batch<double, 4>>::type P, class T>
+    batch<T, 4> select(const batch<T, 4>& a, const batch<T, 4>& b);
+
+    constexpr static int AaBbCcDd = 1;
+    constexpr static int AaCc = 2;
+    constexpr static int BbDd = 3;
+    constexpr static int AaBbEeFf = 4;
+    constexpr static int ABabEFef = 5;
+    constexpr static int CcDdGgHh = 6;
+    constexpr static int CDcdGHgh = 7;
```
HadrienG2: If these possibilities are mutually exclusive, I would make this an enum (class).

From a high-level design point of view, however, I am a bit uneasy with spelling out all the possibilities like this, due to the combinatorics involved. Could it scale all the way to AVX-512's vectors? To possible future extensions to smaller data types (Intel's SIMD instruction set supports 16-bit ints, the theoretical minimum is 8-bit)? To future wider instructions sets like ARM SVE?

Maybe it would be better to wait for precise use cases, and until then only implement obvious stuff like ({a, b, c, d}, {A, B, C, D}) -> ({a, A, b, B}, {c, C, d, D}). Studying the various SIMD instruction sets to see which interleavings are commonly supported could also be a good idea.

include/xsimd/types/xsimd_avx_shuffling.hpp
```diff
@@ -0,0 +1,148 @@
...
+
+        template <class T, size_t N>
+        struct interleave_impl<AaCc, 64, T, N>
+        {
+            batch<T, N> operator()(const batch<T, N>& a, const batch<T, N>& b)
+            {
+                // {a, b, c, d} Δ {A, B, C, D} → {a, A, c, C}
+                batch<int64_t, 4> tmp = _mm256_unpacklo_epi64(bitwise_cast<batch<int64_t, 4>>(a),
+                                                              bitwise_cast<batch<int64_t, 4>>(b));
+                return bitwise_cast<batch<T, N>>(tmp);
```
HadrienG2: My reading of the unpacklo/unpackhi spec is that they actually do {a, A, b, B} and {c, C, d, D} respectively

include/xsimd/types/xsimd_avx_shuffling.hpp
```diff
@@ -0,0 +1,148 @@
...
+            {
+                // {a, b, c, d} Δ {A, B, C, D} → std::pair{{a, A, b, B}, {c, C, d, D}}
+                auto lo = interleave_impl<AaCc, 64, int64_t, 4>{}(bitwise_cast<batch<int64_t, 4>>(a), bitwise_cast<batch<int64_t, 4>>(b));
+                auto hi = interleave_impl<BbDd, 64, int64_t, 4>{}(bitwise_cast<batch<int64_t, 4>>(a), bitwise_cast<batch<int64_t, 4>>(b));
+                batch<int64_t, 4> tmp_up = _mm256_permute2f128_si256(lo, hi, _MM_SHUFFLE(0, 2, 0, 0));
+                batch<int64_t, 4> tmp_bt = _mm256_permute2f128_si256(lo, hi, _MM_SHUFFLE(0, 3, 0, 1));
+                return {
+                    bitwise_cast<batch<T, N>>(tmp_up),
+                    bitwise_cast<batch<T, N>>(tmp_bt)
+                };
```
HadrienG2: ...therefore, this can be implemented using just unpackhi and unpacklo.

include/xsimd/types/xsimd_avx_shuffling.hpp
wolfv: unpacklo/hi on mm256 work on two different mm128 vectors and do the same transformation on each. therefore unpacklo does aA on mm128 1 and cC in the second.

include/xsimd/types/xsimd_avx_shuffling.hpp
wolfv: yep, i already made it an enum in my version, haven't pushed yet! 

yes, i also agree that it might get cluttered .. therefore I was thinking maybe we can use this as sort of "patterns" that are to be used in combination with the bitsize of the interleave.

so e.g. doing an AaCc shuffle on a 8 bit width would result in 

AaBbCcDdEeFfGg_PpQqRrSsTtUuVvWw (I hope I spelled it out correctly).

The other option would be to find meaningful function names (e.g. in MIPP they have 

https://github.com/aff3ct/MIPP/blob/f0aabd427b9557124e7c8d384f4383a8c4f1f5cf/src/mipp_impl_AVX.hxx

- interleavelo2
- interleavehi2
- interleave
- interleavehi
- interleavelo

of all the different types.

include/xsimd/types/xsimd_avx_shuffling.hpp
HadrienG2: Oh yes, you're right, it's actually even weirder than I thought. If I take _mm256_unpacklo_ps, for example, it is defined as...

```
INTERLEAVE_DWORDS(src1[127:0], src2[127:0]){
  dst[31:0] := src1[31:0]
  dst[63:32] := src2[31:0]
  dst[95:64] := src1[63:32]
  dst[127:96] := src2[63:32]
  RETURN dst[127:0]
}

dst[127:0] := INTERLEAVE_DWORDS(a[127:0], b[127:0])
dst[255:128] := INTERLEAVE_DWORDS(a[255:128], b[255:128])
dst[MAX:256] := 0
```

Which means that if you take as input {a, b, c, d, e, f, g, h} and {A, B, C, D, E, F, G, H}, you will do...

```
dst[127:0] := INTERLEAVE_DWORDS({a, b, c, d}, {A, B, C, D}) == {a, A, b, B}
dst[255:128] := INTERLEAVE_DWORDS({e, f, g, h}, {E, F, G, H}) == {e, E, f, F}
```

So the end result is indeed ({a, A, b, B, e, E, f, F}).

HadrienG2: I think I'm starting to understand why the Boost.SIMD devs just threw the towel and said "okay, just give us an arbitrary compile-time permutation and we'll compile it down into whatever the host CPU supports" :)

wolfv: What would be cool is to write a program with some optimizer that can find the optimal permutation sequence -- it could be written in python and use the cost of the instruction docs from agner fog :) (In case you're looking for a research project, haha).

wolfv: There is some prior art on this: https://users.ece.cmu.edu/~franzf/papers/cc08.pdf

wolfv:
More inspiration: https://github.com/g-truc/glm/blob/faf1da52b7ab8fbd4ce770cd56c9f84893d7f233/glm/detail/_swizzle_func.hpp

include/xsimd/types/xsimd_avx_shuffling.hpp
otristan: I'm trying to move some code from Boost.SIMD to xsimd and so far this is the only kind of stuff that I a missing.
trying to shuffle<0, 2, 4, 6, 1, 3, 5, 7> a batch<int16_t, 8>

Would appreciate a split_low and split_high as well :)

Thanks !

JohanMabille:
shuffling is available in xsimd 8.1.0, the function to use is swizzle.

PR #118, opened 2018-06-04
fix bool for AVX and SSE

wolfv:
... and activate tests.


wolfv:
The other option would be to use `!(lhs ^ rhs)` but from checking the docs it seems slower.

PR #127, opened 2018-06-20
fix for Intel on Windows

### LINKED ISSUE #125: Can not build with Intel Compiler on MSVC because cmake fails
Currently the build fails for me with the latest Intel Compiler on MSVC (and Visual Studio 2015). The build aborts in cmake configuration with error
```
-- Performing Test HAS_CPP11_FLAG
-- Performing Test HAS_CPP11_FLAG - Failed
CMake Error at test/CMakeLists.txt:53 (message):
  Unsupported compiler -- xsimd requires C++11 support!

-- Configuring incomplete, errors occurred!
```
I assume the detection is not perfect, because the Intel Compiler has support for `C++11`. Is there anything I can do to help?

JohanMabille:
This prevent the check of c++11 flags for Intel compiler on Windows and should fix #125, however I'm not sure how to set the correct flags for activating SSE / AVX for this compiler on Windows.


emmenlau:
Dear @JohanMabille , I'm also no expert in this, but my understanding is that the `MSVC` token will not be defined. Therefore its my assumption that your PR is the best that can be done to separate Intel Compiler on Windows from other platforms.
If someone has more insight, please share!

PR #133, opened 2018-06-28
char types

wolfv:


wolfv:
@JohanMabille I got it more or less to work now. 
We still need to think about what to do with AVX512. I would like to make the default batch_type for AVX512<int8> the same as AVX2 since char support for AVX512 does not yet have great support among the CPUs

wolfv:
@JohanMabille there is a fix for char -> float loading in here too (there was a missing return statement).

Just to remind ourselves to either backport it or merge this PR beforehand.

include/xsimd/types/xsimd_avx_float.hpp
```diff
@@ -353,7 +353,7 @@ namespace xsimd
 
     inline batch<float, 8>& batch<float, 8>::load_unaligned(const char* src)
     {
-        load_aligned(src);
+        return load_aligned(src);
```
wolfv: @JohanMabille

PR #146, opened 2018-07-08
increase NEON float accuracy for sqrt

wolfv:
This PR increases the approximation accuracy for calculating the square root of a floating point number on pre-ARMv8 hardware. 

It helps to alleviate accuracy problems we've observed in the complex implementation. 

The remaining failing test is `complex tan`.


JohanMabille:
clang 3.7 repeatedly fails because of `XOP not defined` on the ci machines, so let's ignore it for now, it will be remove from the ci in a dedicated PR.

SylvainCorlay:
Kicking the build often fixes that issue. I don't know why.

JohanMabille:
Indeed!

PR #157, opened 2018-08-13
Fix missing semicolons.

bluescarni:
I was mucking around with compiler flags on an AVX512 machine and spotted what look like a couple of cases of missing semicolons on class template specialisations.


wolfv:
Thanks for uncovering these!

SylvainCorlay:
Johan?

wolfv:
we're going to make a release today. 

Btw, if you have other feedback / etc. feel free to drop by the chat: https://gitter.im/QuantStack/Lobby

PR #159, opened 2018-08-15
Fixup sse batches

SylvainCorlay:


JohanMabille:
I think the base type is right, the problem comes from the type returned by `load_aligned` / `load_unaligned` methods, they should return the final type (i.e.batch<T, N>).

SylvainCorlay:
This is ready to go!

SylvainCorlay:
@wolfv this is ready to go!

wolfv:
reviewing now

PR #164, opened 2018-08-22
Changed default cmake behavior to not build the tests

tdegeus:
This facilitates the `make install` feature. If the tests are build one has to worry about `gtest` which could be a nuisance if one just wants to install the header-only library.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -24,7 +24,7 @@ install:
   - conda info -a
   - conda install gtest cmake -c conda-forge
   - conda install xtl -c quantstack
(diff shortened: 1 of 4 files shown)


wolfv:
that will need a change in `.travis.yml` and `.appveyor.yml` too:

https://github.com/QuantStack/xsimd/blob/master/.appveyor.yml#L27


and all the lines with cmake here:

https://github.com/QuantStack/xsimd/blob/master/.travis.yml#L141-L157

tdegeus:
Ok.

But do you agree with the premise?

wolfv:
Yep we all agree!

tdegeus:
Then I'll make the necessary changes.

PR #174, opened 2018-09-11
Fix copysign implementation on windows

serge-sans-paille:


serge-sans-paille:
See https://ci.appveyor.com/project/serge-sans-paille/pythran/build/1.0.839 for the fail that spotted this « error »

serge-sans-paille:
> Does it work if you #undef copysign before including xsimd?

Yeah, that looks like a #define from the ms header. I'm not #undef is an
option then, because it would propagagte to all files including xsimd.

SylvainCorlay:
To prevent the macro expansion, would it work to put `copysign` in parentheses like we do for `max` and `min`. `(copysign)(x, y)`.

serge-sans-paille:
the expansion is an issue in the original ``using std::copysign``, I don't think your trick would work. The travis build only needs to be restarted!

SylvainCorlay:
I would think that the parentheses should be added around the function call. (In the macro for math functions).

serge-sans-paille:
Which function call are you considering? The original statement is just

    using std::copysign;

serge-sans-paille:
Culprint found: https://github.com/python/cpython/blob/master/PC/pyconfig.h#L193
So definitievly *not* an xsimd related error.

PR #175, opened 2018-09-12
add more int tests

wolfv:
There was a bug with the lower-than operator for int8 & int16 types.
And removes usage of _mm256_xor_si256 from `AVX` (note the missing `2`) which did not compile /cc @serge-sans-paille 

additionally, this adds some more tests (also for the constructors from multiple values, incl. min/max of the int type).


wolfv:
hah, ok! found some test failures...

Issue #178, opened 2018-09-21
Macro definition __LITTLE_ENDIAN breaks the build


jpcima:
The source file `xsimd_rem_pio2.hpp` defines the macro `__LITTLE_ENDIAN` under some conditions.

If such a condition is met on Linux systems, as I was told it could happen under Debian package building environment, the macro can redefine the system's definition in `endian.h`.

This can result in a failure in a following conditional such as this:
`#       if defined(__LITTLE_ENDIAN) && (__BYTE_ORDER == __LITTLE_ENDIAN)`

In case of using the Gcc or Clang compiler, maybe xsimd can use the builtin macros of these compilers instead?

JohanMabille:
Fixed in #179

PR #179, opened 2018-09-21
Fix about endianness detection, support more platforms

jpcima:
#178 

renaming `__LITTLE_ENDIAN` and attempting to detect on common platforms and compilers: windows, gcc, clang.
By no means I expect this detection to be exhaustive, but I hope it covers a good deal.
Little endianness would not be detected on Linux OS prior.


JohanMabille:
Thanks for tackling this! I think we don't want to bring in the headers from windows, so maybe we can rely on the default behavior for that (and that would fix the PRs).

PR #181, opened 2018-09-28
WIP

serge-sans-paille:
This is a work in progress version, motivated by its usage in Pythran. The performance impact is signficant: 10% faster than the former boost.simd (scalar mode)

Pro: faster scalar version for trigo
Contra: enables fallbacks by default (compilation time impact ?)

I'm unsure wether it makes sense to generalize this to all scalar functions.

For the record: the default scalar version of FMA uses std::fma which givs very accurta result but is slower than the default implementation. boost.simd used to rely on the dummy version too. This is the key to unleash performance on that case.

### DIFF
--- a/include/xsimd/math/xsimd_scalar.hpp
+++ b/include/xsimd/math/xsimd_scalar.hpp
@@ -8,6 +8,7 @@
 
 #ifndef XSIMD_SCALAR_HPP
 #define XSIMD_SCALAR_HPP
+#include <xsimd/types/xsimd_fallback.hpp>
 
 #include <cmath>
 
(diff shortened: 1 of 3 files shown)


HadrienG2:
Note that the performance with respect to boost.simd comes at the cost of losing IEEE-754 compliance. In that sense, that's not a very fair comparison.

Maybe we should provide some kind of "fast" API which is speed-optimized in addition to the "regular" API which is IEEE-compliant?

serge-sans-paille:
+1 @HadrienG2 better stick to the STL

PR #182, opened 2018-09-28
Fix scalar implementation of sincos on OSX

serge-sans-paille:


serge-sans-paille:
See https://github.com/serge-sans-paille/pythran/issues/1026

PR #206, opened 2018-10-10
Neon uint16

martinRenou:


include/xsimd/types/xsimd_neon_uint16.hpp
```diff
@@ -0,0 +1,390 @@
...
+
+    inline uint16_t batch<uint16_t, 8>::operator[](std::size_t index) const
+    {
+        return m_value[index];
+    }
+
+    namespace detail
+    {
+
+#define XSIMD_INLINE_UNROLL_OP(op, i)  \
```
wolfv: are weable to use the template unroller here?

include/xsimd/types/xsimd_neon_uint16.hpp
```diff
@@ -0,0 +1,390 @@
...
+            {
+                return uint16x8_t{
+                    XSIMD_INLINE_UNROLL_OP_8(/)
+                };
+            }
+
+            static batch_type mod(const batch_type& lhs, const batch_type& rhs)
+            {
+                return uint16x8_t{
+                    XSIMD_INLINE_UNROLL_OP_8(%)
```
wolfv: maybe not! :)

martinRenou: It is a bit confusing, but `XSIMD_INLINE_UNROLL_OP_8` is for 8 elements, not for int8. And we have 8 elements of uint16 in the batch here.

martinRenou: Or maybe I don't understand the problem you're seeing?

martinRenou: Maybe we should rename the macro `XSIMD_INLINE_UNROLL_OP_UINT16x8` so that it's not confusing anymore

wolfv: Sorry, I wasn't clear. I was referring to my statement above regarding usage of the meta-unroller. 
But I think we can't use it in a initializer, so a MACRO is necessary.

martinRenou: I made a template unroller like you suggested :) better than the macro

wolfv: looks great!

PR #208, opened 2018-10-11
WIP: Neon uint32

martinRenou:


martinRenou:
Re-opening later, need to push some tests and I don't want to trigger builds on CI that will fail anyway

PR #224, opened 2018-12-08
Support the almighty unary plus operator

serge-sans-paille:


wolfv:
Looks good to me!

Only question one could ask if we should support the C++ behavior with up-casting to at least 32bit integers, but we don't do it for other ops so it doesn't seem to make sense here.

PR #225, opened 2018-12-11
Correction of batch_bool<floating point type,...>::operator[]

hleclerc:
Hi there,

Many thanks for your fantastic library !

There's a small correction for `batch_bool<T,...>::operator[]` where T is a float or a double: sse and avx operations that give booleans (e.g. for `operator<`) actually fill all the bits in the results. It gives a `-nan` for a true or a 0 for a false if read as a double or a float. Unfortunately, these floating points values do not convert directly to the right boolean :(

In the PR, there are some additionnal tests for this topic.

Best regards.


hleclerc:
Ooops... I've missed it, many thanks :) !

SylvainCorlay:
Thanks for adding some tests as well!

I would close the PR and open one from a feature branch, so that you don't pollute your local master branch with commits that are not upstream.

PR #227, opened 2018-12-11
Eliminate unnecessary semicolons in macros.

mainland:
With this patch, -Wpedantic no longer produces warnings.


JohanMabille:
Thanks for this work!

PR #233, opened 2018-12-13
add mandelbrot example from tsimd

wolfv:
I've got a Cmake for this. 

should I add it so that it adds a new target for the default cmake "examples" or seperate in the examples dir?


JohanMabille:
I think a separate cmake in the `example dir` is the way to go.

PR #234, opened 2018-12-14
correction for batch_bool<T,.>::operator[] for avx/sse double/float

hleclerc:
Hi there,

Many thanks for your fantastic library !

There's a small correction for `batch_bool<T,...>::operator[]` where `T` is a `float` or a `double`: sse and avx operations that give booleans (e.g. for `operator<`) actually fill all the bits in the results. If direclty read as a `double` or a `float`, it gives a `-nan` for a `true` or a `0` for a `false`, and unfortunately, these floating points values do not convert directly to the right boolean values :(

In the PR, there are some additionnal tests for this topic.

Best regards.

### DIFF
--- a/include/xsimd/types/xsimd_avx_double.hpp
+++ b/include/xsimd/types/xsimd_avx_double.hpp
@@ -132,9 +132,7 @@ namespace xsimd
 
     inline bool batch_bool<double, 4>::operator[](std::size_t index) const
     {
-        alignas(32) double x[4];
-        _mm256_store_pd(x, m_value);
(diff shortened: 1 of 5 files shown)


JohanMabille:
Thanks for your contribution! This looks good to me, however the mail in the commit is not the same as the one of your Github account. Do you want to edit the commit so you get the credits, or are you ok to get it merged as is?

hleclerc:
Oops again, I've changed the email in my profile, it should be good now :) Many thanks for this verification !

PR #236, opened 2018-12-14
Correction/rcast breaks aliasing

hleclerc:
Hi there,

There's a correction (again, sorry) for `batch_bool<T,.>::operator[]` for avx/sse double/float: a direct `reinterpret_cast<&...>` may break the implicit aliasing rules of the modern optimizing compilers. For instance, with g++, `batch_bool<T,.>::operator[]` works now with `-ffast-math` and `-O6` (it was actually not always the case before).

Best regards.

### DIFF
--- a/include/xsimd/types/xsimd_avx_double.hpp
+++ b/include/xsimd/types/xsimd_avx_double.hpp
@@ -132,7 +132,7 @@ namespace xsimd
 
     inline bool batch_bool<double, 4>::operator[](std::size_t index) const
     {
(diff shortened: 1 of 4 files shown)


SylvainCorlay:
Thanks! Quick note: there is again a mismatch between your git configuration and github.

hleclerc:
Arggh... right. I'll change it for a next PR !

SylvainCorlay:
Thanks! I don't want to be nitoicky. It is just for you to be properly credited for your work!

PR #237, opened 2018-12-15
Potential aliasing issue

hleclerc:
Hi there,

There's a correction (again, sorry) for batch_bool<T,.>::operator[] for avx/sse double/float: a direct reinterpret_cast<&...> may break the implicit aliasing rules of the modern optimizing compilers. For instance, with g++, batch_bool<T,.>::operator[] works now with -ffast-math and -O6 (it was actually not always the case before). Works also with msvc.

Best regards.

### DIFF
--- a/include/xsimd/types/xsimd_avx_double.hpp
+++ b/include/xsimd/types/xsimd_avx_double.hpp
@@ -132,7 +132,9 @@ namespace xsimd
 
     inline bool batch_bool<double, 4>::operator[](std::size_t index) const
     {
-        return reinterpret_cast<const std::uint64_t*>(&m_value)[index & 3];
+        double v = reinterpret_cast<const double*>(&m_value)[index & 3];
+        std::uint64_t r = reinterpret_cast<std::uint64_t&>(v);
+        return (bool)r;
     }
 
     namespace detail
--- a/include/xsimd/types/xsimd_avx_float.hpp
+++ b/include/xsimd/types/xsimd_avx_float.hpp
@@ -136,7 +136,9 @@ namespace xsimd
 
     inline bool batch_bool<float, 8>::operator[](std::size_t index) const
     {
-        return reinterpret_cast<const std::uint32_t*>(&m_value)[index & 7];
+        float v = reinterpret_cast<const float*>(&m_value)[index & 7];
+        std::uint32_t r = reinterpret_cast<std::uint32_t&>(v);
+        return (bool)r;
     }
 
     namespace detail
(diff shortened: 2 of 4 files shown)


wolfv:
Hi Hugo, 

There is another interesting thing that we thought about copying from tsimd, a similar package, which is to define the batch storage as a union type of 

```
union storage_t {
    std::array<double, 4> arr;
    __m256d v;
};
```

etc. 

This would allow us to just interpret the register as std::array of x and extract the element as arr[x].
As well as forwarding the begin / end iterators of the array.

Do you think such a change would work well with GCC/MSVC?

hleclerc:
Hi Wolf,

I'm not sure to be the perfect guy to answer this question (not sure to be an expert on all the compiler technologies, especially when it comes to Microsoft ones :) ). Nevertheless, spontaneously, I would say that it seems to be a very good idea: enabling  a clearer code while preserving all the optimizations...
To my knowledge, on PODs, unions and reinterpret_casts basically give the same results, but with a more "honest" and "no-surprise" spirit for the unions -- if I've not missed something.
Do you want me to do something on this way ?

wolfv:
I checked the code that get's generated and I am afraid it doesn't look too good for a `load` function on GCC:

https://godbolt.org/z/iKDwJ2

Maybe reinterpret_cast or what you're doing is the better option in the end.

hleclerc:
Hi Wolf,

This is a very interesting discussion !

It seems that the problem with the `load` function comes essentially from the calling conventions: gcc apparently considers that the union type is too "complex" to be passed via AVX registers, even if it's POD and it fits inside :( The consequence is the use of the stack + the need to cleanup, etc...

However, clang seems to consider the contrary: in our case clang returns the union directly in a ymm register, not in the stack ! Arf...

There are good references on calling conventions (as e.g. [agner/calling_conventions](https://www.agner.org/optimize/calling_conventions.pdf), p. 22), but unless I've missed something, it seems that gcc does not follow the rules detailed in the tables I've read. It's a bit unseating.

As can be seen e.g. in https://godbolt.org/z/LStndE, we could can say that once inlined, the calling convention problem vanishes (the `f_xxx` functions contains the same assembly code)... But, well, in a reasonable code, I guess that all the functions are not inlined o)

JohanMabille:
I merged this so the master branch does not have this issue anymore, however it could nice to implement the union solution. If the `load` / `store` methods are not inline, you'll probably have poor performances whatever the generated code is, so we can reasonably think that these functions will be inline.

wolfv:
Thanks for digging deeper :) that's quite curious + good to see that GCC does optimize correctly in the inlined version.

yeah @hleclerc if you want to implement the union idea that would be greatly appreciated -- even if only a prototype for now. 

we've had community members ask for iterators *on* batches and this would be trivial if we implement the union idea.

PR #238, opened 2018-12-15
Fix lax vector conversion and GCC/Clang multiply bug.

easyaspi314:
Notes are in the commit messages.

GCC no longer needs `-flax-vector-conversion`.  I fixed the implicit casts.

In addition, I used intrinsics in `mul` for uint64 on NEON 32. The issue here is that both Clang and GCC will incorrectly switch to scalar multiply, which is slower than the pure NEON alternative on ARMv7a. 

I would like someone to benchmark the modified `mul` code on an aarch64 device. I have it disabled because I assume it would be slower like the same approach on x86_64, but I could be wrong. 

I commented the equivalent C code in the function, as the intrinsics can be a little hard to follow. The intrinsics are required, though, as it is apparently impossible without them.

### DIFF
--- a/include/xsimd/types/xsimd_neon_int16.hpp
+++ b/include/xsimd/types/xsimd_neon_int16.hpp
@@ -162,7 +162,7 @@ namespace xsimd
 
     inline batch<int16_t, 8>& batch<int16_t, 8>::load_aligned(const uint16_t* src)
     {
-        m_value = vld1q_u16(src);
+        m_value = vreinterpretq_s16_u16(vld1q_u16(src));
         return *this;
     }
 
@@ -183,7 +183,7 @@ namespace xsimd
 
     inline void batch<int16_t, 8>::store_aligned(uint16_t* dst) const
     {
-        vst1q_u16(dst, m_value);
+        vst1q_u16(dst, vreinterpretq_u16_s16(m_value));
     }
 
     inline void batch<int16_t, 8>::store_unaligned(uint16_t* dst) const
@@ -387,7 +387,7 @@ namespace xsimd
 
     inline batch<int16_t, 8> operator<<(const batch<int16_t, 8>& lhs, const batch<int16_t, 8>& rhs)
     {
-        return vshlq_s8(lhs, rhs);
+        return vshlq_s16(lhs, rhs);
     }
 }
 
--- a/include/xsimd/types/xsimd_neon_int8.hpp
+++ b/include/xsimd/types/xsimd_neon_int8.hpp
@@ -161,7 +161,7 @@ namespace xsimd
 
     inline batch<int8_t, 16>& batch<int8_t, 16>::load_aligned(const uint8_t* src)
     {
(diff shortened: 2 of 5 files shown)


wolfv:
Thanks a lot! This is great :)

We could check the generated assembly for the `mul` on aarch64 using godbolt!

I came up with this for ARMv8-a 

https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(j:1,lang:c%2B%2B,source:'%23include+%22xsimd/xsimd.hpp%22%0A%0A%23include+%3Cvector%3E%0A%23include+%3Ccstdint%3E%0A%0Auint64x2_t+good(uint64x2_t+lhs,+uint64x2_t+rhs)+%7B%0A++++//+return+uint64x2_t%7B+lhs%5B0%5D+*+rhs%5B0%5D,+lhs%5B1%5D+*+rhs%5B1%5D+%7D%3B%0A++++return+lhs+*+rhs%3B%0A%7D%0A%0Aauto+bad(uint64x2_t+lhs,+uint64x2_t+rhs)+%7B%0A++++uint32x2_t+lhs_lo+%3D+vmovn_u64+++(lhs)%3B%0A++++uint32x2_t+lhs_hi+%3D+vshrn_n_u64+(lhs,++++32)%3B%0A++++uint32x2_t+rhs_lo+%3D+vmovn_u64+++(rhs)%3B%0A++++uint32x2_t+rhs_hi+%3D+vshrn_n_u64+(rhs,++++32)%3B%0A++++uint64x2_t+result+%3D+vmull_u32+++(lhs_hi,+rhs_lo)%3B%0A+++++++++++++++result+%3D+vmlal_u32+++(result,+lhs_lo,+rhs_hi)%3B%0A+++++++++++++++result+%3D+vshlq_n_u64+(result,+32)%3B%0A+++++++++++++++result+%3D+vmlal_u32+++(result,+lhs_lo,+rhs_lo)%3B%0A++++return+result%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50.04268081836052,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:arm64g820,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',trim:'1'),lang:c%2B%2B,libs:!((name:xsimd,ver:trunk),(name:fmt,ver:trunk)),options:'-std%3Dc%2B%2B17+-O3+-march%3Darmv8-a+-flax-vector-conversions',source:1),l:'5',n:'0',o:'ARM64+gcc+8.2+(Editor+%231,+Compiler+%231)+C%2B%2B',t:'0')),k:49.95731918163949,l:'4',m:100,n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4

wolfv:
btw i think there is no native multipliation for uint64x2_t on arm 64 either

PR #240, opened 2018-12-21
union { __mxxxd reg; array<int_type,N> arr; } for batch_bool, AVX/SSE

hleclerc:
Hi there,

As discussed in a previous pull request, there's a proposition where for AVX and SSE, `batch_bool<floating point,N>` attributes become unions. These unions contain either a SIMD type or an array to allow selection by index (and then to be able to forward `array.begin()`, `array.end()`, etc...).

It has been done only for batch_bool, floating points and AVX/SSE. I guess that to be coherent, it would be nice to do the same thing for generic batches and other SIMD instruction sets... but I wanted to be sure to be on the right track before changing everything (and exposing the world to bad things :) ).

For `batch_bool`, taking an array of ints solves the problem with `-ffast-math` (which changes the conversion value of `nan` to `bool`). In this case, the potential aliasing issue is naturally solved: `__mxxx` types, even if suffixed with `d` are not assumed to contain only a given class of values. For an example: [https://godbolt.org/z/LO35aw](https://godbolt.org/z/LO35aw) (where the return of `h` is optimized, which is not the case for `g`, even if operations are purely on doubles :) ).

For execution time, I tested on simple cases and haven't seen any difference. Besides, it may be good to add some benchmarks for some `batch_bool` operations (and maybe add some tzcnt or popcnt kinds of operations).

### DIFF
--- a/include/xsimd/types/xsimd_avx_double.hpp
+++ b/include/xsimd/types/xsimd_avx_double.hpp
@@ -43,8 +43,12 @@ namespace xsimd
         bool operator[](std::size_t index) const;
 
     private:
+        union storage_t {
+            std::array<std::uint64_t, 4> arr;
+            __m256d                      reg;
+        };
 
-        __m256d m_value;
+        storage_t m_value;
     };
 
     /********************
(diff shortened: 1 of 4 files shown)


wolfv:
Awesome!

include/xsimd/types/xsimd_sse_double.hpp
```diff
@@ -103,36 +108,34 @@ namespace xsimd
...
-        return m_value;
+        return m_value.reg;
     }
 
     inline bool batch_bool<double, 2>::operator[](std::size_t index) const
     {
-        double v = reinterpret_cast<const double *>(&m_value)[index & 1];
-        std::uint64_t r = reinterpret_cast<std::uint64_t&>(v);
-        return (bool)r;
+        return bool(m_value.arr[index & 1]);
```
wolfv: I'm not even sure if we still want to return `index & 1` or if `[index]` suffices.

wolfv:
the build issue seems to be that one needs to include `<array>` on windows and apple.

include/xsimd/types/xsimd_avx_double.hpp
```diff
@@ -43,8 +43,12 @@ namespace xsimd
         bool operator[](std::size_t index) const;
 
     private:
+        union storage_t {
```
wolfv: nitpicking on code style: brackets on newline pls

include/xsimd/types/xsimd_avx_double.hpp
```diff
@@ -103,38 +107,36 @@ namespace xsimd
...
     {
+        m_value.reg=_mm256_castsi256_pd(_mm256_set1_epi32(-(int)b));
     }
 
     inline batch_bool<double, 4>::batch_bool(bool b0, bool b1, bool b2, bool b3)
-        : m_value(_mm256_castsi256_pd(
-              _mm256_setr_epi32(-(int)b0, -(int)b0, -(int)b1, -(int)b1,
-                                -(int)b2, -(int)b2, -(int)b3, -(int)b3)))
     {
+        m_value.reg=_mm256_castsi256_pd(
```
wolfv: further nitpicking: whitespace around `=`

hleclerc:
Ooops :) I have made corrections for the style + include of array.

About the remainder operation, I really don't know the reason it was there before. The users know the limits of the index value. IMHO, If they want to make a modulo, it's not the responsibility of the library. But maybe it would deserve a collegiate discussion... I don't know :)

wolfv:
Let's see if @JohanMabille has an opinion on the indexing 

btw. i think we should do the same refactoring for the float + integer classes... we could start again with AVX / SSE so we can run the benchmarks to confirm that there is no speed issue.

Would you be interested in taking the lead on those implementations as well?

PR #252, opened 2019-01-16
fix macro redefinition warning when XSIMD_FORCE_X86_INSTR_SET is defined

marcrambo:
Second attempt ;)


JohanMabille:
The failure is due to the conda switch, I'm gonna push the fix to this PR today.

PR #255, opened 2019-01-25
Add missing include guard for algorithms.hpp

zhihaoy:
```
  C:\[...]\include\xsimd/stl/algorithms.hpp(69): error C2995: 'void xsimd::
transform(I1,I2,O1,UF &&)': function template has already been defined [...]
  C:\[...]\include\xsimd/stl/algorithms.hpp(128): error C2995: 'void xsimd:
:transform(I1,I2,I3,O1,UF &&)': function template has already been defined [...]
```

Please consider releasing a new version right after merging this...

### DIFF
--- a/include/xsimd/stl/algorithms.hpp
+++ b/include/xsimd/stl/algorithms.hpp
@@ -6,6 +6,9 @@
 * The full license is in the file LICENSE, distributed with this software. *
 ****************************************************************************/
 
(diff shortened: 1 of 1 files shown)


zhihaoy:
> Is a release on Monday early enough for you?

No need to rush now, vcpkg has made a local patch so that my CI will be fine to proceed.  Thanks for the care.

PR #256, opened 2019-01-27
add CRTP base class

wolfv:


wolfv:
This will allow us to implement iterator proxies.

wolfv:
@marcrambo @andre-bergner This implements the STL like iterator stuff, with an iterator proxy. As a next step, we can implement the aligned range etc.

Also, some additional work needs to be done to support complex numbers.

Let me know if you have comments!

wolfv:
This is how it can be used right now: 

https://github.com/QuantStack/xsimd/pull/256/files#diff-92fa2b3f2ab65ab3de662f07baded954R184

wolfv:
@JohanMabille this is ready for review :)

PR #267, opened 2019-02-10
storage moved to base class

JohanMabille:


include/xsimd/types/xsimd_fallback.hpp
```diff
@@ -690,6 +693,18 @@ namespace xsimd
         }
     }
 
+    template <typename T, std::size_t N>
+    inline T& batch<T, N>::operator[](std::size_t index)
+    {
+        return this->m_value[index % base_type::size];
```
wolfv: Let's hope modern compilers can optimize this modulo away. I am pretty hopeful they do :)

JohanMabille: I'm pretty sure too. Otherwise we can remove this feature (it was not here in the previous version), it applies to fallback simd and I don't know what are their requirements regarding performances.

wolfv: haha, i can't find the implementation for other batches... in which file is it?

JohanMabille: It's in `xsimd_base` now.

PR #268, opened 2019-02-10
Do not dereference end iterators

zhihaoy:
It causes exceptions in MSVC STL in Debug config when debug
iterators are enabled (default).


JohanMabille:
Thanks for the fix!

PR #273, opened 2019-02-25
add warning if no SIMD instructions found, fix include

wolfv:


wolfv:
@ArthurAllshire can you check this PR against your crash example?

Also as specified, try with 

```
-march=armv8.1-a -mfpu=neon
```

include/xsimd/types/xsimd_types_include.hpp
```diff
@@ -67,6 +67,14 @@
 #include "xsimd_neon_complex.hpp"
 #endif
 
+#if XSIMD_ARM_INSTR_SET == XSIMD_VERSION_NUMBER_NOT_AVAILABLE && XSIMD_X86_INSTR_SET == XSIMD_VERSION_NUMBER_NOT_AVAILABLE
+	#if defined(XSIMD_ENABLE_FALLBACK)
+		#warning "No SIMD instructions enabled, using fallback mode."
+	#else
+		#error "Please compile with SIMD instructions enabled or activate the fallback mode. (e.g. for x86 -march=native or for ARM -mfpu=neon)"
```
JohanMabille: This prevents libraries from including `xsimd.hpp` if they don't enable simd instructions nor fallback (which is the case when testing for instance), I think the warning should be enough.

include/xsimd/types/xsimd_traits.hpp
```diff
@@ -11,6 +11,7 @@
 
 #include <type_traits>
 
+#include "../types/xsimd_base.hpp"
```
JohanMabille: This file is not supposed to be included directly, `xsimd_types_include.hpp` is responsible for including the required batches and related functions. So either the types / functions defined in `xsimd_base.hpp` should be moved to `xsimd_utils.hpp` or the files defining math functions for batches should only be included when simd instructions are enabled. I find this last option cleaner, however it is a bigger (potentially breaking) change.

wolfv: as discussed, this is for the next big refactoring

include/xsimd/types/xsimd_types_include.hpp
wolfv: made it a warning

PR #288, opened 2019-06-21
Update README.md for Conan installation instructions

omaralvarez:
This PR adds the necessary instructions for installing and using xsimd with the [Conan](https://conan.io/) package manager. I have also added some CMake instructions to create a new project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,16 +39,7 @@ ARM          | ARMv7, ARMv8
 
 ## Installation
 
-`xsimd` is a header-only library. 
-
-You can directly install it from the sources:
-
-```bash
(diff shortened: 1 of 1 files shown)


JohanMabille:
Thanks for the PR!

Nitpicking remark: could you change the order of how to install the library? We prefer to have the package managers section first, then the installation from sources.

The example of `cmake` should go in the `Building and Running the Tests` section rather than `Usage`, what do you think?

omaralvarez:
I agree with everything. I am making the changes as we speak.

JohanMabille:
Thanks! I'll merge this PR, feel free to open a new one if you want to add a section with an example of CMake, I will be happy to review and merge!

PR #296, opened 2019-07-04
WIP - Fixed flags for ppc architecture

JohanMabille:


hroncok:
testing a build

hroncok:
it builds

JohanMabille:
Just to be sure, it builds on all the platform, so if I tag a new release, you will be able to package it for Fedora? Or are there other issues that I should fix before releasing?

hroncok:
I've already backported both PRs and done the ARM workaround in Fedora. It builds on all our architectures.

JohanMabille:
wonderfull!

PR #299, opened 2019-07-29
Fix/xsimd scalar missing functions

serge-sans-paille:


serge-sans-paille:
These ones are need for pythran.

include/xsimd/math/xsimd_scalar.hpp
```diff
@@ -48,6 +49,7 @@ namespace xsimd
     using std::log1p;
     using std::log2;
     using std::log;
+    using std::modf;
```
wolfv: should we also add a test for modf? is it in xsimd?

serge-sans-paille: modf is not in the vector version of xsimd, so I did'nt want to add test for only the scalar mode. take that as an incentive to implement the vector version :-)

wolfv: :)

PR #301, opened 2019-07-31
Support isinf/isnan for complex number using numpy semantic

serge-sans-paille:
Only for scalar mode, let's implement this for complex numbers at some point!

### DIFF
--- a/include/xsimd/math/xsimd_scalar.hpp
+++ b/include/xsimd/math/xsimd_scalar.hpp
@@ -94,6 +94,20 @@ namespace xsimd
     }
 #endif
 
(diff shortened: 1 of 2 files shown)


wolfv:
can we remove the function in utils namespace then 

https://github.com/QuantStack/xsimd/blob/55996ddaea1e716548167639d5dd8a7d10202fd7/test/xsimd_test_utils.hpp#L34-L44

Also, is our SIMD isinf function doing the correct thing for xsimd batches?

serge-sans-paille:
I didn't implement the batch version ^^!

PR #323, opened 2019-11-20
Fixed avx512 batch for int16 selection

### LINKED ISSUE #320: _mm256_blendv_epi16 does not exist
The used methods _mm256_blendv_epi16 in 
https://github.com/xtensor-stack/xsimd/blob/a8d6f7855d33d3263a4d7804e990510093c6b457/include/xsimd/types/xsimd_avx512_int16.hpp#L321-L322

do not exist.

JohanMabille:
Fixes #320 for real

### DIFF
--- a/include/xsimd/types/xsimd_avx512_int16.hpp
+++ b/include/xsimd/types/xsimd_avx512_int16.hpp
@@ -318,8 +318,8 @@ namespace xsimd
                 XSIMD_SPLIT_AVX512(a);
                 XSIMD_SPLIT_AVX512(b);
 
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
So the split does what we were hoping wrt duplication?

JohanMabille:
The split simply extracts the higher and the lower parts, without changing anything.
However, every comparison intrinsics operating on batches of X bits integers returns a mask where contiguous blocks of X bits are set to the same value and I think it is reasonable to assume that any function returning a mask should have similar behavior.

PR #337, opened 2019-12-13
Disabled Knight Landings SIMD

JohanMabille:


JohanMabille:
The truth is out there

PR #345, opened 2019-12-17
corrected lt for unsigned integers

### LINKED ISSUE #339: uint32_t bug on AVX2
Using uint32_t batches on an AVX2 machine causes wrong results when comparing values.

The code: 
```cpp
#include <iostream>
#include <xsimd/xsimd.hpp>
int main() {
  xsimd::batch<uint32_t, 8> a(0, 1, 2, 3, 4, 5, 6, 7);
  xsimd::batch<uint32_t, 8> b(5);
  auto c = a - b;
  auto d = c < 3;
  std::cout << c << std::endl;
  std::cout << d << std::endl;

  return 0;
}
```
produces: 
```
(4294967291, 4294967292, 4294967293, 4294967294, 4294967295, 0, 1, 2)
(T, T, T, T, T, T, T, T)
```

wolfv:
this fixes #339


wolfv:
cc @JohanMabille i think this is ready to go. We could consider a new release after merge.

PR #355, opened 2020-02-11
rem_pio2: undef `ONCE0`

timblechmann:
`ONCE0` is defined, but not undefined. this causes possible name clashes downstream


JohanMabille:
Good catch! Thanks.

PR #356, opened 2020-03-24
use arithmetic right shift SSE/AVX intrinsics for signed integer >>

astrohawk:
The signed integer right shift operators (operator>>) for SSE, AVX, and AVX512 use the logical right shift intrinsics, not the arithmetic right shift intrinsics. This results in incorrect shift result for negative numbers and is also inconsistent with the neon right shift operators which correctly use the signed right shift intrinsics.
I have also modified the basic integer test to make every second test value negative when testing signed integers in order to increase test coverage to negative numbers.


wolfv:
Looks good to me! 

@JohanMabille what do you think?

JohanMabille:
Thanks for the fix!

PR #357, opened 2020-03-25
Fix avx512_fallback_batch_bool constructor and load_values

astrohawk:
This caused avx512 int16_t and uint16_t unit tests to fail when avx512bw
was not enabled (only avx512f enabled)


JohanMabille:
Thanks!

@wolfv This looks good to me, what do you think?

wolfv:
Looks good to me too :)

PR #358, opened 2020-03-25
Clean up abs implementation for Intel intrinsics

astrohawk:
This includes:
1. Using the abs intrinsics instead of the sign intrinsics for 8, 16,
and 32 bit arithmetic, e.g. _mm_abs_epi8/16/32 instead of
_mm_sign_epi8/16/32 (both are SSSE3), using _mm256_abs_epi8/16/32
instead of _mm256_sign_epi8/16/32 (both are AVX2).
2. Add 64 bit abs intrinsics that came with AVX512VL to SSE and AVX
3. Add guards around 8 and 16 bit AVX512 abs intrinsics that require
AVX512BL

### DIFF
--- a/include/xsimd/types/xsimd_avx512_int16.hpp
+++ b/include/xsimd/types/xsimd_avx512_int16.hpp
@@ -336,7 +336,14 @@ namespace xsimd
         {
             static batch_type abs(const batch_type& rhs)
             {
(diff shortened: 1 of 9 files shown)


JohanMabille:
Thanks!

The build on travis passed but it seems they have some issue to notify github.

PR #360, opened 2020-03-28
Add new batch_cast function to expand support for type conversion

astrohawk:
Hi guys. I really hope you're ok with me adding such a big new feature.
I think this is a nice addition to the API and I actually have a use
case for this on a personal project I'm working on.

Currently there is support for converting floats to ints (to_int) and
ints to floats (to_float) but there is no way to do the equivalent of a
static_cast, i.e. request a value of some specific type be converted to
another specific type. The new batch_cast function will allow you to do
this and leverage any intrinsics that exist to accelerate that
conversion.

Example usage:
```C++
xsimd::batch<int32_t, 4> b_int(...);
xsimd::batch<float, 4> f_float = xsimd::batch_cast<float>(b_int);
```

Let me know if you would prefer to have the batch_cast template type to
take a full batch type instead of just the underlying type, e.g.
```C++
xsimd::batch<int32_t, 4> b_int(...);
xsimd::batch<float, 4> f_float = xsimd::batch_cast<batch<float, 4>>(b_int);
```
I implemented it the way it is now as I think it provides a simpler API
to users, i.e. if they want to convert some other vector to a vector of
floats, they wouldn't be required to write something like
```C++
auto b_other = ...;
auto b_float = xsimd::batch_cast<batch<float, b_other.size>>(b_other);
```
The downside of this is that it can potentially result in confusing
compiler errors if the user attempts to cast to an unrepresentable
vector, i.e. cast a `batch<double, 2>` to an `int8_t` (i.e.
`batch<int8_t, 2>`). If the user is forced to specify this type
themselves (e.g. `batch_cast<batch<int8_t, 2>)(...)>`), it may result in
simpler compiler errors and therefore be easier them to understand
what error they have made. It would also make the API consistent between
bitwise_cast and batch_cast.

Finally, I also fixed some of the exiting double-to-int64 to_int and
int64-to-double to_float fallback implementations as they produced
incorrect results when converting values outside the representable range
of int32. This was due to the fact that these conversions were done via
int32, i.e. double-to-int64 was actually double-to-int32-to-int64 and
int64-to-double was actually int64-to-int32-to-double.

### DIFF
--- a/include/xsimd/config/xsimd_instruction_set.hpp
+++ b/include/xsimd/config/xsimd_instruction_set.hpp
@@ -114,6 +114,10 @@
     #define XSIMD_AVX512VL_AVAILABLE 1
 #endif
 
+#if defined(__AVX512DQ__)
+    #define XSIMD_AVX512DQ_AVAILABLE 1
+#endif
+
 #if defined(__AVX512BW__)
     #define XSIMD_AVX512BW_AVAILABLE 1
 #endif
--- a/include/xsimd/types/xsimd_avx512_conversion.hpp
+++ b/include/xsimd/types/xsimd_avx512_conversion.hpp
@@ -13,10 +13,10 @@
 
 #include "xsimd_avx512_double.hpp"
 #include "xsimd_avx512_float.hpp"
+#include "xsimd_avx512_int8.hpp"
+#include "xsimd_avx512_int16.hpp"
 #include "xsimd_avx512_int32.hpp"
 #include "xsimd_avx512_int64.hpp"
-#include "xsimd_avx512_int16.hpp"
-#include "xsimd_avx512_int8.hpp"
 
 namespace xsimd
 {
@@ -57,7 +57,18 @@ namespace xsimd
 
     inline batch<int64_t, 8> to_int(const batch<double, 8>& x)
     {
-        return _mm512_cvtepi32_epi64(_mm512_cvttpd_epi32(x));
+#if defined(XSIMD_AVX512DQ_AVAILABLE)
+        return _mm512_cvttpd_epi64(x);
+#else
+        return batch<int64_t, 8>(static_cast<int64_t>(x[0]),
+                                 static_cast<int64_t>(x[1]),
+                                 static_cast<int64_t>(x[2]),
+                                 static_cast<int64_t>(x[3]),
+                                 static_cast<int64_t>(x[4]),
+                                 static_cast<int64_t>(x[5]),
+                                 static_cast<int64_t>(x[6]),
+                                 static_cast<int64_t>(x[7]));
+#endif
     }
 
     inline batch<float, 16> to_float(const batch<int32_t, 16>& x)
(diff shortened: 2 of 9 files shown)


wolfv:
Hi awesome change! Regarding the compiler errors, do you think we can make use of `static_assert` to give some hints? https://en.cppreference.com/w/cpp/language/static_assert

JohanMabille:
This is a really nice change! I agree with you that specifying the value type is more convenient than specifyifng the full batch type.

As suggested by @wolfv using static_assert will produce nice error messages.

astrohawk:
I'll play around with using void_t (I may have to reimplement as it's C++17) to create a is_valid_batch check then add another commit

astrohawk:
I think it's actually going to be quite difficult to add a static_assert that will give more informative compiler errors as the invalid batch is specified as the return type of the batch_cast function. This means that the compiler errors out after parsing the function signature and and any static_assert declaration located in the body of the function will have no effect.

JohanMabille:
A solution would be to remove these functions from the overload resolution set when the sizes of the batches are different. This leads to "no matching function to call ..." error.

Then you can introduce an overload (that is enabled ONLY when the sizes of the batches are different), which returns something different (it can even be auto, of the simd_fallbatch, or any type that would not prevent the compilation), and add the static_assert in that catch all function.

Just keep in mind that all this metaprogramming stuff shoul dbe compatible with C++11.

astrohawk:
Because the current implementation doesn't allow the user to ask for a specific batch output type, only the value type of the resulting batch type, it's actually not an issue of different batch sizes. The API enforces this. It's that some requested conversions with the same batch size produce invalid resulting batch types, for example if you attempted to cast a `batch<double, 2>` to a int8_t:
```C++
xsimd::batch<double, 2> b_double(...);
xsimd::batch<int8_t, 2> b_int8 = xsimd::batch_cast<int8_t>(b_int);
```
`batch<int8_t, 2>` is an invalid batch type and the compiler will complain about it:
```
error: invalid use of incomplete type ‘class xsimd::batch<signed char, 2>’
```
Personally, I don't think this error is too hard to understand but I think it is worth discussing the pros and cons of both approaches.

I have also been very careful to make sure this works in C++11. In fact I haven't even used an auto return type as this is C++14. Being able to use an auto return type would have actually solved this issue as the compiler would have had to parse the function body in order to deduce the return type of the function, allowing the static_assert to fail and a nice compiler error message to be outputted.

JohanMabille:
I agree with you that when the LHS batch is entirely specified, the error is easy to understand.
I was rather thinking of this kind of code:

```cpp
batch<double, 4> bd = ....;
auto r = batch_cast<int8_t>(bd);
```

In that case there will be a "no matching call to batch_cast ..." error.

Regarding the API of `bitwise_cast`, I realized that we chose to pass the full batch type instead of the value type so that we can use it for any available instruction set (for instance we can use it for batch<double, 2> even when AVX instruction set is available). However `bitwise_cast` is more low-level and less likely to be called by end-user than `batch_cast`.

So I'm still in favor of keeping the API of `batch_cast` as provided in this PR, we can complicate it to handle batch template parameter later if that's required.

astrohawk:
We could change the definition to be the following:
```C++
template <class T_out, class T_in, std::size_t N>
inline typename std::enable_if<is_valid_batch<T_out, N>::value, batch<T_out, N>>::type
batch_cast(const batch<T_in, N>& x)
{
    return batch_cast_impl<T_in, T_out, N>::run(x);
}

template <class T_out, class T_in, std::size_t N>
inline typename std::enable_if<!is_valid_batch<T_out, N>::value>::type
batch_cast(const batch<T_in, N>&)
{
    static_assert(is_valid_batch<T_out, N>::value, "Unable to cast batch to specified type; resulting batch type is not valid");
}
```
which returns the error:
```
/XXX/xsimd/test/xsimd_basic_test.cpp: In member function ‘virtual void xsimd_is_valid_batch_Test::TestBody()’:
/XXX/xsimd/test/xsimd_basic_test.cpp:1111:10: error: ‘void out’ has incomplete type
 1111 |     auto out = xsimd::batch_cast<int8_t>(in);
      |          ^~~
In file included from /XXX/xsimd/include/xsimd/types/xsimd_traits.hpp:16,
                 from /XXX/xsimd/include/xsimd/xsimd.hpp:17,
                 from /XXX/xsimd/test/xsimd_basic_test.hpp:19,
                 from /XXX/xsimd/test/xsimd_basic_test.cpp:18:
/XXX/xsimd/include/xsimd/types/xsimd_base.hpp: In instantiation of ‘typename std::enable_if<(! xsimd::is_valid_batch<T_out, N>::value)>::type xsimd::batch_cast(const xsimd::batch<T_out, N>&) [with T_out = signed char; T_in = float; long unsigned int N = 4; typename std::enable_if<(! xsimd::is_valid_batch<T_out, N>::value)>::type = void]’:
/XXX/xsimd/test/xsimd_basic_test.cpp:1111:44:   required from here
/XXX/xsimd/include/xsimd/types/xsimd_base.hpp:1786:49: error: static assertion failed: Unable to cast batch to specified type; resulting batch type is not valid
 1786 |         static_assert(is_valid_batch<T_out, N>::value, "Unable to cast batch to specified type; resulting batch type is not valid");
      |                                                 ^~~~~
```
Obviously the message we return can be changed, though due to the limitations of static assert, we wouldn't actually be able to return the invalid batch type in the static assert message itself.
A twist on this option would be to only have the first definition, i.e. skip the second function definition that contains the static_assert. This would result in the following error as you mentioned in the previous comment:
```
/XXX/xsimd/test/xsimd_basic_test.cpp: In member function ‘virtual void xsimd_is_valid_batch_Test::TestBody()’:
/XXX/xsimd/test/xsimd_basic_test.cpp:1111:44: error: no matching function for call to ‘batch_cast<int8_t>(xsimd::batch<float, 4>&)’
 1111 |     auto out = xsimd::batch_cast<int8_t>(in);
      |                                            ^
In file included from /XXX/xsimd/include/xsimd/types/xsimd_traits.hpp:16,
                 from /XXX/xsimd/include/xsimd/xsimd.hpp:17,
                 from /XXX/xsimd/test/xsimd_basic_test.hpp:19,
                 from /XXX/xsimd/test/xsimd_basic_test.cpp:18:
/XXX/xsimd/include/xsimd/types/xsimd_base.hpp:1777:5: note: candidate: ‘template<class T_out, class T_in, long unsigned int N> typename std::enable_if<xsimd::is_valid_batch<T_out, N>::value, xsimd::batch<T_in, N> >::type xsimd::batch_cast(const xsimd::batch<T_out, N>&)’
 1777 |     batch_cast(const batch<T_in, N>& x)
      |     ^~~~~~~~~~
/XXX/xsimd/include/xsimd/types/xsimd_base.hpp:1777:5: note:   template argument deduction/substitution failed:
/XXX/xsimd/include/xsimd/types/xsimd_base.hpp: In substitution of ‘template<class T_out, class T_in, long unsigned int N> typename std::enable_if<xsimd::is_valid_batch<T_out, N>::value, xsimd::batch<T_in, N> >::type xsimd::batch_cast(const xsimd::batch<T_out, N>&) [with T_out = signed char; T_in = float; long unsigned int N = 4]’:
/XXX/xsimd/test/xsimd_basic_test.cpp:1111:44:   required from here
/XXX/xsimd/include/xsimd/types/xsimd_base.hpp:1777:5: error: no type named ‘type’ in ‘struct std::enable_if<false, xsimd::batch<signed char, 4> >’
```

This can be contrasted with the existing code:
```C++
template <class T_out, class T_in, std::size_t N>
inline batch<T_out, N> batch_cast(const batch<T_in, N>& x)
{
    return batch_cast_impl<T_in, T_out, N>::run(x);
}
```
which returns the error:
```
/XXX/xsimd/test/xsimd_basic_test.cpp: In member function ‘virtual void xsimd_is_valid_batch_Test::TestBody()’:
/XXX/xsimd/test/xsimd_basic_test.cpp:1111:44: error: invalid use of incomplete type ‘class xsimd::batch<signed char, 4>’
 1111 |     auto out = xsimd::batch_cast<int8_t>(in);
      |                                            ^
In file included from /XXX/xsimd/include/xsimd/types/xsimd_base.hpp:25,
                 from /XXX/xsimd/include/xsimd/types/xsimd_traits.hpp:16,
                 from /XXX/xsimd/include/xsimd/xsimd.hpp:17,
                 from /XXX/xsimd/test/xsimd_basic_test.hpp:19,
                 from /XXX/xsimd/test/xsimd_basic_test.cpp:18:
/XXX/xsimd/include/xsimd/types/xsimd_utils.hpp:26:11: note: declaration of ‘class xsimd::batch<signed char, 4>’
   26 |     class batch;
      |           ^~~~~
In file included from /XXX/xsimd/include/xsimd/types/xsimd_traits.hpp:16,
                 from /XXX/xsimd/include/xsimd/xsimd.hpp:17,
                 from /XXX/xsimd/test/xsimd_basic_test.hpp:19,
                 from /XXX/xsimd/test/xsimd_basic_test.cpp:18:
/XXX/xsimd/include/xsimd/types/xsimd_base.hpp: In instantiation of ‘xsimd::batch<T_in, N> xsimd::batch_cast(const xsimd::batch<T_out, N>&) [with T_out = signed char; T_in = float; long unsigned int N = 4]’:
/XXX/xsimd/test/xsimd_basic_test.cpp:1111:44:   required from here
/XXX/xsimd/include/xsimd/types/xsimd_base.hpp:1790:28: error: return type ‘class xsimd::batch<signed char, 4>’ is incomplete
 1790 |     inline batch<T_out, N> batch_cast(const batch<T_in, N>& x)
      |                            ^~~~~~~~~~
/XXX/xsimd/include/xsimd/types/xsimd_base.hpp:1792:52: error: invalid use of incomplete type ‘class xsimd::batch<signed char, 4>’
 1792 |         return batch_cast_impl<T_in, T_out, N>::run(x);
      |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~
In file included from /XXX/xsimd/include/xsimd/types/xsimd_base.hpp:25,
                 from /XXX/xsimd/include/xsimd/types/xsimd_traits.hpp:16,
                 from /XXX/xsimd/include/xsimd/xsimd.hpp:17,
                 from /XXX/xsimd/test/xsimd_basic_test.hpp:19,
                 from /XXX/xsimd/test/xsimd_basic_test.cpp:18:
/XXX/xsimd/include/xsimd/types/xsimd_utils.hpp:26:11: note: declaration of ‘class xsimd::batch<signed char, 4>’
   26 |     class batch;
      |           ^~~~~
```
Note that the output in this case actually prints the invalid type.

Personally, I'm not sure that adding lots of template metaprogramming in the function signature is such a good idea. It makes it very difficult for somebody who is not comfortable with template metaprogramming to understand what the function is doing. Also the message the static assert returns does not contain significantly more information to make the trade-off worth it; in fact in some ways it actually contains even less info (it does not output the invalid type). I could make it so that it prints out both errors (both the invalid type produced by the compiler, plus the message in the static assert) but then we risk printing too much information in the compiler error message and overwhelming the user, resulting in the worst of both words.

I'm happy to defer the decision on which option we take to you. Sorry for the long write up and thank you for taking to time to respond to these comments.

JohanMabille:
Ah I wasn't sure that the invalid type would be printed with calls like `auto b = batch_cast`. Thanks for checking it!
In that case, I totally agree with you, no need to complicate the design to introduce messages with less infofmation!

> Sorry for the long write up

Please don't be ;) Having detailed explanations is always a good thing, especially for "future us" if we have to retrieve why we did things this way.

Thanks again for your time!

PR #380, opened 2020-09-03
Various minor improvements

emmenlau:
This PR adds a number of small improvements to build with pedantic build settings using `clang-cl` on MSVC. The changes are:
 - cmake should try not to override the `CMAKE_BUILD_TYPE` or the `-march`, if any of those are given by the user. This is so that users can easily compile for a specific build type or target architecture (i.e. distributors may set -march to an older target).
 - Removed the `EXCLUDE_FROM_ALL` from building tests, benchmarks and examples. There are cmake options like `-DBUILD_TESTS=ON/OFF` to enable or disable these targets, so they may as well be built by default?
 - A few cases of shadowed parameters have been fixed by renaming the parameter with a prefix 'a'
 - Fixed two cases where parameters were not valid for clang or generally not valid on MSVC


JohanMabille:
I finally had the time to review it, thanks very much this is very neat!

PR #384, opened 2020-09-12
fix static_assert syntax error and use fallback mode when XSIMD_ENABL…

zouyxdut:


JohanMabille:
Thanks! The filaure is unrelated (it's due to a change on Intel website), but I'd like to fix it before merging your PR so that it does not break the master branch (I will do that in a dedicated PR and merge your after).

Issue #387, opened 2020-09-18
Missing fma implementation for xcomplex


pesterie:
See [here](https://godbolt.org/z/adf66o)

serge-sans-paille:
@JohanMabille I'll take that one.

JohanMabille:
@serge-sans-paille notice there is already #388 opened.

serge-sans-paille:
ok, it resembles my implementation, with a few quirks. I'll use it as a basis (and also extend it so support fms and friends). @pesterie are you fine with that?

pesterie:
Hi @JohanMabille & @serge-sans-paille !
@serge-sans-paille yes, sure, feel free to iterate from this !
Will have a look when your done !

serge-sans-paille:
@pesterie implementation available in #445

pesterie:
Sweet @serge-sans-paille ! Thx !

PR #388, opened 2020-09-22
Fix fma for complex numbers (std and xtl)

pesterie:
This is a really fast written fix for #387 let me know if I missed something ! ;)


JohanMabille:
Thanks for this PR, the code is really neat!

The error on Windows looks like a compiler issue, but I won't have the time to investigate it before next week.

serge-sans-paille:
Obsoleted by #445

PR #395, opened 2020-10-19
Fix missing include for gcc-11

serge-sans-paille:
Required for std::numeric_limits


hroncok:
See also https://github.com/xtensor-stack/xsimd/pull/415

PR #396, opened 2020-10-21
CMake: Modernized GTest-integration

emmenlau:
This is a minor PR that improves the integration of testing and specifically of GTest. It brings better compatibility with FindGTest from (future) cmake.

Please let me know if it would need more motivation? :-)


JohanMabille:
A detailed explanation of the reasons for this PR can be found in https://github.com/xtensor-stack/xtl/pull/220#issuecomment-713641553.

PR #411, opened 2021-01-24
Eliminate -Wpedantic warnings and avoid dereferencing type-punned pointer.

mainland:
These two patches 1) fix `-Wpedantic` warnings about unnecessary semicolons and 2) avoid dereferencing a type-puned pointer, which can break strict aliasing.


JohanMabille:
Thanks!

PR #419, opened 2021-02-23
Fix any() and all() on Neon

pitrou:
Also improve the bool reduction tests, so as to exhibit the bug.

### DIFF
--- a/include/xsimd/types/xsimd_neon_bool.hpp
+++ b/include/xsimd/types/xsimd_neon_bool.hpp
@@ -386,14 +386,26 @@ namespace xsimd
 
             static bool all(const batch_type& rhs)
             {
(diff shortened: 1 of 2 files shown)


pitrou:
I validated that this patch fixes the issue we witnessed on the Arrow test suite: https://travis-ci.com/github/pitrou/arrow/jobs/485701576

cyb70289:
> I validated that this patch fixes the issue we witnessed on the Arrow test suite: https://travis-ci.com/github/pitrou/arrow/jobs/485701576

Above URL is the successful test after applying this patch.
The failed test suite is at https://travis-ci.com/github/pitrou/arrow/jobs/485572758.

JohanMabille:
Thanks!

PR #426, opened 2021-03-02
Test refactoring

JohanMabille:
@pitrou you may be interested in this, I finally decided to refactor the tests to have something more intuitive. Any feedback is welcome.

Here is an example of error message when a test fails:

```
[ RUN      ] sse/batch/int64_t.arithmetic
/home/yoyo/dev/xtensor-stack/xsimd/test/test_batch.cpp:66: Failure
Expected equality of these values:
  res
    Which is: { 4, -5 }
  expected
    Which is: { 4, 5 }
  while testing batch + batch
[  FAILED  ] sse/batch/int64_t.arithmetic, where TypeParam = xsimd::batch<long, 2ul> (0 ms)
```


test/CMakeLists.txt
```diff
@@ -178,7 +180,7 @@ set(XSIMD_TESTS
     xsimd_tester.hpp
     xsimd_test_utils.hpp
     xsimd_trigonometric_test.hpp
-    xsimd_trigonometric_test.cpp
+    xsimd_trigonometric_test.cpp]]
 )
 
 add_executable(test_xsimd ${XSIMD_TESTS} ${XSIMD_HEADERS})
```
pitrou, line 186: I'm curious: you need to add headers to the executable sources?

test/test_batch.cpp
```diff
@@ -0,0 +1,134 @@
...
+*                                                                          *
+* The full license is in the file LICENSE, distributed with this software. *
+****************************************************************************/
+
+#include <functional>
+
+#include "test_utils.hpp"
+
+template <class B>
+class batch : public testing::Test
```
pitrou, line 16: FTR, I would probably name it `batch_test` or something similar, calling it `batch` invites confusion with `xsimd::batch` and may produce head-scratching.

test/test_batch.cpp
```diff
@@ -0,0 +1,134 @@
...
+{
+    this->test_constructors();
+}
+
+TYPED_TEST_P(batch, arithmetic)
+{
+    this->test_arithmetic();
+}
+
+REGISTER_TYPED_TEST_SUITE_P(
```
pitrou, line 121: Note that `TYPED_TEST_SUITE_P` is the more verbose form of type-parametered test, for when you don't know instantiation types upfront. Usually it's simpler to use `TYPED_TEST_SUITE` which doesn't force you to list test functions again.

pitrou (commented):
Nice! This looks much more comfortable to work with indeed.

test/CMakeLists.txt
JohanMabille: This allows cmake to detect changes in the header and trigger a new build when you run the make command.

pitrou: Ah, I would have sworn it detected dependencies automatically. My bad.

test/test_batch.cpp
JohanMabille: That was its initial name. However, the complete output is something like:

```
[----------] 9 tests from sse/batch/double, where TypeParam = xsimd::batch<double, 2ul>
[ RUN      ] sse/batch/double.load_store

```

I found the `batch_test` name a bit repetitive in the output. But I can definitely revert that

test/test_batch.cpp
JohanMabille: Indeed, but the issue here is that you don't know instantiation types upfront. They will depend on the available instruction set on the machine, thus the `TYPED_TEST_SUITE_P`.

pitrou: Personally, I would go with concatenating. From experience, when adding a new test function it's a bit cumbersome that you have to remember to list it in the macro call (also that macro call becomes very long if there are many test functions).

JohanMabille: Actually, having this `TYPED_TEST_SUITE_P` makes it easier to identify which instruciton set is used in the tests since its name is used as the prefix.

JohanMabille: I found another way to have the instruction set name in the output, so let's go with concatenating.

SylvainCorlay:
Woot!

PR #429, opened 2021-03-14
Rename shadowing variables

neok-m4700:
Fix warnings using either `$ g++ -Wshadow=compatible-local` or `$ clang++ -Wshadow-uncaptured-local`.
```bash
[...]/include/xsimd/types/xsimd_avx_int8.hpp:413:62: warning: declaration of ‘int32_t rhs’ shadows a parameter [-Wshadow=compatible-local]
  413 |         return avx_detail::shift_impl([](int8_t val, int32_t rhs) { return val << rhs; }, lhs, rhs);
      |                                                      ~~~~~~~~^~~
[...]/include/xsimd/types/xsimd_avx_int8.hpp:408:79: note: shadowed declaration is here
  408 |     inline batch<int8_t, 32> operator<<(const batch<int8_t, 32>& lhs, int32_t rhs)
      |                                                                       ~~~~~~~~^~~
```

### DIFF
--- a/include/xsimd/types/xsimd_avx512_int16.hpp
+++ b/include/xsimd/types/xsimd_avx512_int16.hpp
@@ -498,7 +498,7 @@ namespace xsimd
         return _mm512_srai_epi16(lhs, rhs);
 #endif
 #else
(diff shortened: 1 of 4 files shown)


JohanMabille:
Thanks!

PR #434, opened 2021-03-18
Add zip_hi/zip_lo for shuffling operations

guyuqi:
Add zip_hi/zip_lo for shuffling operations mentioned in issues#412.

`test_xsimd` is passed Arm64 & x86:

```
Instruction set: ARMv8 64bit Neon
[==========] Running 369 tests from 85 test suites.
[----------] Global test environment set-up.
[----------] 3 tests from algorithms
[ RUN      ] algorithms.binary_transform
[       OK ] algorithms.binary_transform (0 ms)
[ RUN      ] algorithms.unary_transform
......
....
..
snip
...
[----------] 2 tests from trigonometric_test/arm_double, where TypeParam = xsimd::batch<double, 2ul>
[ RUN      ] trigonometric_test/arm_double.trigonometric
[       OK ] trigonometric_test/arm_double.trigonometric (3 ms)
[ RUN      ] trigonometric_test/arm_double.reciprocal
[       OK ] trigonometric_test/arm_double.reciprocal (2 ms)
[----------] 2 tests from trigonometric_test/arm_double (6 ms total)

[----------] Global test environment tear-down
[==========] 369 tests from 85 test suites ran. (237 ms total)
[  PASSED  ] 369 tests.

```

### DIFF
--- a/include/xsimd/types/xsimd_base.hpp
+++ b/include/xsimd/types/xsimd_base.hpp
@@ -283,6 +283,12 @@ namespace xsimd
     batch_type_t<X> select(const typename simd_batch_traits<X>::batch_bool_type& cond, const simd_base<X>& a, const simd_base<X>& b);
 
     template <class X>
+    batch_type_t<X> zip_lo(const simd_base<X>& lhs, const simd_base<X>& rhs);
+
+    template <class X>
+    batch_type_t<X> zip_hi(const simd_base<X>& lhs, const simd_base<X>& rhs);
+
+    template <class X>
     typename simd_batch_traits<X>::batch_bool_type
     isnan(const simd_base<X>& x);
 
@@ -1775,6 +1781,36 @@ namespace xsimd
         using value_type = typename simd_batch_traits<X>::value_type;
         using kernel = detail::batch_kernel<value_type, simd_batch_traits<X>::size>;
         return kernel::select(cond, a(), b());
+    }
+
+    /**
+     * Unpack and interleave data from the LOW half of batches \c lhs and \c rhs.
+     * Store the results in the Return value.
+     * @param lhs a batch of integer or floating point or double precision values.
+     * @param rhs a batch of integer or floating point or double precision values.
+     * @return a batch of the low part of shuffled values.
+     */
+    template <class X>
+    inline batch_type_t<X> zip_lo(const simd_base<X>& lhs, const simd_base<X>& rhs)
+    {
+        using value_type = typename simd_batch_traits<X>::value_type;
+        using kernel = detail::batch_kernel<value_type, simd_batch_traits<X>::size>;
+        return kernel::zip_lo(lhs(), rhs());
+    }
+
+    /**
+     * Unpack and interleave data from the HIGH half of batches \c lhs and \c rhs.
+     * Store the results in the Return value.
+     * @param lhs a batch of integer or floating point or double precision values.
+     * @param rhs a batch of integer or floating point or double precision values.
+     * @return a batch of the high part of shuffled values.
+     */
+    template <class X>
+    inline batch_type_t<X> zip_hi(const simd_base<X>& lhs, const simd_base<X>& rhs)
+    {
+        using value_type = typename simd_batch_traits<X>::value_type;
+        using kernel = detail::batch_kernel<value_type, simd_batch_traits<X>::size>;
+        return kernel::zip_hi(lhs(), rhs());
     }
 
     /**
--- a/include/xsimd/types/xsimd_neon_double.hpp
+++ b/include/xsimd/types/xsimd_neon_double.hpp
@@ -580,6 +580,16 @@ namespace xsimd
                 return vbslq_f64(cond, a, b);
             }
 
+            static batch_type zip_lo(const batch_type& lhs, const batch_type& rhs)
+            {
+                return vzip1q_f64(lhs, rhs);
+            }
+
+            static batch_type zip_hi(const batch_type& lhs, const batch_type& rhs)
+            {
+                return vzip2q_f64(lhs, rhs);
+            }
+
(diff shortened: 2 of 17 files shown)


guyuqi:
Build failed in _clan_:
`2021-03-18T10:45:02.5270618Z /home/vsts/work/1/s/include/xsimd/types/xsimd_neon_float.hpp:644:24: error: use of undeclared identifier 'vzip1q_f32'; did you mean 'vzipq_f32'?`

`vzip1q_f32` is defined in [Arm64 GCC](https://raw.githubusercontent.com/gcc-mirror/gcc/master/gcc/config/aarch64/arm_neon.h).

JohanMabille:
According to [this](https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=vzip1q_f32), it is available on 64 bits arhcitecture only.

include/xsimd/types/xsimd_neon_double.hpp
```diff
@@ -580,6 +580,18 @@ namespace xsimd
...
+            {
+                return vzip1q_f64(lhs, rhs);
+            }
+
+            static batch_type zip_hi(const batch_type& lhs, const batch_type& rhs)
+            {
+                return vzip2q_f64(lhs, rhs);
+            }
+#endif
+
```
JohanMabille, line 594: I think we should provide a fallback implementation when `XSIMD_ARM_INSTR_SET < XSIMD_ARM8_64_VERSION`. Otherwise the code will fail to compile if the user calls the `zip_lo` / `zip_hi` methods

include/xsimd/types/xsimd_base.hpp
```diff
@@ -1777,6 +1783,36 @@ namespace xsimd
...
+     * @param lhs a batch of integer or floating point or double precision values.
+     * @param rhs a batch of integer or floating point or double precision values.
+     * @return a batch of the low part of shuffled values.
+     */
+    template <class X>
+    inline batch_type_t<X> zip_lo(const simd_base<X>& lhs, const simd_base<X>& rhs)
+    {
+        using value_type = typename simd_batch_traits<X>::value_type;
+        using kernel = detail::batch_kernel<value_type, simd_batch_traits<X>::size>;
+        return kernel::zip_lo(lhs(), rhs());
```
JohanMabille, line 1798: This requires to define `zip_lo` for all `batch_kernel` classes, otherwise the code may fail to compile if the user calls `zip_lo` (or maybe the PR is still in progress?)

include/xsimd/types/xsimd_base.hpp
```diff
@@ -1777,6 +1783,36 @@ namespace xsimd
...
+     * @param lhs a batch of integer or floating point or double precision values.
+     * @param rhs a batch of integer or floating point or double precision values.
+     * @return a batch of the high part of shuffled values.
+     */
+    template <class X>
+    inline batch_type_t<X> zip_hi(const simd_base<X>& lhs, const simd_base<X>& rhs)
+    {
+        using value_type = typename simd_batch_traits<X>::value_type;
+        using kernel = detail::batch_kernel<value_type, simd_batch_traits<X>::size>;
+        return kernel::zip_hi(lhs(), rhs());
```
JohanMabille, line 1813: Same remark here.

include/xsimd/types/xsimd_neon_double.hpp
guyuqi: Really thanks for your comments.
In my original thought, it's meaningless for users to call `zip_lo/hi` on Arm_32 which does not support ARM8_64.
And consider that: 
1. when user calls `zip_lo/hi ` on `x86`,  it's ok that Arm `zip_lo/hi` implementation would not be compiled.
2. when user calls `zip_lo/hi ` on `ARM8_64` , that's definitely ok.  

So I didn't add the fallback implementation of `zip_lo/hi` .

But for code consistency and unexpected user's calling on Arm_32, it makes sense to add fallback implementation to source code tree .
I'll add the fallback implementation soon.

On the other hand, it might suffer performance loss if user calls fallback implementation of `zip_lo/hi ` on < `ARM8_64`.
IMHO, we should explain it in our documentation to prevent users from calling `zip_lo/hi` on Arm_32. 

WDYT where we should put the explanation in ?  
Comments directly in the code or in [xsimd/docs/](https://github.com/xtensor-stack/xsimd/tree/master/docs)?

guyuqi:
It's tricky that Arm32/v7 do not support data type: double(F64) like [vreinterpretq_s64_f64](https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=vreinterpretq_s64_f64).

So users should not be allowed to call `zip_lo/hi` with data type: F64 on Arm32/v7.

guyuqi:
@JohanMabille 
Added fallback implementation for Arm32/v7 @3rd commit.

JohanMabille:
Sorry for the late comment, the failure is unrelated to your change and will be fixed in #446

guyuqi:
@JohanMabille Thanks for the info. 
Rebased it.

JohanMabille:
Thanks! The current implementation looks good to me. I think we miss the implementation for AVX and AVX512, as long as some tests.

guyuqi:
> Thanks! The current implementation looks good to me. I think we miss the implementation for AVX and AVX512, as long as some tests.

Thanks for the review.
Zip_hi/lo were implemented for 128bit data-width of register:
```
x86:  SSE
Arm: Neon
```
For 256 and 512 data-width: 
```
x86: avx/avx512
Arm64: SVE
```
How about to implement them in another PR?

JohanMabille:
> How about to implement them in another PR?

I'm fine with that, but in that case the tests should be added in the second PR (otherwise they won't build for AVX and AVX512 or you will have to define a new list of batches specific to these functions).

Let's merge ;)

PR #438, opened 2021-03-19
Added aarch64 tests back

JohanMabille:


wolfv:
@JohanMabille I am currently paying for a Graviton 2 instance on AWS (10$ a month). We can easily install a self-hosted runner on it to run the tests if we want to.

JohanMabille:
Yeah looks like we don't really have the choice :s I've struggled a lot with this when I mgrated the CI from travis to Azure, and honestly I still don't understand why it fails.

pitrou:
@cyb70289 FYI :-)

serge-sans-paille:
Obsoleted by #479

PR #439, opened 2021-03-20
Provide batch_constant and batch_bool_constant, and use them to speci…

serge-sans-paille:
…alize the select function

in SSE4 and AVX, it's possible to pass an immediate to the blending intrinsic
instead of a register. Provide the required support in terms of batch_constant
and batch_bool_constant, and fallback to the vector version otherwise.

### DIFF
--- a/include/xsimd/types/xsimd_avx512_int_base.hpp
+++ b/include/xsimd/types/xsimd_avx512_int_base.hpp
@@ -114,43 +114,62 @@ namespace xsimd
 
     namespace avx512_detail
     {
-        template<class Tup, std::size_t... Is>
-        __m512i revert_args_set_epi8(Tup&& t, detail::index_sequence<Is...>)
+        inline __m512i int_init(std::integral_constant<std::size_t, 1>,
+                         int8_t t0, int8_t t1, int8_t t2, int8_t t3,
+                         int8_t t4, int8_t t5, int8_t t6, int8_t t7,
+                         int8_t t8, int8_t t9, int8_t t10, int8_t t11,
+                         int8_t t12, int8_t t13, int8_t t14, int8_t t15,
+                         int8_t t16, int8_t t17, int8_t t18, int8_t t19,
+                         int8_t t20, int8_t t21, int8_t t22, int8_t t23,
+                         int8_t t24, int8_t t25, int8_t t26, int8_t t27,
+                         int8_t t28, int8_t t29, int8_t t30, int8_t t31,
+                         int8_t t32, int8_t t33, int8_t t34, int8_t t35,
+                         int8_t t36, int8_t t37, int8_t t38, int8_t t39,
+                         int8_t t40, int8_t t41, int8_t t42, int8_t t43,
+                         int8_t t44, int8_t t45, int8_t t46, int8_t t47,
+                         int8_t t48, int8_t t49, int8_t t50, int8_t t51,
+                         int8_t t52, int8_t t53, int8_t t54, int8_t t55,
+                         int8_t t56, int8_t t57, int8_t t58, int8_t t59,
+                         int8_t t60, int8_t t61, int8_t t62, int8_t t63)
         {
-            // funny, this instruction is not yet implemented in clang or gcc (will come in future versions)
 #if defined(__clang__) || __GNUC__
             return __extension__ (__m512i)(__v64qi)
             {
-                static_cast<char>(std::get<Is>(std::forward<Tup>(t)))...
+              t0, t1, t2, t3, t4, t5, t6, t7, t8, t9, t10, t11, t12, t13, t14, t15,
+              t16, t17, t18, t19, t20, t21, t22, t23, t24, t25, t26, t27, t28, t29, t30, t31,
+              t32, t33, t34, t35, t36, t37, t38, t39, t40, t41, t42, t43, t44, t45, t46, t47,
+              t48, t49, t50, t51, t52, t53, t54, t55, t56, t57, t58, t59, t60, t61, t62, t63
             };
 #else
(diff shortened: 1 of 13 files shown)


serge-sans-paille:
@JohanMabille It's still a WIP, esp. wrt to the batch_constant / batch_bool_constant API, but early feedbacks are welcome.

include/xsimd/types/xsimd_avx512_int_base.hpp
```diff
@@ -114,45 +114,64 @@ namespace xsimd
...
+                         int8_t t24, int8_t t25, int8_t t26, int8_t t27,
+                         int8_t t28, int8_t t29, int8_t t30, int8_t t31,
+                         int8_t t32, int8_t t33, int8_t t34, int8_t t35,
+                         int8_t t36, int8_t t37, int8_t t38, int8_t t39,
+                         int8_t t40, int8_t t41, int8_t t42, int8_t t43,
+                         int8_t t44, int8_t t45, int8_t t46, int8_t t47,
+                         int8_t t48, int8_t t49, int8_t t50, int8_t t51,
+                         int8_t t52, int8_t t53, int8_t t54, int8_t t55,
+                         int8_t t56, int8_t t57, int8_t t58, int8_t t59,
+                         int8_t t60, int8_t t61, int8_t t62, int8_t t63)
```
JohanMabille, line 133: What is the benefit of this change?

include/xsimd/types/xsimd_avx512_int16.hpp
```diff
@@ -333,6 +333,12 @@ namespace xsimd
                 XSIMD_RETURN_MERGED_AVX(res_lo, res_hi);
             #endif
             }
+
+            template<bool... Values>
+            static batch_type select(const batch_bool_constant<value_type, Values...>& cond, const batch_type& a, const batch_type& b)
+            {
+              return select(cond(), a, b);
+            }
```
JohanMabille, line 341: Since the select method has the same implementation for all the batches of int, I wonder if it could be moved to `avx512_int_kernel_base` to avoid code duplication.

I understand we cannot have a generic implementation in `simd_base` since some implementation uses specific intrinsics (not the case for AVS512 though).

include/xsimd/types/xsimd_avx_int16.hpp
```diff
@@ -277,6 +277,12 @@ namespace xsimd
                 XSIMD_RETURN_MERGED_SSE(res_low, res_high);
 #endif
             }
+
+            template<bool... Values>
+            static batch_type select(const batch_bool_constant<value_type, Values...>& cond, const batch_type& a, const batch_type& b)
+            {
+              return select(cond(), a, b);
+            }
```
JohanMabille, line 285: Same remark as for avx512, maybe these methods can be moved to `avx_int_kernel_base`.

include/xsimd/types/xsimd_neon_int16.hpp
```diff
@@ -344,6 +344,12 @@ namespace xsimd
             {
                 return vbslq_s16(cond, a, b);
             }
+
+            template<bool... Values>
+            static batch_type select(const batch_bool_constant<value_type, Values...>& cond, const batch_type& a, const batch_type& b)
+            {
+                return select(cond(), a, b);
```
JohanMabille, line 351: Same remark with `neon_int_kernel_base`

include/xsimd/types/xsimd_sse_int16.hpp
```diff
@@ -192,6 +192,12 @@ namespace xsimd
                 return _mm_or_si128(_mm_and_si128(cond, a), _mm_andnot_si128(cond, b));
 #endif
             }
+
+            template<bool... Values>
+            static batch_type select(const batch_bool_constant<value_type, Values...>& cond, const batch_type& a, const batch_type& b)
+            {
+                return select(cond(), a, b);
+            }
```
JohanMabille, line 200: Same remark with `sse_int_kernel_base`

JohanMabille (commented):
Nitpicking (although I know thie PR is early stage):

- opening curly braces appear on their own line
-  we use 4 spaces indentation.

include/xsimd/types/xsimd_avx512_int_base.hpp
serge-sans-paille: A very simple one: it makes the code compile without hitting clang template instantiation limit. i think this path was not tested before :-)

serge-sans-paille:
@JohanMabille this should be reviewable now

include/xsimd/types/xsimd_base.hpp
```diff
@@ -1700,6 +1701,28 @@ namespace xsimd
...
+     * @param a batch values for truthy condition.
+     * @param b batch value for falsy condition.
+     * @return the result of the selection.
+     */
+    template <class X, bool... Masks>
+    inline batch_type_t<X> select(const batch_bool_constant<typename simd_batch_traits<X>::value_type, Masks...>& cond, const simd_base<X>& a, const simd_base<X>& b)
+    {
+        using value_type = typename simd_batch_traits<X>::value_type;
+        using kernel = detail::batch_kernel<value_type, simd_batch_traits<X>::size>;
+        return kernel::select(cond, a(), b());
```
JohanMabille, line 1723: Just to be sure I understand how it works: when the kernel for a given batch does not provide an overload of select accepting a `batch_bool_constant`, `cond` is first converted to the corresponding `batch_type` and then the regular `select` method is called?

JohanMabille (approved):

serge-sans-paille:
> Just to be sure I understand how it works: when the kernel for a given batch
> does not provide an overload of select accepting a batch_bool_constant, cond is
> first converted to the corresponding batch_type and then the regular select
> method is called?

yes

JohanMabille:
Brillant!

PR #446, opened 2021-03-29
Enable testing with clang on AVX512

serge-sans-paille:


.azure-pipelines/azure-pipelines-linux-clang.yml
```diff
@@ -29,6 +29,9 @@ jobs:
           llvm_version: '9'
         clang_10:
           llvm_version: '10'
+        clang_10_avx512:
+          llvm_version: '10'
+          avx512: 1
     pool:
```
JohanMabille, line 35: There's alreay a configuration for testing avx512 with gcc, so I'm not sure that adding a new one is required, not that it will actually work.

serge-sans-paille: gcc and clang have a different set of warning. But yeah, that doesn't seem to be the problem here.

.azure-pipelines/unix-build.yml
```diff
@@ -7,10 +7,6 @@ steps:
       if [[ $(enable_xtl_complex) == 1 ]]; then
         conda install xtl -c conda-forge
       fi
-      if [[ $(avx512) == 1 ]]; then
-        conda install gcc_linux-64 gxx_linux-64;
-        sh $(Build.SourcesDirectory)/install_sde.sh;
-      fi
```
JohanMabille, line 13: No need to remove everything, only the install of `gcc_linux-64` and `gxx_linux-64`.

.azure-pipelines/unix-build.yml
```diff
@@ -52,11 +48,10 @@ steps:
   - script: |
       source activate xsimd
       if [[ $(avx512) == 1 ]]; then
-        ../../sde-external-8.56.0-2020-07-05-lin/sde64 -skx -- ./test_xsimd ;
+        qemu-x86_64 ./test_xsimd ;
       elif [[ $(cross_compile) == 1 ]]; then
```
JohanMabille, line 52: This should be kept

JohanMabille (changes requested):

.azure-pipelines/unix-build.yml
```diff
@@ -7,9 +7,8 @@ steps:
       if [[ $(enable_xtl_complex) == 1 ]]; then
         conda install xtl -c conda-forge
       fi
-      if [[ $(avx512) == 1 ]]; then
-        conda install gcc_linux-64 gxx_linux-64;
-        sh $(Build.SourcesDirectory)/install_sde.sh;
+      if test "x$(avx512)" = "x1" ; then
```
JohanMabille, line 10: What is the benefit of this test compared to the previous one?

serge-sans-paille: That's an old idiom, it's a reflex: https://stackoverflow.com/questions/6852612/test-for-empty-string-with-x

JohanMabille (approved):

.azure-pipelines/unix-build.yml
JohanMabille: Ok, thanks for the explanation!

PR #447, opened 2021-03-29
Fix avx512 CI configuration

JohanMabille:


JohanMabille:
Removing the install of gcc from conda-forge enabled AVX512 tests. This will be done in #446, closing as duplicate.

Issue #314, opened 2019-10-10
Saturated add and subtract seems to be missing.


schrpe:
Do you have any guidelines how one would go and implement saturated addition and subtraction?

ThomasRetornaz:
+1
It could be very usefull for image processing library.
Same comment than Peter, if you have guidelines?
Regards
TR

wolfv:
probably making a free function like `xsimd::sadd(a, b)` would be the way to go.

Then you could implement it for the different architectures in the kernels as a xsimd::kernel::sadd(..).

And in the last step you can use the function which is implemented in the kernel inside the `xsimd::sadd` function. 

For example, here is the kernel implementation for AVX int32 for add:

https://github.com/xtensor-stack/xsimd/blob/38eed51285edad0dd75b552c99cdd8178565e24d/include/xsimd/types/xsimd_avx_int32.hpp#L200-L207

Now, you could just copy-paste that part and call it `sadd`, for a start.
There is some template magic to select the appropriate kernel. Your function might then look like:

```cpp
template <class T>
auto sadd(xsimd::simd_base<T>& lhs, xsimd::simd_base<T>& rhs) {
        using value_type = typename simd_batch_traits<X>::value_type;
        using kernel = detail::batch_kernel<value_type, simd_batch_traits<X>::size>;
        return kernel::sadd(lhs(), rhs());
}
```

would be great if you give it a shot! We're happy to assist further :)

ThomasRetornaz:
i'll give it a try

serge-sans-paille:
[`sadd`](https://xsimd.readthedocs.io/en/latest/api/arithmetic_index.html#_CPPv4I00E4saddDTpl1x1yERK5batchI1T1AERK5batchI1T1AE) and [`ssub`](https://xsimd.readthedocs.io/en/latest/api/arithmetic_index.html#_CPPv4I00E4ssubDTmi1x1yERK5batchI1T1AERK5batchI1T1AE) are now implemented for all supported targets,

PR #454, opened 2021-04-14
Ref 314

serge-sans-paille:
This is a recommit of #314 rebased with some code cleanup and commit split, and hopefully a few bug fixes to come


include/xsimd/types/xsimd_avx512_int32.hpp
```diff
@@ -176,6 +176,45 @@ namespace xsimd
                 return _mm512_sub_epi32(lhs, rhs);
             }
 
+            static batch_type sadd(const batch_type& lhs, const batch_type& rhs)
```
serge-sans-paille, line 179: I'll benchmark that approach compared to the one based on a comparison + a blend.

serge-sans-paille: Actually, found a nice and efficient solution based on min/max \o/

serge-sans-paille:
@JohanMabille ready for review. I did some benchmarking of the generic adds method, applying it to `int16_t` where we have a reference intrinsic. the home made implementation still acheieves ~3Instruction per cycle, so no pipeline stall. It's roughly 4 time slower than the intrinsic, which is somehow expected. And 4 time faster than the sequential naive version, which is good too :-)

include/xsimd/types/xsimd_utils.hpp
```diff
@@ -397,6 +397,63 @@ namespace xsimd
...
+            }
+            else
+            {
+                ret=(T)(lhs - rhs);
+            }
+
+        }
+        return ret;
+    }
+
```
JohanMabille, line 456: Why having these functions in `xsimd_utils.hpp` instead of `xsimd_fallback.hpp` ?

serge-sans-paille: Correct. I'll update that.

serge-sans-paille:
@JohanMabille ready for another round of review :-)

JohanMabille (dismissed):

include/xsimd/math/xsimd_scalar.hpp
```diff
@@ -453,6 +481,58 @@ namespace xsimd
...
+                return std::numeric_limits<T>::lowest();
+            }
+            else
+            {
+                return lhs - rhs;
+            }
+
+        }
+    }
+
```
JohanMabille, line 535: Do you think it would make sense to have `sadd` and `ssub` in `xsimd_fallback.hpp` and have `xsimd_scalar.hpp` depending on it? The idea is to have the `math` folder depenting on the `types` folder and to avoid cyclic dependencies between folders. Even if it's not a big deal at the folder level, it helps avoiding circular include at the header level.

If there are many functions like this (or more to come), it could even make sense to create an implmeentation file in `types` on which both `xsimd_fallback.hpp` and `xsimd_scalar.hpp` depend.

serge-sans-paille: > If there are many functions like this (or more to come), it could even make sense to create an implementation file in types on which both xsimd_fallback.hpp and xsimd_scalar.hpp depend.

I second that. We already have `ipow`, `sadd`, `ssub` ... which name do you suggest?

JohanMabille: xsimd/types/xsimd_math_utils.hpp or xsimd/types/xsimd_basic_math.hpp or xsimd/types/xsimd_common_math.hpp ?

serge-sans-paille:
@JohanMabille reen and cleaned-up o/

include/xsimd/math/xsimd_math_utils.hpp
```diff
@@ -0,0 +1,104 @@
+/***************************************************************************
+* Copyright (c) Johan Mabille, Sylvain Corlay, Wolf Vollprecht and         *
+* Martin Renou                                                             *
+* Copyright (c) QuantStack                                                 *
+*                                                                          *
+* Distributed under the terms of the BSD 3-Clause License.                 *
+*                                                                          *
+* The full license is in the file LICENSE, distributed with this software. *
+****************************************************************************/
```
JohanMabille, line 9: I think this file should live in the `types` subfolder instead of `math` to guarantee the non cyclic dependency math -> types

JohanMabille (changes requested):

include/xsimd/math/xsimd_math_utils.hpp
serge-sans-paille: This file is only included from the math folder, so I think it's the right place.

include/xsimd/types/xsimd_fallback.hpp
```diff
@@ -16,6 +16,8 @@
 #include <cmath>
 #include <utility>
 
+#include "xsimd/math/xsimd_scalar.hpp"
```
JohanMabille, line 19: @serge-sans-paille It's included here too.

JohanMabille:
Awesome!

JohanMabille (approved):

PR #455, opened 2021-04-15
Test and fix min/max implementation on uint32_t for AVX2

serge-sans-paille:


serge-sans-paille:
All green, cc @JohanMabille

JohanMabille:
Great! thank you!

PR #457, opened 2021-04-18
Fix segmentation fault on mingw

mattn:
On Windows, using msys2/mingw64 compiler, examples always crash with SEGV on ___chkstk_ms. This can be avoid with making static variable.

### DIFF
--- a/examples/mandelbrot.cpp
+++ b/examples/mandelbrot.cpp
@@ -229,7 +229,7 @@ int main()
     const float y1            = 1;
     const int maxIters        = 256;
 
-    alignas(64) std::vector<int> buf(width * height);
+    std::vector<int, xsimd::aligned_allocator<int, XSIMD_DEFAULT_ALIGNMENT>> buf(width * height);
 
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
Thanks for spotting this! What do you think of allocating that array on the heap explicitly instead of using static storage?

mattn:
Thanks.

examples/mandelbrot.cpp
```diff
@@ -228,7 +229,7 @@ int main()
     const float y1            = 1;
     const int maxIters        = 256;
 
-    alignas(64) std::array<int, width * height> buf;
+    alignas(64) std::vector<int> buf(width * height);
```
serge-sans-paille, line 232: `alignas(64)` does not affect heap allocation here, so you need to use

```
std::vector<int, xsimd::aligned_allocator<int, XSIMD_DEFAULT_ALIGNMENT>>;
```

instead

mattn: Thank you

serge-sans-paille:
Great! Thanks for your patience @mattn

PR #459, opened 2021-04-22
Add extract_pair for shifting 128bit data

guyuqi:
Extracts the lowest vector elements from the second source and the highest vector
elements from the first source. The index: 'n' specifies the lowest vector element
to extract from the first source register.
Concatenates the results into th Return value.

### DIFF
--- a/include/xsimd/types/xsimd_fallback.hpp
+++ b/include/xsimd/types/xsimd_fallback.hpp
@@ -967,7 +967,7 @@ namespace xsimd
             static batch_type extract_pair(const batch_type& lhs, const batch_type& rhs, const int n)
             {
                 batch_type b_concatenate;
-                for (int i = 0 ; i < (N - n); ++i)
+                for (int i = 0 ; i < static_cast<int>(N - n); ++i)
                 {
                     b_concatenate[i] = lhs[i + n];
                     if(i < n)
--- a/include/xsimd/types/xsimd_fallback.hpp
+++ b/include/xsimd/types/xsimd_fallback.hpp
@@ -962,6 +962,21 @@ namespace xsimd
                 }
                 return b_hi;
             }
+
+            /* 0 <= n <= N/2 */
+            static batch_type extract_pair(const batch_type& lhs, const batch_type& rhs, const int n)
+            {
+                batch_type b_concatenate;
+                for (int i = 0 ; i < (N - n); ++i)
+                {
+                    b_concatenate[i] = lhs[i + n];
+                    if(i < n)
+                    {
+                        b_concatenate[N - 1 - i] = rhs[n - 1 - i];
+                    }
+                }
+                return b_concatenate;
+            }
         };
     }
 
--- a/include/xsimd/types/xsimd_neon_uint8.hpp
+++ b/include/xsimd/types/xsimd_neon_uint8.hpp
@@ -349,6 +349,12 @@ namespace xsimd
                 return vcombine_u8(tmp.val[0], tmp.val[1]);
 #endif
             }
+
+            static batch_type extract_pair(const batch_type& lhs, const batch_type& rhs, const int n)
+            {
+                return vextq_u8(lhs, rhs, n);
+            }
+
         };
     }
 
--- a/include/xsimd/types/xsimd_neon_uint8.hpp
+++ b/include/xsimd/types/xsimd_neon_uint8.hpp
@@ -352,7 +352,13 @@ namespace xsimd
 
             static batch_type extract_pair(const batch_type& lhs, const batch_type& rhs, const int n)
             {
-                return vextq_u8(lhs, rhs, n);
+                switch(n)
+                {
+                    case 0: return lhs;
+                    XSIMD_REPEAT_16_v2(vextq_u8);
+                    default: break;
+                }
+                return batch_type(uint8_t(0));
             }
 
         };
--- a/include/xsimd/types/xsimd_sse_int8.hpp
+++ b/include/xsimd/types/xsimd_sse_int8.hpp
@@ -237,7 +237,20 @@ namespace xsimd
 
             static batch_type extract_pair(const batch_type& lhs, const batch_type& rhs, const int n)
             {
+#if XSIMD_X86_INSTR_SET >= XSIMD_X86_SSSE3_VERSION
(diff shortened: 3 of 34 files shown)


guyuqi:
Let me fix CI issues first.

guyuqi:
A little bit confused about the failed log in CI:
```
/home/vsts/work/1/s/include/xsimd/types/xsimd_neon_uint8.hpp:355:24: error: argument to '__builtin_neon_vextq_v' must be a constant integer
2021-04-27T09:53:57.3826864Z                 return vextq_u8(lhs, rhs, n);
```
which is defined in:
```
            static batch_type extract_pair(const batch_type& lhs, const batch_type& rhs, const int n)
            {
                return vextq_u8(lhs, rhs, n);
            }

```
`n` is a constant integer

JohanMabille:
I think the compiler expects a constant known at build time.

include/xsimd/types/xsimd_sse_int8.hpp
```diff
@@ -234,6 +234,25 @@ namespace xsimd
             {
                 return _mm_unpackhi_epi8(lhs, rhs);
             }
+
+            static batch_type extract_pair(const batch_type& lhs, const batch_type& rhs, const int n)
+            {
+#if XSIMD_X86_INSTR_SET >= XSIMD_X86_SSSE3_VERSION
+                return _mm_alignr_epi8(rhs, lhs, n);
+#else
```
serge-sans-paille, line 242: looks like you could use a shuffle instead of this sequential fallback, right?

include/xsimd/types/xsimd_neon_uint8.hpp
```diff
@@ -349,6 +349,13 @@ namespace xsimd
                 return vcombine_u8(tmp.val[0], tmp.val[1]);
 #endif
             }
+
+            static batch_type extract_pair(const batch_type& lhs, const batch_type& rhs, const int n)
+            {
+                const int index = n;
```
serge-sans-paille, line 355: this assignment looks odd to me ^^!

include/xsimd/types/xsimd_fallback.hpp
```diff
@@ -962,6 +962,21 @@ namespace xsimd
                 }
                 return b_hi;
             }
+
+            /* 0 <= n <= N/2 */
+            static batch_type extract_pair(const batch_type& lhs, const batch_type& rhs, const int n)
+            {
```
serge-sans-paille, line 968: there should be an assert on the value of `n`

serge-sans-paille (changes requested):
You also need to a doc entry.

include/xsimd/types/xsimd_neon_uint8.hpp
guyuqi: Yep, I added it to try to fix build issue as below which was not reproduced in my local: environment.: 
```
/home/vsts/work/1/s/include/xsimd/types/xsimd_neon_uint8.hpp:355:24: error: argument to '__builtin_neon_vextq_v' must be a constant integer
2021-04-27T09:53:57.3826864Z                 return vextq_u8(lhs, rhs, n);
```
But it also doesn't work.

serge-sans-paille:
There are a lot of sequentiel implementation in you code (e.g. for avx512). Why not using a shuffle?

guyuqi:
@serge-sans-paille  Sorry for the late reply. 


> You also need to a doc entry.

Could you kindly please explain what 'doc entry' means?
Does it mean to add comments/document here to describe the implemetation details? 



> There are a lot of sequential implementation in you code (e.g. for avx512). Why not using a shuffle?


This PR is mainly to implemented for 128bit data-width (SSE/Neon).
It also implemented test cases just for Neon/SSE(128bit).

For AVX2/512 cases, I partly leveraged the directly available intrinsics of AVX2/512 and used a lot of sequential implementation for the other data-width cases which are not available by Intel Intrinsics.

For the 256 and 512 data-width:
```
x86: AVX2/AVX512
Arm64: SVE
```
How about to completely implement them in another PR?

serge-sans-paille:
Thanks @guyuqi for the updates. I like the switch approach -  the compiler will fold the result if the paramater is a constant and it gives some extra flexibility. I'm wondering why you're still fallingback to the sequential implementation in many case, where a shuffle would do the trck (if I understand correctly) ?

guyuqi:
> Thanks @guyuqi for the updates. I like the switch approach - the compiler will fold the result if the paramater is a constant and it gives some extra flexibility. I'm wondering why you're still fallingback to the sequential implementation in many case, where a shuffle would do the trck (if I understand correctly) ?

Thanks for your comments. @serge-sans-paille 

The reasons for fallback sequential implementation are that:

1. For Intel AXV2 256 bit and AVX512:
It just supports epi8, epi32 and epi64 and lacks support of `short`, `float `and `double` for `alignr`.
```
_mm256_alignr_epi8
_mm256_alignr_epi32
_mm256_alignr_epi64
```
```
_mm512_alignr_epi8
_mm512_alignr_epi32
_mm512_alignr_epi64
```
For `short`, `float `and `double`, we might not get the good performance to hack the function of `alignr` to implement data shifting from a pair of vectors by leveraging some other fundamental Intel simd intrinsics (storeu, lddqu, ... ).

2. This PR is for shifting 128bit data (SSE/Neon).
It also implemented test cases just for Neon/SSE(128bit).
These fallback sequential implementations are also just used for compiling under cases of AVX2/AVX512 .
And for the upcoming Arm64 SVE(256bit +),  how about to completely implement them in another PR (for 256bit and 512bit)? 
There might be more efforts to SVE features implementation, 
and performance tunning for hacking `alignr` by Intel simd intrinsics mentioned aboved.

serge-sans-paille:
@JohanMabille I don't know if we are fine with a partially implemented PR, what's your opinion? I'd say that we should at least add a FIXME on each function with a sequential implementation, or I can base a new PR on this one and add the missing pieces?

JohanMabille:
I think it's fine to have a partially implemented PR given the amount of changes a complete PR would give. We already did that in the past. But then we need to finish the implementation quickly (before the next release).

guyuqi:
@serge-sans-paille 
All tests are passed on Arm64 and x86.

serge-sans-paille:
Thanks for addressing all the errors @guyuqi

PR #461, opened 2021-04-26
Improve aligned allocator implementation

serge-sans-paille:
Detect invalid alignment under debug mode.
Use platform allocator.


JohanMabille:
Thanks!

PR #462, opened 2021-04-26
Make it possible to test fallback implementation on non-x86, non-arm …

serge-sans-paille:
…system

This simplifies the work for packager who can then package and run tests for
xsimd on any platform, whether it's supported or not.


JohanMabille:
Great!

PR #463, opened 2021-04-29
detect sincos() availability at configure time

veprbl:
The `__sincosf` is a part of Apple SDK and may not be available in an arbitrary toolchain for C++. This switches to compile-time detection of intrinsics presence.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,6 +118,37 @@ target_include_directories(xsimd INTERFACE
 
 target_compile_features(xsimd INTERFACE cxx_std_11)
 
+include(CheckCXXSourceCompiles)
+check_cxx_source_compiles(
+    "
+    #include <cmath>
+    int main() {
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
> people not using cmake

Who is that? ;)

serge-sans-paille:
I think I have another approach, trait based, that would be pure C++, let me investigate that :-)

serge-sans-paille:
What about something along these lines: https://gcc.godbolt.org/z/Y33MaGxzz

JohanMabille:
@veprbl Thanks for tackling this!

I think a pure C++ detection with metaprogramming (enable_if / tag dispatching, whatever can do the job) would be preferable. CMake should be used for building only.

serge-sans-paille:
Alternate implementation in https://github.com/xtensor-stack/xsimd/pull/465

JohanMabille:
Superseded with #465

PR #465, opened 2021-04-30
Implement a more robust detection for sincos / __sincos function family

serge-sans-paille:


serge-sans-paille:
@veprbl  can you double check this works in your setup?

veprbl (approved):
Compiles with gcc and clang on both x86_64-linux and x86_64-darwin

JohanMabille:
Awesome! Thanks @serge-sans-paille and @veprbl !

PR #466, opened 2021-05-02
Feature/explicit instruction sets

serge-sans-paille:


serge-sans-paille:
@JohanMabille ready for review!

serge-sans-paille:
Note to self: I need to update the doc

PR #467, opened 2021-05-06
Added missing arch

JohanMabille:


include/xsimd/config/xsimd_arch.hpp
```diff
@@ -1,114 +1,175 @@
...
+        struct fma4 : avx
+        {
+        };
+
+        // AMD specific (very few old processors)
+        struct xop : fma4
+        {
+        };
+
+        struct avx2
```
JohanMabille, line 78: @serge-sans-paille should avx2 inherit from avx?

serge-sans-paille: I'd say.

serge-sans-paille:
apart from the inheritance nit, looks good to me.

PR #481, opened 2021-05-30
Feature/xsimd refactoring on going

serge-sans-paille:


serge-sans-paille:
Grrrrrrrrreeeeeen :-) cc @JohanMabille
I'll keep adding commits to that branch

PR #485, opened 2021-06-02
Add extra namespace in Basic Usage example

nilsdeppe:
Thanks for the great library!

Add the `arch` namespace needed to get access to `avx::batch`.

It looks like the README uses `batch<double, 4>`. If that's preferred I can update the PR to do that instead.


JohanMabille:
Thanks for this fix. Both usages are possible (either with the size or with through the arch namespace). A huge refactoring is on-going so that users can specify explicitly the architectur instead of the size.

PR #489, opened 2021-06-13
Supress constructor from pointer

serge-sans-paille:


serge-sans-paille:
cc @JohanMabille that's the change we've been discussing recently. It makes `xsimd::zero<T>()` useless.

include-refactoring/xsimd/arch/xsimd_sse2.hpp
```diff
@@ -29,6 +29,23 @@ namespace xsimd {
       return _mm_add_pd(self, other);
     }
 
+    // all
+    template<class A> bool all(batch<float, A> const& self, requires<sse2>) {
+      return _mm_movemask_ps(self) == 0x0F;
+    }
```
JohanMabille, line 35: `_mm_movemask_ps` is available with SSE instruction set, therefore this should be moved back to `xsimd_sse.hpp`.

include-refactoring/xsimd/arch/xsimd_sse2.hpp
```diff
@@ -29,6 +29,23 @@ namespace xsimd {
...
+    template<class A> bool all(batch<double, A> const& self, requires<sse2>) {
+      return _mm_movemask_pd(self) == 0x03;
+    }
+
+    // any
+    template<class A> bool any(batch<float, A> const& self, requires<sse2>) {
+      return _mm_movemask_ps(self) != 0;
```
JohanMabille, line 42: Same remark here

PR #490, opened 2021-06-16
Convert bytes/shorts/words/longs to batch_vector

guyuqi:
Set data array of int8, int16, int32 and int64 to simd batch vector.

### DIFF
--- a/include/xsimd/types/xsimd_avx512_conversion.hpp
+++ b/include/xsimd/types/xsimd_avx512_conversion.hpp
@@ -52,7 +52,7 @@ namespace xsimd
      *  to batch functions                    *
      ******************************************/
 
-    void bytes_to_vector(batch<uint8_t, 64>& vec,
+    void bytes_to_batch_vector(batch<uint8_t, 64>& vec,
                          int8_t b63, int8_t b62, int8_t b61, int8_t b60,
                          int8_t b59, int8_t b58, int8_t b57, int8_t b56,
                          int8_t b55, int8_t b54, int8_t b53, int8_t b52,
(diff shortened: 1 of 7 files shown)


guyuqi:
Implement convertsion of bytes/shorts/words/longs to batch_vector which is needed by Arrow.
The related test cases were also added. 
The all CI cases are passed. @serge-sans-paille

guyuqi:
@serge-sans-paille  Could you please take a look at this PR ? thanks.

include/xsimd/types/xsimd_avx512_conversion.hpp
```diff
@@ -281,6 +281,113 @@ namespace xsimd
...
+                         int8_t b39, int8_t b38, int8_t b37, int8_t b36,
+                         int8_t b35, int8_t b34, int8_t b33, int8_t b32,
+                         int8_t b31, int8_t b30, int8_t b29, int8_t b28,
+                         int8_t b27, int8_t b26, int8_t b25, int8_t b24,
+                         int8_t b23, int8_t b22, int8_t b21, int8_t b20,
+                         int8_t b19, int8_t b18, int8_t b17, int8_t b16,
+                         int8_t b15, int8_t b14, int8_t b13, int8_t b12,
+                         int8_t b11, int8_t b10, int8_t b9, int8_t b8,
+                         int8_t b7, int8_t b6, int8_t b5, int8_t b4,
+                         int8_t b3, int8_t b2, int8_t b1, int8_t b0);
```
JohanMabille, line 307: - I think the API could accept an `std::array<int8_t>` instead of 64 input arguments. This would not require a huge change on the client side: `bytes_to_batch_vector(my_batch, {b63, b62, .... b0});` and would make the code easier to maintain.
- Regarding the name, what about `array_to_batch`, or `cast_to_batch`? You coud even add a template parameter for the batch type (even if only implemented for `unit8_t for now). The reason is it is better to rely on overloading than trying to have information about the types in the function name.

JohanMabille (changes requested):
Sorry for this very late review!

include/xsimd/types/xsimd_avx512_conversion.hpp
guyuqi: @JohanMabille  Agree with you. Let me refine it. 
Thanks for your comments and suggestions.

guyuqi:
The new PR#539 is to refactor the array_to_batch for xsimd was switched to new implementation, 
So close it.

PR #491, opened 2021-06-16
Support for NEON instruction set

JohanMabille:


include-refactoring/xsimd/arch/xsimd_arm7.hpp
```diff
@@ -0,0 +1,58 @@
...
+    namespace kernel
+    {
+        using namespace types;
+
+        namespace detail
+        {
+            template <class... T>
+            struct arm_dispatcher
+            {
+                using container_type = std::tuple<T (*)(T, T) ...>;
```
serge-sans-paille, line 20: note: you may want to make it even more generic for unary functions etc :-)

include-refactoring/xsimd/arch/xsimd_arm7.hpp
```diff
@@ -0,0 +1,58 @@
...
+                const container_type m_func;
+
+                template <class U>
+                U run(U lhs, U rhs) const
+                {
+                    using func_type = U (*)(U, U);
```
serge-sans-paille, line 26: this definition would be used in the container_type -> put it in a using definition at struct scope

include-refactoring/xsimd/arch/xsimd_arm7.hpp
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        template <class T, class A, class = typename std::enable_if<std::is_integral<T>::value, void>::type>
+        batch<T, A> add(batch<T, A> const& lhs, batch<T, A> const& rhs, requires<arm7>)
+        {
+            std::cout << "coincoin" << std::endl;
+            using register_type = typename batch<T, A>::register_type;
+            constexpr detail::arm_dispatcher_type dispatcher =
+            {
+                std::make_tuple(vaddq_u8, vaddq_s8, vadd_qu16, vaddq_s16, vaddq_u32, vaddq_s32, vaddq_u64, vaddq_s64);
```
serge-sans-paille, line 45: I wonder if a constructor in arm_dispatcher could make that line even simpler?

include-refactoring/xsimd/arch/xsimd_sse2.hpp
```diff
@@ -3,9 +3,6 @@
 
 #include "../types/xsimd_sse2_register.hpp"
 
-#include <emmintrin.h>
```
serge-sans-paille, line 6: thanks for cleaning this

include-refactoring/xsimd/config/xsimd_config.hpp
```diff
@@ -64,4 +64,25 @@
...
+    #if __ARM_ARCH >= 7
+        #define XSIMD_WITH_ARM7 1
+    #else
+        #define XSIMD_WITH_ARM7 0
+    #endif
+
+    #if __ARM_ARCH >= 8
+        #ifdef __aarch64__
+            #define XSIMD_WITH_ARM8_64 1
+            #define XSIMD_WITH_ARM8_32 1
```
serge-sans-paille, line 77: you could move that one outside of the ` #ifdef __aarch64__`

include-refactoring/xsimd/config/xsimd_cpuid.hpp
```diff
@@ -1,139 +1,151 @@
 #ifndef XSIMD_CPUID_HPP
 #define XSIMD_CPUID_HPP
 
-#include "../types/xsimd_all_registers.hpp"
 #include <cstring>
 #include <algorithm>
 #include <array>
 
+#include "../types/xsimd_all_registers.hpp"
```
serge-sans-paille, line 8: I tend to include our header first, it helps detecting missing dependencies

include-refactoring/xsimd/config/xsimd_cpuid.hpp
```diff
@@ -1,139 +1,151 @@
...
+                 arm8_64 = 0;
+                 best = arm8_32::version();
+#else
+                 arm7 = 1;
+                 arm8_32 = 0;
+                 arm8_64 = 0;
+                 best = arm7::version();
+#endif
+//#elif defined(__ARM_NEON) || defined(_M_ARM)
+//                 neon = bool(getauxval(AT_HWCAP) & HWCAP_NEON);
```
serge-sans-paille, line 58: why did you remove the runtime detection code?

include-refactoring/xsimd/types/xsimd_all_registers.hpp
```diff
@@ -1,8 +1,12 @@
-#include "../types/xsimd_sse_register.hpp"
-#include "../types/xsimd_sse2_register.hpp"
-#include "../types/xsimd_sse3_register.hpp"
-#include "../types/xsimd_sse4_1_register.hpp"
-#include "../types/xsimd_sse4_2_register.hpp"
-#include "../types/xsimd_avx_register.hpp"
-#include "../types/xsimd_avx2_register.hpp"
-#include "../types/xsimd_avx512f_register.hpp"
+#include "xsimd_sse_register.hpp"
```
serge-sans-paille, line 1: or `'./'` to make it explicit it's a local header?

include-refactoring/xsimd/types/xsimd_arm7_register.hpp
```diff
@@ -0,0 +1,94 @@
...
+            using arm_vector_type = typename std::conditional<std::is_signed<T>::value,
+                                                              signed_arm_vector_type<T>,
+                                                              unsigned_arm_vector_type<T>
+                                                             >::type;
+
+            using char_arm_vector_type = typename std::conditional<std::is_signed<char>::value,
+                                                                   signed_arm_vector_type<char>,
+                                                                   unsigned_arm_vector_type<char>
+                                                                  >::type;
+        }
```
serge-sans-paille, line 74: neat!

include-refactoring/xsimd/types/xsimd_avx512f_register.hpp
```diff
@@ -3,23 +3,24 @@
...
-  struct avx512f : generic {
-    static constexpr bool supported() { return XSIMD_WITH_AVX512F; }
-    static constexpr bool available() { return true; }
-    static constexpr unsigned version() { return generic::version(3, 1, 0); }
-    static constexpr std::size_t alignment() { return 64; }
-  };
+namespace xsimd
+{
+    struct avx512f : generic
+    {
```
serge-sans-paille, line 9: With this reindenting, you're taking the risk of a difficult merge :-)

serge-sans-paille: Plus it makes the review more difficult for me :'(

include-refactoring/xsimd/arch/xsimd_arm7.hpp
JohanMabille: It would avoid the `std::make_tuple` call.

include-refactoring/xsimd/config/xsimd_cpuid.hpp
JohanMabille: It's just for testing for now, I'm not sure how we should fill the different fields. There were only neon and neon64 while we have arm7, arm8_32 and arm8_64 registers. But arm8_32 is supposed to be fully compatible with arm7, so I'm don't know if we should keep both of simplify (although I find keeping both more explicit).

include-refactoring/xsimd/types/xsimd_all_registers.hpp
JohanMabille: In other projects of the xtensor-stack, we use `<>` for standard headers and "package/headers" for dependencies, so it's usually clear that an include with the file name only means it's a local heder. But both are fine for me.

include-refactoring/xsimd/types/xsimd_avx512f_register.hpp
JohanMabille: arf sorry for that, I could not resist! I did it only on the register files, which re very small and unlikely to change a lot. I won't do it on other files.

JohanMabille:
@serge-sans-paille this is ready for a review ;)

include-refactoring/xsimd/arch/xsimd_arm7.hpp
```diff
@@ -0,0 +1,1942 @@
...
+                                                                             uint16x8_t, int16x8_t,
+                                                                             uint32x4_t, int32x4_t,
+                                                                             float32x4_t>;
+
+            /**************************************
+             * enabling / disabling metafunctions *
+             **************************************/
+
+            template <class T>
+            using enable_integral_t = typename std::enable_if<std::is_integral<T>::value, int>::type;
```
serge-sans-paille, line 152: out of curiosity: why `int` and not `void` in the `enable_if` ?

include-refactoring/xsimd/arch/xsimd_avx512f.hpp
```diff
@@ -280,11 +280,11 @@ namespace xsimd {
     }
 
     template<class A> batch<float, A> bitwise_not(batch<float, A> const& self, requires<avx512f>) {
-      return _mm256_xor_ps(self, _mm256_castsi256_ps(_mm256_set1_epi32(-1)));
+      return _mm512_xor_ps(self, _mm512_castsi512_ps(_mm512_set1_epi32(-1)));
```
serge-sans-paille, line 283: wow, did you find that one when adding new tests?

include-refactoring/xsimd/arch/xsimd_isa.hpp
```diff
@@ -53,9 +53,20 @@
 #include "./xsimd_avx512bw.hpp"
 #endif
 
+#if XSIMD_WITH_ARM7
```
serge-sans-paille, line 56: could be `#if XSIMD_WITH_ARM7 || XSIMD_WITH_ARM8_32` then

include-refactoring/xsimd/config/xsimd_arch.hpp
```diff
@@ -91,15 +93,19 @@ namespace xsimd {
         : join<typename Arch::template extend<Archs...>, Args...> {};
   } // namespace detail
 
-  using all_x86_architectures = arch_list<avx512bw, avx512dq, avx512cd, avx512f, fma5, avx2, /*xop, fma4,*/ avx, fma3, sse4_2, sse4_1, /*sse4a,*/ ssse3, sse3, sse2, sse>;
-  //using all_arm_architectures = arch_list<neon64, neon>;
-  using all_architectures = typename detail::join</*all_arm_architectures,*/ all_x86_architectures>::type;
+  struct unsupported {};
+  using all_x86_architectures = arch_list<avx512bw, avx512dq, avx512cd, avx512f, fma5, avx2, /*fma3, xop, fma4,*/ avx, sse4_2, sse4_1, /*sse4a, ssse3,*/ sse3, sse2, sse>;
```
serge-sans-paille, line 97: whoopsie: looks like we should uncomment `sse4a, ssse3` in that list

include-refactoring/xsimd/config/xsimd_config.hpp
```diff
@@ -105,4 +105,28 @@
...
 
+#ifdef __ARM_NEON
+    #if __ARM_ARCH >= 7
+        #define XSIMD_WITH_ARM7 1
+    #else
+        #define XSIMD_WITH_ARM7 0
+    #endif
+
+    #if __ARM_ARCH >= 8
+        #ifdef __aarch64__
```
serge-sans-paille, line 116: Should we use `_M_ARM64` here too?

include-refactoring/xsimd/types/xsimd_api.hpp
```diff
@@ -212,13 +212,6 @@ B bitwise_cast(batch<T, A> const& x) {
...
 }
 
-/**
- * @ingroup simd_batch_bitwise
- *
- * Computes the bitwise not of batch \c x.
- * @param x batch involved in the operation.
- * @return the result of the bitwise not.
- */
 template<class T, class A>
```
serge-sans-paille, line 215: Why this doc removing?

include-refactoring/xsimd/types/xsimd_api.hpp
```diff
@@ -1541,8 +1549,8 @@ bool all(batch_bool<T, A> const& x) {
  * @return a boolean scalar.
  */
 template<class T, class A>
-bool any(batch_bool<T, A> const& x) {
-  return kernel::any<A>(x, A{});
+bool any(batch_bool<T, A> const& self) {
```
serge-sans-paille, line 1552: please keep `x` so that it matches the doxygen

include-refactoring/xsimd/types/xsimd_api.hpp
```diff
@@ -1528,8 +1536,8 @@ batch<T, A> zip_lo(batch<T, A> const& x, batch<T, A> const& y) {
  * @return a boolean scalar.
  */
 template<class T, class A>
-bool all(batch_bool<T, A> const& x) {
-  return kernel::all<A>(x, A{});
+bool all(batch_bool<T, A> const& self) {
```
serge-sans-paille, line 1539: please keep `x` so that it matches the doxygen

include-refactoring/xsimd/types/xsimd_batch.hpp
```diff
@@ -180,6 +181,11 @@ struct batch_bool : types::get_bool_simd_register_t<T, A> {
   template<size_t... Is>
   batch_bool(bool const* data, detail::index_sequence<Is...>);
 
+  template <class U, class... V, size_t I, size_t... Is>
+  static register_type make_register(detail::index_sequence<I, Is...>, U u, V... v);
```
serge-sans-paille, line 185: Why not keeping the base type as a POD type? It would make this kind of function useless

include-refactoring/xsimd/types/xsimd_utils.hpp
```diff
@@ -89,6 +89,18 @@ namespace xsimd
     template <class T>
     using as_unsigned_integer_t = typename as_unsigned_integer<T>::type;
 
+    /*********************
+     * as_signed_integer *
+     *********************/
+
+    template <class T>
+    struct as_signed_integer : std::make_signed<T>
```
serge-sans-paille, line 97: this could be a `using directive`

include-refactoring/xsimd/types/xsimd_utils.hpp
```diff
@@ -323,7 +381,127 @@ namespace xsimd
...
+#define XSIMD_REPEAT_32_N_v2(op, addx)   \
+    XSIMD_REPEAT_16_N_v2(op, 0 + addx);  \
+    XSIMD_REPEAT_16_N_v2(op, 16 + addx);
+
+#define XSIMD_REPEAT_32_v2(op)           \
+    XSIMD_REPEAT_32_0_v2(op, 0);
 
+#define XSIMD_REPEAT_64_v2(op)           \
+    XSIMD_REPEAT_32_0_v2(op, 0);         \
+    XSIMD_REPEAT_32_N_v2(op, 32);
```
serge-sans-paille, line 504: The return of the macro... Why????

serge-sans-paille: Please use the trick I've been using for Intel arch instead

test-refactoring/test_utils.hpp
```diff
@@ -35,7 +35,47 @@ class simd_test_names
...
+        if (std::is_same<value_type, float>::value) { return prefix + "float"; }
+        if (std::is_same<value_type, double>::value) { return prefix + "double"; }
+        if (std::is_same<value_type, std::complex<float>>::value) { return prefix + "complex<float>"; }
+        if (std::is_same<value_type, std::complex<double>>::value) { return prefix + "complex<double>"; }
+#ifdef XSIMD_ENABLE_XTL_COMPLEX
+        if (std::is_same<value_type, xtl::xcomplex<float>>::value) { return prefix + "xcomplex<float>"; }
+        if (std::is_same<value_type, xtl::xcomplex<double>>::value) { return prefix + "xcomplex<double>"; }
+#endif
+
+        return prefix + "unknow_type";
```
serge-sans-paille, line 78: thanks for doing this.

CMakeLists.txt
```diff
@@ -37,42 +37,45 @@ message(STATUS "xsimd v${${PROJECT_NAME}_VERSION}")
...
-${XSIMD_INCLUDE_DIR}/xsimd/types/xsimd_utils.hpp
-${XSIMD_INCLUDE_DIR}/xsimd/types/xsimd_sse2_register.hpp
-${XSIMD_INCLUDE_DIR}/xsimd/types/xsimd_generic_arch.hpp
+${XSIMD_INCLUDE_DIR}/xsimd/stl/algorithms.hpp
 ${XSIMD_INCLUDE_DIR}/xsimd/types/xsimd_all_registers.hpp
-${XSIMD_INCLUDE_DIR}/xsimd/types/xsimd_avx512f_register.hpp
-${XSIMD_INCLUDE_DIR}/xsimd/types/xsimd_sse_register.hpp
-${XSIMD_INCLUDE_DIR}/xsimd/types/xsimd_sse4_1_register.hpp
+${XSIMD_INCLUDE_DIR}/xsimd/types/xsimd_api.hpp
+${XSIMD_INCLUDE_DIR}/xsimd/types/xsimd_arm7_register.hpp
```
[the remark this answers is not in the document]
JohanMabille: Actually we have three kinds of arm archs, arm7, arm8 32 bits and arm8 64 bits, but arm7 and arm8 32 bits share the same instruction set (at least regarding Neon instrutions). I planned to remove arm8_32 and rename arm7 into neon and arm_64 into neon64 in a dedicated PR (so we got this one merged and can merge future changes without the risk of a complicated rebase).

include-refactoring/xsimd/arch/generic/xsimd_generic_details.hpp
```diff
@@ -116,17 +116,24 @@ namespace xsimd {
...
-        template<class A, class From, class To>
-        using conversion_type = typename std::conditional<has_fast_conversion<A, From, To>::value, with_fast_conversion, with_slow_conversion>::type;
+
+      template <class A, class From, class To, class = void>
+      struct conversion_type_impl
+      {
+          using type = with_slow_conversion;
+      };
+
+      using xsimd::detail::void_t;
```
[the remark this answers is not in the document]
JohanMabille: Because I am actually in `xsimd::kernel::detail` namespace.

include-refactoring/xsimd/arch/xsimd_arm7.hpp
```diff
@@ -0,0 +1,1942 @@
...
+            template <template <class> class return_type, class... T>
+            struct arm_dispatcher_base
+            {
+                struct unary
+                {
+                    using container_type = std::tuple<return_type<T> (*)(T)...>;
+                    const container_type m_func;
+
+                    template <class U>
+                    return_type<U> run(U rhs) const
```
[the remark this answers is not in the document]
JohanMabille: That's definitely a batter name and will be more consistent with the other parts of the library.

include-refactoring/xsimd/arch/xsimd_arm7.hpp
JohanMabille: Because in the case of SFINAE with functions, you may want to be able to enable / disable the functions for different conditions. The following would be illegal:

```cpp
template <class T, class = std::enable_if_t<std::is_integral<T>, value>>
void my_func1(T t) { ... }

template <class T, class = std::enable_if_t<std::is_floating_point<T>, value>>
void my_func(T t) { ...}
```

This is because default template parameters are not part of the function signature, therefore you are defining the same function twice.

The following, on the other hand, is totally fine:

```cpp
template <class T, std::enable_if_t<std::is_integral<T>::value, int> = 0>
void my_func(T t) {... }

template <class T, std::enable_if_t<std::is_floating_point<T>::value, int> = 0>
void my_func(T t) { ... }
```

Since the `enable_if` does not resolve to a default template parameter, but to a non type parameter when the condition is true.

include-refactoring/xsimd/arch/xsimd_avx512f.hpp
JohanMabille: Yes, I added the `bitwise_not` test.

include-refactoring/xsimd/types/xsimd_api.hpp
JohanMabille: Will fix that

include-refactoring/xsimd/types/xsimd_batch.hpp
JohanMabille: Not sur to get you. The base type is still a POD type. The reason for this function is that I wante to instantiate the register type with a single bool value. This requires either a `kernel::broadcast` function for batch bool, or repeating the bool N times so that we can do `register{arg...}`.

include-refactoring/xsimd/types/xsimd_utils.hpp
JohanMabille: Theoretically yes, but I wanted to keep it consistent with `as_unsigned_integer` and the pattern from the STL: use `struc traits_name`  and inheritance so that you can specialize it, and `using traits_name_t = typename traits_name<T>::type` for keepping the code concise.

include-refactoring/xsimd/types/xsimd_utils.hpp
JohanMabille: Sorry for that, I've implemented that before you showed me the trick and forgot to make a second pass. I'm gonna fix it.

include-refactoring/xsimd/arch/generic/xsimd_generic_details.hpp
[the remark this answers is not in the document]
serge-sans-paille: That's indeed a very good reason :-)

PR #494, opened 2021-06-21
Implement permute conversion for Apache Arrow simd reference

guyuqi:
permute 2x128, 8x32 for 256bit/512bit.
The PR is the simple wrap of  `permute ` used by Arrow.


serge-sans-paille:
We now have a working version of `xsimd::swizzle` with dynamic masks that supersedes this PR, closing.

PR #496, opened 2021-06-23
Add new Algorithms using explicit batch type

michaelbacci:
New algorithms are added:

- count
- count_if
- transform_batch
- reduce_batch

Tests are updated.
README.md updated.

Using the added algorithms I've create a `nanmean_fast` function where the benchmark are the faster respect other implementation:

```
2021-06-24T00:59:51+02:00
Running ./bench_math
Run on (16 X 2300 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 256 KiB (x8)
  L3 Unified 16384 KiB (x1)
Load Average: 2.46, 2.32, 2.13
------------------------------------------------------------------------------
Benchmark                                    Time             CPU   Iterations
------------------------------------------------------------------------------
BM_nanmean/250                             487 ns          486 ns      1313789
BM_nanmean/1024                           2036 ns         2034 ns       348920
BM_nanmean_with_xtensor/250               5533 ns         5529 ns       116447
BM_nanmean_with_xtensor/1024             20085 ns        20063 ns        35643
BM_nanmean_xt/250                         1461 ns         1460 ns       478433
BM_nanmean_xt/1024                        4973 ns         4969 ns       138529
BM_nanmean_fast/250                        273 ns          272 ns      2553337
BM_nanmean_fast/1024                      1014 ns         1014 ns       674205
```

-  `BM_nanmean` has a classic C style while
-  `BM_nanmean_with_xtensor` is essentially `xt::mean(xt::filter(e, !xt::isnan(e)))` where `e` is a tensor
- `BM_nanmean_xt` is the `xt::nanmean`
- `BM_nanmean_fast` use the added `reduce_batch` and `count_if` algorithms

### DIFF
--- a/include/xsimd/stl/algorithms.hpp
+++ b/include/xsimd/stl/algorithms.hpp
@@ -11,125 +11,175 @@
 #ifndef XSIMD_ALGORITHMS_HPP
 #define XSIMD_ALGORITHMS_HPP
 
+#include <type_traits>
 #include "../memory/xsimd_load_store.hpp"
 
 namespace xsimd
 {
+    namespace detail 
+    {
+        template <typename... >
+        using void_t_v = void;
+
+        template <class T, class = void>
+        struct has_increment : std::false_type {};
+
+        template <class T>
+        struct has_increment<T, void_t_v<decltype(++std::declval<T>())>> : std::true_type {};
+
+        template <class... Args>
+        struct have_increment : all_true<has_increment<Args>::value...> {};
+
+        template <class... Args>
+        struct not_have_increment : all_true<!has_increment<Args>::value...> {};
+
+        template <class I1, class I2, class O1, class UF, class UFB>
+        using enable_unary_algorithm_t = typename std::enable_if<have_increment<I1, I2, O1>::value && not_have_increment<UF, UFB>::value, int>::type;
+
+        template <class I1, class I2, class I3, class O1, class UF>
+        using enable_binary_algorithm_t = typename std::enable_if<have_increment<I1, I2, I3,O1>::value && not_have_increment<UF>::value, int>::type;
+
+        template <class I1, class I2, class O1, class UF, class UFB>
+        void transform(I1 first, I2 last, O1 out_first, UF&& f, UFB&& fb)
+        {
+            using value_type = typename std::decay<decltype(*first)>::type;
+            using traits = simd_traits<value_type>;
+            using batch_type = typename traits::type;
+
+            std::size_t size = static_cast<std::size_t>(std::distance(first, last));
+            std::size_t simd_size = traits::size;
+
+            const auto* ptr_begin = &(*first);
+            auto* ptr_out = &(*out_first);
+
+            std::size_t align_begin = xsimd::get_alignment_offset(ptr_begin, size, simd_size);
+            std::size_t out_align = xsimd::get_alignment_offset(ptr_out, size, simd_size);
+            std::size_t align_end = align_begin + ((size - align_begin) & ~(simd_size - 1));
+
+            if (align_begin == out_align)
+            {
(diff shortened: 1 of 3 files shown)


JohanMabille (changes requested):
Thanks for your contribution! The implementation is really neat. I have some questions regarding the names of the functions as you can see below.

Regarding the failure in the tests, I think you have to include `xsimd_fallback.hpp` in the cpp file, so that the compiler can find the default implementation when a type is not supported by the current instructions set.

test/test_algorithms.cpp
```diff
@@ -216,6 +275,102 @@ TEST_F(xsimd_reduce, using_custom_binary_function)
     }
 }
 
+TEST(algorithms, reduce_batch)
+{
+    const double nan = std::numeric_limits<double>::quiet_NaN();
```
JohanMabille, line 280: For ARM, vectorization for `double` is available only on 64bits arch. Therefore, this test should be guarded with something like
`#if XSIMD_ARM_INSTR_SET >= XSIMD_ARM8_64_NEON_VERSION || XSIMD_X86_INSTR_SET >= XSIMD_X86_SSE2_VERSION`

include/xsimd/stl/algorithms.hpp
```diff
@@ -11,125 +11,187 @@
...
+    struct has_increment<T, void_t<decltype(++std::declval<T>())>> : std::true_type {};
 
-        std::size_t align_begin = xsimd::get_alignment_offset(ptr_begin, size, simd_size);
-        std::size_t out_align = xsimd::get_alignment_offset(ptr_out, size, simd_size);
-        std::size_t align_end = align_begin + ((size - align_begin) & ~(simd_size - 1));
+    template <class T>
+    using enable_if_increment = typename std::enable_if<has_increment<T>::value>::type;
+
+    template <class T>
+    using enable_if_not_increment = typename std::enable_if<!has_increment<T>::value>::type;
```
JohanMabille, line 32: I think it would be better to move these metafunctions in some `detail` namespace, they're not supposed to be part of the API.

include/xsimd/stl/algorithms.hpp
```diff
@@ -11,125 +11,187 @@
...
     }
 
-    template <class I1, class I2, class I3, class O1, class UF>
-    void transform(I1 first_1, I2 last_1, I3 first_2, O1 out_first, UF&& f)
+    template <class I1, class I2, class O1, class UF, class UFB,
+              typename = enable_if_increment<I1>,
+              typename = enable_if_increment<I2>,
+              typename = enable_if_increment<O1>,
+              typename = enable_if_not_increment<UF>,
+              typename = enable_if_not_increment<UFB>>
```
JohanMabille, line 157: It would be more readable to gather these conditions so that you can use a single `enable_if` condition. That could be something like:

```cpp
template <class... Args>
struct have_increment : all_true<has_increment<Args>::value...> {};

template <class... Args>
struct not_have_increment : all_true<!has_increment<Args>::value...> {};

template <class I1, class I2, class I3, class UF, class UFB>
using enable_binary_algorithm_t = typename std::enable_if<have_increment<I1, I2, O1>::value && not_have_increment<UF, UFB>::value, int>::type;
```

Besides, default template parameters are not considered by the compiler for overload selection, so it's better to use the `enable_if` as the template parameter evaluating to an int when it's valid:

```cpp
template <class I1, class I2, class O1, class UF, class UFB,
                 enable_binary_algorithm_t<I1, I2, O1, UF, UFB> = 0>
...
```

AntoinePrv:
Algorithm now live in https://github.com/xtensor-stack/xsimd-algorithm/

PR #500, opened 2021-07-04
Feature/avx migration

serge-sans-paille:


test-refactoring/avx2.sh
```diff
@@ -0,0 +1,173 @@
+#!/bin/sh
```
serge-sans-paille, line 1: @JohanMabille We're not running these tests during CI, but I found them super useful during the portage so I'm checking them in

PR #504, opened 2021-07-16
Fix non-constexpr static const variable overhead

yumeyao:
non-constexpr `static const` variable initialization overhead (guard variable and checking).

removing the `static` keyword removes the overhead

see https://godbolt.org/z/3b8sGnhcj.


serge-sans-paille:
LGTM, thanks

PR #506, opened 2021-07-17
Add remaining bitwise_cast permutations

astrohawk:
The existing bitwise_cast implementation only had support for casting
between floats, doubles, int32s and int64s. This change adds support
for casting to/from int8s and int16s as well as all the unsigned ints.

I chose to add explicit instantiations for all the different
permutations as opposed to using partial template specialisation or
SFINAE as I wanted to keep the implementation simple despite that
resulting in a very verbose implementation.

### DIFF
--- a/include/xsimd/types/xsimd_avx512_conversion.hpp
+++ b/include/xsimd/types/xsimd_avx512_conversion.hpp
@@ -242,45 +242,106 @@ namespace xsimd
      * bitwise cast functions implementation *
      *****************************************/
 
-    XSIMD_BITWISE_CAST_INTRINSIC(float, 16,
-                                 double, 8,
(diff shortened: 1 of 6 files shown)


astrohawk:
Hey @JohanMabille, is this ok to go in anytime soon or does it conflict with some of the refactoring you're doing on some of the other branches?

JohanMabille:
@astrohawk sorry for the late reply. Thanks for adding this, we have reimplemented the bitwise cast in the refactoring branch and will merge it soon in master. However, I htink it makes sense to merge your PR so that we can backport your work in a 7.x branch.

astrohawk:
@JohanMabille thanks so much for the merge. Looking forward to 7.x! Keep up the awesome work.

PR #507, opened 2021-07-18
Debugging sin function

JohanMabille:


JohanMabille:
Closing now since the bude is fixed.

PR #508, opened 2021-07-18
Feature/cleanup

serge-sans-paille:


serge-sans-paille:
Interestingly, the `u8_to_u*` functions were redundant with `bitcast` operation.

PR #532, opened 2021-08-04
Refactor examples

DerThorsten:
this makes the Mandelbrot example run again.
I refactored the code a bit such that it runs for all supported archs.

Downside of the generic approach is that I rely on `typeid(arch).name()` to print the name of the arch.
One could add a name function to the arch itself.
smth like 
```c++
    struct avx2 : avx
    {
        static constexpr std::string name() {return "avx2";}
        static constexpr bool supported() { return XSIMD_WITH_AVX2; }
        static constexpr bool available() { return true; }
        static constexpr unsigned version() { return generic::version(2, 2, 0); }
    };
```

### DIFF
--- a/examples/mandelbrot.cpp
+++ b/examples/mandelbrot.cpp
@@ -16,7 +16,6 @@
 #include <iostream>
 #include <string>
 #include <vector>
-#include <typeinfo>       // operator typeid
 
 #include "pico_bench.hpp"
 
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
> Downside of the generic approach is that I rely on `typeid(arch).name()` to print the name of the arch.
 #533 adds a `name()` method that returns a `char const*` (can't use an `std::string` which is not constexpr).

SylvainCorlay:
Happy to get this in as soon as the `typeid` call is not required :)

DerThorsten:
`arch::name()` is used instead of `typeid`.
Can be merged.

SylvainCorlay:
Thanks!

PR #535, opened 2021-08-05
added XSIMD_VERSION_* defines back to config

DerThorsten:
the config file is missing `XSIMD_VERSION_*`  which are still needed in the CMakeLists.
Not sure if 8.0.0 is the correct version of master

### DIFF
--- a/include/xsimd/config/xsimd_config.hpp
+++ b/include/xsimd/config/xsimd_config.hpp
@@ -1,6 +1,9 @@
 #ifndef XSIMD_CONFIG_HPP
 #define XSIMD_CONFIG_HPP
 
+#define XSIMD_VERSION_MAJOR 8
+#define XSIMD_VERSION_MINOR 0
+#define XSIMD_VERSION_PATCH 0
 
 /**
  * high level free functions


serge-sans-paille:
I'd like to avoid these defines. Why are they needed?

SylvainCorlay:
Note: before the refactor, it was probably in a top-level header file that we parsed from CMake. (See e.g. an example of that in the xtensor project).

This allowed to
 - have the information in C++ macros for downstream packages including the headers.
 - still define the version only in one place.

This probably went away with the refactor.

serge-sans-paille:
> parsed from CMake. (See e.g. an example of that in the xtensor project).

I don't see the xtensor reference, can you point me at it?

DerThorsten:
The parsing in CMake is still there as it should be  so I think everything is fine
https://github.com/xtensor-stack/xsimd/blob/d4bf79fec205f4bad71d1ad4521400f29d65b59d/CMakeLists.txt#L20

SylvainCorlay:
Yes what I meant is that it should be similar in xsimd

PR #537, opened 2021-08-14
Allow ommision of Arch argument for load / load_aligned / load_unaligned

serge-sans-paille:


serge-sans-paille:
This is a requirement from the pythran codebase

serge-sans-paille:
@JohanMabille please don't merge, I need to add a few tests

PR #539, opened 2021-08-19
Refactor array_to_batch implementation

guyuqi:
xsimd was switched to new implementation.
The PR is to refactor arry_to_batch for the previous PR #490


guyuqi:
Let me fix build issues in CI.

guyuqi:
Thanks for your comments on PR #490. @JohanMabille 
For xsimd was re-implemented and added Architecture parameter: "A" to Class 'Batch',
 this PR is to refactor PR#490.

I also added the related test cases. All CI issues were fixed and tests cases were passed on x86(sse, avx and avx512) and Arm64.
Could you please help to review it, thanks.

serge-sans-paille:
I'm not a big fan of this PR in current state : we already have https://github.com/xtensor-stack/xsimd/blob/master/include/xsimd/types/xsimd_batch.hpp#L356 that does a pretty similar job. I'd rather fill the holes in the design than introducing an extra function that overlaps with existing ones.

serge-sans-paille:
@guyuqi : let me reformulate, the above comment may sound a bit "straight to the face" ^^!

We appreciate the effort, we just need to find a way to interact in a nice manner with the new infrastructure. What about introducing a new constructor like

```
template<class Iterator> batch(Iterator begin, Iterator end)
```

that forwards to the appropriate kernel? That looks like a nice generalization of your proposal.

guyuqi:
@serge-sans-paille 
Thanks for your explanation. It makes sense.

Could you please explain more details about the new constructor ? : 
`template<class Iterator> batch(Iterator begin, Iterator end)
`

guyuqi:
@serge-sans-paille
Agreed that the PR does a pretty similar job like `set` in :  https://github.com/xtensor-stack/xsimd/blob/master/include/xsimd/types/xsimd_batch.hpp#L356.

The implementation in: https://github.com/xtensor-stack/xsimd/blob/master/include/xsimd/arch/xsimd_avx512f.hpp#L991-#L999
```
    template<class A, class T, detail::enable_signed_integer_t<T> = 0>
    batch<T, A> set(batch<T, A> const&, requires_arch<avx512f>, T v0, T v1, T v2, T v3, T v4, T v5, T v6, T v7,
                               T v8, T v9, T v10, T v11, T v12, T v13, T v14, T v15,
                                T v16, T v17, T v18, T v19, T v20, T v21, T v22, T v23,
                                T v24, T v25, T v26, T v27, T v28, T v29, T v30, T v31,
                                T v32, T v33, T v34, T v35, T v36, T v37, T v38, T v39,
                                T v40, T v41, T v42, T v43, T v44, T v45, T v46, T v47,
                                T v48, T v49, T v50, T v51, T v52, T v53, T v54, T v55,
                                T v56, T v57, T v58, T v59, T v60, T v61, T v62, T v63
```
To make the code easier to maintain, how about  use `std::array` instead of the 64 input arguments? 

And also thanks for your suggestion for the new constructor:
`template<class Iterator> batch(Iterator begin, Iterator end)
`


Does it mean:
(intuitive example)
```
template<class T, class A, class Iterator>
batch(T, A) bytes_array_to_batch(Iterator begin, Iterator end, A{})
{
     return _mm256_set_epi8( *end, *(end - 1), *(end - 2), ... , *(end - Iterator->size() -1));
}
```
I'm sorry I don't have a good understanding of this new Iterator template.
Could you kindly help to give more explanation for this new constructor, or an simple example? 

Thanks. :-)

serge-sans-paille:
> @serge-sans-paille
> Agreed that the PR does a pretty similar job like `set` in : https://github.com/xtensor-stack/xsimd/blob/master/include/xsimd/types/xsimd_batch.hpp#L356.

Thanks for being more open minded than I am :-)

> To make the code easier to maintain, how about use `std::array` instead of the 64 input arguments?

Let's consider this function as an internal one. Using an std::array may be valuable but it's probably for another PR.

> And also thanks for your suggestion for the new constructor:
> `template<class Iterator> batch(Iterator begin, Iterator end) `
> 
> Does it mean:
> (intuitive example)
> 
> ```
> template<class T, class A, class Iterator>
> batch(T, A) bytes_array_to_batch(Iterator begin, Iterator end, A{})
> {
>      return _mm256_set_epi8( *end, *(end - 1), *(end - 2), ... , *(end - Iterator->size() -1));
> }
> ```

Very close! Let's use the `set` function instead and that really looks good to me!

guyuqi:
@serge-sans-paille
OK, let me reformulate it.
I will:
1. Close this PR.
2. Refine the `set` in another PR: 
    1) refine the API input arguments to `std:array`  as mentioned above.
    2) Make the function `set ` as an internal one and introduce a new constructor with `Iterator ` that is also called `set ` to forwards to the appropriate kernel like:
```
template<class T, class A, class Iterator>
batch(T, A) set(Iterator begin, Iterator end, A{})
{
     return _mm256_set_epi8( *end, *(end - 1), *(end - 2), ... , *(end - Iterator->size() -1));
}

```

Do you have more comments on it?
Please correct me if I missed something or something misunderstood.

JohanMabille:
Hi @guyuqi,

Sorry for jumping so late in this. The idea would be to keep this PR to implement a new constructor for the `batch` class, that accepts a pair of iterators:

```cpp
template <class It>
batch(It first, It last);
```

Its implementation would be very similar to [this constructor](https://github.com/xtensor-stack/xsimd/blob/master/include/xsimd/types/xsimd_batch.hpp#L35).

Notice that the `set` functions are already internal to the library, you cannot find them in `https://github.com/xtensor-stack/xsimd/blob/master/include/xsimd/types/xsimd_api.hpp`.

 The next PR should refactor the set functions themselves. We should replace the existing implementation with two overloads of `set` accepting iterator pairs only, enables/disabled with SFINAE depending on the contiguity of the iterator. For iterators from a contiguous container, the implementation should call loadu. For Iterators from a non-contiguous container, the implementation should be similar to the existing one (i.e., calling _mmXX_setr_YY passing all the arguments).

guyuqi:
> Hi @guyuqi,
> 
> Sorry for jumping so late in this. The idea would be to keep this PR to implement a new constructor for the `batch` class, that accepts a pair of iterators:
> 
> ```c++
> template <class It>
> batch(It first, It last);
> ```
> 
> Its implementation would be very similar to [this constructor](https://github.com/xtensor-stack/xsimd/blob/master/include/xsimd/types/xsimd_batch.hpp#L35).
> 
> Notice that the `set` functions are already internal to the library, you cannot find them in `https://github.com/xtensor-stack/xsimd/blob/master/include/xsimd/types/xsimd_api.hpp`.
> 
> The next PR should refactor the set functions themselves. We should replace the existing implementation with two overloads of `set` accepting iterator pairs only, enables/disabled with SFINAE depending on the contiguity of the iterator. For iterators from a contiguous container, the implementation should call loadu. For Iterators from a non-contiguous container, the implementation should be similar to the existing one (i.e., calling _mmXX_setr_YY passing all the arguments).

@JohanMabille 
Sorry for the late reply. Thanks for your explanation. I'll refine this PR to implement a new constructor that accepts a pair of iterators.
And I also would like to refactor the `set` like you suggested. Thanks.

guyuqi:
@JohanMabille 
Refactor the implementation of functions and constructor with iterators.
Tests were passed on x86 and Arm64.

guyuqi:
Rebased and clang format.

Issue #548, opened 2021-08-31
Some functions broken with "-ffast-math" enabled


RafaGago:
Coming from:
https://github.com/xtensor-stack/xsimd/issues/515

Sorry for closing and reopening. I got mislead because I thought I did a mistake. It seems that the bug is fixed on gcc but remains on clang. Snippet:

```
#include <cmath>
#include <stdlib.h>
#include <string.h>
#include <xsimd/xsimd.hpp>

static void check (double in, double cm, double xs)
{
  constexpr double epsilon = 0.1; // error 10%
  constexpr double min     = 1. - epsilon;
  constexpr double max     = 1. + epsilon;

  double fac = cm / xs;
  if (fac < min || fac > max) {
    printf ("intput: %f, std: %f, xsimd: %f\n", in, cm, xs);
  }
}

int main (int argc, char const* argv[])
{
  for (double i = 0.; i < 100.; i += 0.01) {
    check (i, tanh (i), xsimd::tanh (xsimd::batch<double> {i}).get (0));
  }
  return 0;
};
```

`clang++-12 -O2 -ffast-math main.cpp -o main -I <xsimd_path>`

Output:

```
input: 0.000000, std: 0.000000, xsimd: 0.000000
input: 0.630000, std: 0.558052, xsimd: 0.333333
input: 0.640000, std: 0.564900, xsimd: 0.333333
input: 0.650000, std: 0.571670, xsimd: 0.333333
input: 0.660000, std: 0.578363, xsimd: 0.333333
input: 0.670000, std: 0.584980, xsimd: 0.333333
input: 0.680000, std: 0.591519, xsimd: 0.333333
input: 0.690000, std: 0.597982, xsimd: 0.333333
input: 0.800000, std: 0.664037, xsimd: 0.600000
input: 0.810000, std: 0.669590, xsimd: 0.600000
input: 0.820000, std: 0.675070, xsimd: 0.600000
input: 0.830000, std: 0.680476, xsimd: 0.600000
input: 0.840000, std: 0.685809, xsimd: 0.600000
input: 0.850000, std: 0.691069, xsimd: 0.600000
input: 0.860000, std: 0.696258, xsimd: 0.600000
input: 0.870000, std: 0.701374, xsimd: 0.600000
input: 0.880000, std: 0.706419, xsimd: 0.600000
input: 0.890000, std: 0.711394, xsimd: 0.600000
input: 0.900000, std: 0.716298, xsimd: 0.600000
[...]

PR #550, opened 2021-08-31
nearbyhint: Fix for ffast-math

RafaGago:
See commit text.

This fixes:
#548

Notice that I wasn't able to verify it on Visual Studio, just g++ and clang.

### DIFF
--- a/include/xsimd/arch/generic/xsimd_generic_math.hpp
+++ b/include/xsimd/arch/generic/xsimd_generic_math.hpp
@@ -1707,6 +1707,40 @@ namespace xsimd {
     }
 
 
+#if !defined(__FAST_MATH__)
+    template <class T>
+    T conformant_add_then_sub (T x, T v)
+    {
+        return x + v - v;
+    }
+#else
+#if defined(__clang__)
+    // available on clang 4
+    #define XSIMD_NO_OPTIMIZATION_ATTRIBUTE __attribute__((optnone))
+    #define XSIMD_NO_OPTIMIZATION_PRAGMA
+#elif defined(__GNUC__)
+    // available on GCC 4.9
+    #define XSIMD_NO_OPTIMIZATION_ATTRIBUTE __attribute__((optimize("O0")))
+    #define XSIMD_NO_OPTIMIZATION_PRAGMA
+#elif defined(_MSC_VER)
+    // available Visual Studio 2015
+    #define XSIMD_NO_OPTIMIZATION_ATTRIBUTE
+    #define XSIMD_NO_OPTIMIZATION_PRAGMA __pragma(optimize("", off))
+#else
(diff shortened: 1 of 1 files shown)


include/xsimd/arch/generic/xsimd_generic_math.hpp
```diff
@@ -1707,6 +1707,40 @@ namespace xsimd {
     }
 
 
+#if !defined(__FAST_MATH__)
```
serge-sans-paille, line 1710: Interesting approach! I wonder if the following works as expected and is faster or not:
```
namespace details {
template<class T>
T force_add(T x, T y) {
#if defined(__FAST_MATH__) && (defined(__clang__) || defined(__GNUC__))
__attribute__((noinline))
#endif
    return x + y;
}
}
```

then call it appropriately

serge-sans-paille:
Can you give a try to the following patch instead: if it works, it's much less intrusive

```
diff --git a/include/xsimd/arch/generic/xsimd_generic_math.hpp b/include/xsimd/arch/generic/xsimd_generic_math.hpp
index 56e4d98..2ef41de 100644
--- a/include/xsimd/arch/generic/xsimd_generic_math.hpp
+++ b/include/xsimd/arch/generic/xsimd_generic_math.hpp
@@ -1722,8 +1722,8 @@ namespace xsimd {
         // to v. That's not what we want, so prevent compiler optimization here.
         // FIXME: it may be better to emit a memory barrier here (?).
 #ifdef __FAST_MATH__
-        volatile batch_type d0 = v + t2n;
-        batch_type d = *(batch_type*)(void*)(&d0) - t2n;
+        volatile auto d0 = (v + t2n).data;
+        batch_type d = batch_type(d0) - t2n;
 #else
         batch_type d0 = v + t2n;
         batch_type d = d0 - t2n;
```

include/xsimd/arch/generic/xsimd_generic_math.hpp
RafaGago: Now on $DAILY_JOB.

Without trying I'd say that a competent compiler will still optimize it away.

RafaGago:
Now I'm at $DAILY_JOB I don't have the codebase at hand, but I'd say that local scope volatiles are not enough to stop clang.

Notice that to remove the volatile instead of doing "*(batch_type*)(void*)(&d0)" you can do "const_cast<batch const&> (d0)"

include/xsimd/arch/generic/xsimd_generic_math.hpp
RafaGago: But notice that the optimization we want to break is ffast-math related. Inlining should make no difference.

DiamonDinoia:
I think master now handles this.

PR #551, opened 2021-09-02
Minimalistic fast math support

serge-sans-paille:


SylvainCorlay:
Does this actually enable the testing?

serge-sans-paille:
No, I gave up with testing under `-ffast-math` because it breaks *a lot* of function and I'm notsure we want to actually support that. This PR is just to help for #550 . Just renamed it.

DiamonDinoia:
same here master now should handle this

PR #552, opened 2021-09-04
Feature/migration guide

serge-sans-paille:


serge-sans-paille:
@DerThorsten / @JohanMabille did I miss something important there?

JohanMabille:
@serge-sans-paille sorry for the late reply. I think the most important changes are mentionned. We will update the migration guide if we find missing changes when upgrading xtensor to xsimd 8.

PR #553, opened 2021-09-08
Fix avx512 build error from clang

cyb70289:
I'm building apache arrow with latest xsimd and met with many errors
when using clang and enabling avx512. This patch fixes the issues.

### DIFF
--- a/include/xsimd/arch/xsimd_avx512f.hpp
+++ b/include/xsimd/arch/xsimd_avx512f.hpp
@@ -148,10 +148,10 @@ namespace xsimd {
       else {
         switch(sizeof(T)) {
           case 1: {
-            uint64_t mask_low0 = _mm512_cmp_epu32_mask(batch<uint32_t, A>(self.data) & batch<uint32_t, A>(0x000000FF), batch<uint32_t, A>(other.data)  & batch<uint32_t, A>(0x000000FF), Cmp);
-            uint64_t mask_low1 = _mm512_cmp_epu32_mask(batch<uint32_t, A>(self.data) & batch<uint32_t, A>(0x0000FF00), batch<uint32_t, A>(other.data)  & batch<uint32_t, A>(0x0000FF00), Cmp);
-            uint64_t mask_high0 = _mm512_cmp_epu32_mask(batch<uint32_t, A>(self.data) & batch<uint32_t, A>(0x00FF0000), batch<uint32_t, A>(other.data)  & batch<uint32_t, A>(0x00FF0000), Cmp);
-            uint64_t mask_high1 = _mm512_cmp_epu32_mask(batch<uint32_t, A>(self.data) & batch<uint32_t, A>(0xFF000000), batch<uint32_t, A>(other.data)  & batch<uint32_t, A>(0xFF000000), Cmp);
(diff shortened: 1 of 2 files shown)


include/xsimd/arch/xsimd_avx512f.hpp
```diff
@@ -148,10 +148,10 @@ namespace xsimd {
       else {
         switch(sizeof(T)) {
           case 1: {
-            uint64_t mask_low0 = _mm512_cmp_epu32_mask(batch<uint32_t, A>(self.data) & batch<uint32_t, A>(0x000000FF), batch<uint32_t, A>(other.data)  & batch<uint32_t, A>(0x000000FF), Cmp);
-            uint64_t mask_low1 = _mm512_cmp_epu32_mask(batch<uint32_t, A>(self.data) & batch<uint32_t, A>(0x0000FF00), batch<uint32_t, A>(other.data)  & batch<uint32_t, A>(0x0000FF00), Cmp);
-            uint64_t mask_high0 = _mm512_cmp_epu32_mask(batch<uint32_t, A>(self.data) & batch<uint32_t, A>(0x00FF0000), batch<uint32_t, A>(other.data)  & batch<uint32_t, A>(0x00FF0000), Cmp);
-            uint64_t mask_high1 = _mm512_cmp_epu32_mask(batch<uint32_t, A>(self.data) & batch<uint32_t, A>(0xFF000000), batch<uint32_t, A>(other.data)  & batch<uint32_t, A>(0xFF000000), Cmp);
+            uint64_t mask_low0 = _mm512_cmp_epu32_mask((batch<uint32_t, A>(self.data) & batch<uint32_t, A>(0x000000FF)), (batch<uint32_t, A>(other.data)  & batch<uint32_t, A>(0x000000FF)), Cmp);
```
cyb70289, line 151: `_mm512_cmp_xxx_mask` are macros in clang. The `,` between template augments confuses the compiler.
Adding parentheses fixes the issue.

include/xsimd/config/xsimd_config.hpp
```diff
@@ -128,6 +128,9 @@
  * Set to 1 if AVX512F is available at compile-time, to 0 otherwise.
  */
 #ifdef __AVX512F__
+#if defined(__clang__) && __clang_major__ >= 6
+    #define XSIMD_WITH_AVX512F 1
+#else
     // AVX512 instructions are supported starting with gcc 6
     // see https://www.gnu.org/software/gcc/gcc-6/changes.html
     #if defined(__GNUC__) && __GNUC__ < 6
```
cyb70289, line 136: clang also defines `__GNUC__`, but looks it's fixed to 4, even for latest clang-12 [1]. So avx512 is always disabled when build xsimd with clang.
Add explicit clang checking to fix the issue.
[1] https://godbolt.org/z/n6cEexPfn

cyb70289:
@pitrou

include/xsimd/config/xsimd_config.hpp
pitrou, line 136: Nit: can collapse `#else` and `#if` into `#elif`.

pitrou:
Do we know why CI didn't catch this on master?

JohanMabille:
@cyb70289 thanks for the fixes!
@pitrou I think we test avx512 with gcc only, I will add a build ofr testing it with clang.

PR #554, opened 2021-09-09
Fix build error on Arm64

cyb70289:


include/xsimd/arch/xsimd_neon64.hpp
```diff
@@ -658,7 +658,7 @@ namespace xsimd
             {
                 if (n == I)
                 {
-                    return vextq_f64(lhs, rhs, n);
+                    return vextq_f64(lhs, rhs, I);
```
cyb70289, line 661: last augment of `vextq` must be constant

cyb70289:
cc @JohanMabille @pitrou

pitrou:
Again, it seems weird that the ARM CI checks didn't catch this? (I see at least one named "Linux_1 gcc_4_aarch64")

JohanMabille:
Thanks!

PR #563, opened 2021-09-14
Activate SSE2 test on windows

serge-sans-paille:


cyb70289:
Turns out AVX2 is always enabled in the source code
https://github.com/xtensor-stack/xsimd/blob/master/test/CMakeLists.txt#L92
@serge-sans-paille

serge-sans-paille:
thanks @cyb70289 . I now have a clean reproducer for your bug, is that okay if I cherry-pick your fix in this branch and adjust it?

cyb70289:
Sure, no problem

serge-sans-paille:
All green. Thanks @cyb70289 for the hints! and cc @JohanMabille for the merge.

include/xsimd/config/xsimd_config.hpp
```diff
@@ -238,6 +238,10 @@
 #define XSIMD_WITH_SSE2 1
 #endif
 
+#if defined(_M_X64) || (defined(_M_IX86_FP) && _M_IX86_FP >= 2)
+#define XSIMD_WITH_SSE2 1
+#endif
```
cyb70289, line 243: If we only enable SSE2 here, then SSE3,SSE4 will not be supported by xsimd if built with msvc.
I think this is a problem and will still break many code.
MSVC X64 must support SSE4_2 (actually /arch:SSE2 is an invalid option for msvc x64).
What about define `XSIMD_WITH_SSE4_2` if _M_X64?

serge-sans-paille: According to the doc there's indeed no explicit flag for SSE3 and later (based on https://docs.microsoft.com/fr-fr/cpp/build/reference/arch-x86?view=msvc-160) but I wonder if it's actually supported if no arch flag is specified ?

cyb70289: X64 msvc only supports /arch:AVX?, https://docs.microsoft.com/en-us/cpp/build/reference/arch-x64?view=msvc-160

My understanding is /arch:XXX anables msvc to auto vectorize c++ code using at most the XXX instructions.
But it doesn't prevent us from using intrinsics directly. Compiler should supports all level simd instructions.

cyb70289 (approved):
Thanks @serge-sans-paille !

serge-sans-paille:
@cyb70289 unfortunately, I'm not able to reproduce the windows error locally (on `xsimd::extract_pair`) could you have a look?

include/xsimd/arch/xsimd_ssse3.hpp
```diff
@@ -31,7 +31,7 @@ namespace xsimd {
       template<class T, class A, std::size_t I, std::size_t... Is>
       batch<T, A> extract_pair(batch<T, A> const& self, batch<T, A> const& other, std::size_t i, ::xsimd::detail::index_sequence<I, Is...>) {
         if(i == I) {
-          return _mm_alignr_epi8(self, other, sizeof(T) * I);
+          return _mm_alignr_epi8(other, self, sizeof(T) * (I));
```
cyb70289, line 34: `other` and `self` swapped?

serge-sans-paille:
yeah, note that this makes the validation pass :-) So unless the validation
itself is not correctly written...

serge-sans-paille:
@cyb70289 can you confirm e.g. on your code base that current implementation is correct, i.e. that the tests are well-written?

cyb70289:
> @cyb70289 can you confirm e.g. on your code base that current implementation is correct, i.e. that the tests are well-written?

I use `extract_pair` in my simd code. It works correctly on both x86 and arm.
So I guess maybe xsimd test is not correct. Will have a look at my side.

cyb70289:
Looks to me this test code to generate expected result is wrong:
https://github.com/xtensor-stack/xsimd/blob/master/test/test_extract_pair.cpp#L35-L42

`_mm_alignr_epi8(lhs, rhs, 1)` should return (*from lsb to msb*) `rhs[1] rhs[2] ... rhs[15] lhs[0]`,
but the test code expects `lhs[1] lhs[2] ... lhs[15] rhs[0]`.
https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_alignr_epi8&expand=302

include/xsimd/arch/xsimd_neon.hpp
```diff
@@ -1411,7 +1411,7 @@ namespace xsimd
             {
                 if (n == I)
                 {
-                    return vextq_u8(lhs, rhs, I);
+                    return vextq_u8(rhs, lhs, I);
```
cyb70289, line 1414: Ah, `vextq_xx` is indeed different from `_mm_alignr_xx` 👍

serge-sans-paille:
All green! cc @JohanMabille this one costed me quite a lot of effort :-)

JohanMabille:
Congratulation on this one!

PR #565, opened 2021-09-16
Update vectorized_code.rst

compnerd:
Add missing `:` in namespace delimiter.   This should correct the rendering of the documentation.


serge-sans-paille:
Clean & easy. Thanks!

PR #570, opened 2021-09-28
Fix build error on mingw32

cyb70289:
Mingw32 build failed with missing _mm_cvtsi128_si64 definition, which
is only available on 64bit platform.

### DIFF
--- a/include/xsimd/arch/xsimd_avx2.hpp
+++ b/include/xsimd/arch/xsimd_avx2.hpp
@@ -173,7 +173,15 @@ namespace xsimd {
                 __m256i tmp2 = _mm256_add_epi64(self, tmp1);
                 __m128i tmp3 = _mm256_extracti128_si256(tmp2, 1);
                 __m128i res = _mm_add_epi64(_mm256_castsi256_si128(tmp2), tmp3);
+#if defined(__x86_64__)
                 return _mm_cvtsi128_si64(res);
+#else
+                __m128i m;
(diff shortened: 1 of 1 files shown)


include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -173,7 +173,15 @@ namespace xsimd {
                 __m256i tmp2 = _mm256_add_epi64(self, tmp1);
                 __m128i tmp3 = _mm256_extracti128_si256(tmp2, 1);
                 __m128i res = _mm_add_epi64(_mm256_castsi256_si128(tmp2), tmp3);
+#if defined(__x86_64__)
```
cyb70289, line 176: Copied from SSE2 code
https://github.com/xtensor-stack/xsimd/blob/master/include/xsimd/arch/xsimd_sse2.hpp#L426-L434

cyb70289: Error log at https://github.com/apache/arrow/pull/11142/checks?check_run_id=3727229675#step:9:1265

cyb70289: Yes, it's an issue only on 32 bits.

cyb70289: Feel free to pick this change if it's useful.

serge-sans-paille:
Obsoleted by #571

PR #571, opened 2021-09-28
Test compilation on 32 bit target

serge-sans-paille:


serge-sans-paille:
@cyb70289 I cherry-picked your patch and it fixed validation, looks good now!

PR #574, opened 2021-09-29
Fix ptrdiff_t undefined error

cyb70289:


include/xsimd/memory/xsimd_aligned_allocator.hpp
```diff
@@ -14,6 +14,7 @@
 #include "../config/xsimd_arch.hpp"
 
 #include <algorithm>
+#include <cstddef>
```
cyb70289, line 17: See error log at apache arrow ci link
https://github.com/apache/arrow/pull/11142/checks?check_run_id=3739324298#step:8:1246

PR #576, opened 2021-09-30
Fix "int_sequence not declared" error when build Neon with C++14

cyb70289:


include/xsimd/types/xsimd_utils.hpp
```diff
@@ -240,6 +240,7 @@ namespace xsimd
 
             template <typename... Ts>
             using index_sequence_for = make_index_sequence<sizeof...(Ts)>;
+        #endif
```
cyb70289, line 243: Symbols below should be available no matter c++ version.
Original code only defines those symbols under c++11, and causes error when build xsimd neon if -std=c++14.
`xsimd::detail::int_sequence’ has not been declared`

serge-sans-paille: The `int_sequence` code looks super redundant with the `index_sequence` code. I'll propose a patch that make the whole stuff more generic and take your remark into account.

serge-sans-paille:
Can you confirm #580 fixes the issue?

cyb70289:
> Can you confirm #580 fixes the issue?

Yes, #580 fixes this issue. Thanks.

PR #578, opened 2021-09-30
[WIP] Fix Neon build error if gcc -O0

### LINKED ISSUE #573: NEON64 code link error if gcc -O1 or -O0
A simple test code
------------------------
```cpp
#include <cstddef>
#include <xsimd/xsimd.hpp>

int main() {
  using simd_batch = xsimd::batch<int8_t, xsimd::neon64>;

  simd_batch a(1);
  simd_batch b(2);
  simd_batch c = a | b;

  return xsimd::any(b < c);
}
```

Compiler with gcc -O1 or -O0
---------------------------------------
```bash
$ g++ -O1 neon-debug.cc  -I./include -march=armv8-a -std=c++11

/tmp/cc89XtTV.o: In function `main':
neon-debug.cc:(.text+0x10): undefined reference to `vorrq_s8(__Int8x16_t, __Int8x16_t)'
neon-debug.cc:(.text+0x1c): undefined reference to `vcltq_s8(__Int8x16_t, __Int8x16_t)'
collect2: error: ld returned 1 exit status
```

Other info
-------------
- Tested with gcc-7.5, 9.3, 10.3, all failed
- OK if -O2, -O3
- OK for clang, no matter -O?

cyb70289:
Fixes #573


include/xsimd/arch/xsimd_neon.hpp
```diff
@@ -834,6 +834,21 @@ namespace xsimd
...
+            BITWISE_OR_TYPE(u8, uint8x16_t)
+            BITWISE_OR_TYPE(s8, int8x16_t)
+            BITWISE_OR_TYPE(u16, uint16x8_t)
+            BITWISE_OR_TYPE(s16, int16x8_t)
+            BITWISE_OR_TYPE(u32, uint32x4_t)
+            BITWISE_OR_TYPE(s32, int32x4_t)
+            BITWISE_OR_TYPE(u64, uint64x2_t)
+            BITWISE_OR_TYPE(s64, int64x2_t)
+            #undef BITWISE_OR_TYPE
+
```
cyb70289, line 851: Defines a bunch of thin wrappers around vorrq_xxx intrinsics.
@JohanMabille , do you think it's okay? or any better approach?

serge-sans-paille: @cyb70289 the pattern is used through this file, so it's probably a good idea to have a more generic approach. What do you think of the following snippet:

https://godbolt.org/z/MoGPqWoqz

cyb70289: The problem is that we cannot pass intrinsics to any function. In gcc -O0 or O1, it will take the passed intrinsics function pointer as normal function pointer and try to find the symbol at link time.

Please see https://godbolt.org/z/PG79js33a

`make_wrapped_functions(_mm_or_si128)`  (line 25) generates below code.
```
_GLOBAL__sub_I_t:
        mov     QWORD PTR t[rip], OFFSET FLAT:_Z12_mm_or_si128Dv2_xS_
        ret
```
It returns address of a symbol `_mm_or_si128` that gcc thinks exists somewhere. It will fail at link time as there's no such function implementation.

If you uncomment line 26 (and comment line 25), which enables a direct wrapping function,  the code is okay. The intrinsic assembly `por xmm0, xmm1` is inside the generated code.
```
my_or128(long long __vector(2), long long __vector(2)):
        por     xmm0, xmm1
        ret
_GLOBAL__sub_I_t:
        mov     QWORD PTR t[rip], OFFSET FLAT:my_or128(long long __vector(2), long long __vector(2))
        ret
```

JohanMabille:
Hi,

Sorry for the late reply. Even if I would like to avoid macro as much as possible, it looks like we don't have the choice here, I don't have a better alternative to propose.

cyb70289:
I also don't like macros.
For C++14, a better approach may be defining only one macro to wrap all intrinsics.
`#define F(f) [](auto... args){ return f(args...); }`
Then we have to change `make_tuple(vorrq_u8, vorrq_s8, ...)` to `make_tuple(F(vorrq_u8), F(vorrq_s8), ...)`.
Haven't tested, looks it can work.
For C++11, we need to infer function parameter type, should be doable.

PR #580, opened 2021-09-30
Make make_integer_sequence generic

serge-sans-paille:
So that it can be used for make_index_sequence and make_int_sequence.


serge-sans-paille:
Should obsolete #576

PR #581, opened 2021-10-01
Fix SSE2 not being detected correctly on MSVC for 32-bit targets

marzer:
MSVC annoyingly does not define `__SSE2__` at all, instead relying on the fact that:
- SSE2 is supported unconditionally for x86-64 targets (which you already handle)
- SSE2 is opt-in for x86-32 targets and is detected by checking [`_M_IX86_FP >= 2`](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-160`)

This PR adds the requisite checks to handle the 32-bit target scenario on MSVC.

### DIFF
--- a/include/xsimd/config/xsimd_config.hpp
+++ b/include/xsimd/config/xsimd_config.hpp
@@ -245,7 +245,7 @@
 #define XSIMD_WITH_SSE3 1
 #endif
 
-#if XSIMD_WITH_SSE3
+#if XSIMD_WITH_SSE3 || (defined(_M_IX86_FP) && _M_IX86_FP >= 2)
 #undef XSIMD_WITH_SSE2
 #define XSIMD_WITH_SSE2 1
 #endif


include/xsimd/config/xsimd_config.hpp
```diff
@@ -245,7 +245,7 @@
 #define XSIMD_WITH_SSE3 1
 #endif
 
-#if XSIMD_WITH_SSE3
+#if XSIMD_WITH_SSE3 || (defined(_M_IX86_FP) && _M_IX86_FP >= 2)
```
serge-sans-paille, line 248: Should it be `|| (!defined(__clang__) && (defined(_M_IX86_FP) && _M_IX86_FP >= 2))` ? Can you try with clang-cl ?

marzer: Clang-cl supports the `_M_IX86_FP` flag correctly (https://github.com/microsoft/clang/blob/master/test/Driver/cl-x86-flags.c) so this PR stunt negatively impact that target

serge-sans-paille:
#582 turns on testing on 32bit targets. No issue related to instruction set detection spotted, how strange...

marzer:
@serge-sans-paille it's likely that clang-cl also defines __SSE2__ (since it's Clang after all), in which case it would have already worked correctly there, and my changes would only impact 'real' MSVC

serge-sans-paille:
Yeah, that's why I think we should mimic the `if !defined(__clang__)` from above ;-)

serge-sans-paille:
On Mon, Oct 04, 2021 at 08:17:42AM -0700, Mark Gillard wrote:
> Er, sure? Can't say I understand why, though.
> 
> If clang-cl doesn't define __SSE2__, and instead fully relies on the MSVC
> _M_IX86_FP, then my fix works as intended.
> 
> If it does define __SSE2__, then my fix won't have any effect because the
> detection have already been resolved correctly per
> 
> #ifdef __SSE2__
> #define XSIMD_WITH_SSE2 1
> #else
> #define XSIMD_WITH_SSE2 0
> #endif
> 
> at the top of the file.

Yeah, and in that case, we'll be defining that macro twice, hitting
(potentially) a warning :-)

marzer:
Sorry but you are mistaken - the MSVC specific block explicitly `#undefs` it first. There will be no warning.

serge-sans-paille:
I'm glad to be wrong :-) Trust you on that one!

PR #582, opened 2021-10-02
Activate test on 32bit windows

serge-sans-paille:


serge-sans-paille:
@JohanMabille the errors on windows 32bit are strange, does that ring any bell?

PR #612, opened 2021-10-19
Enable compiling arm/neon with MSVC for windows on arm64

niyas-sait:
This patch enables building arm/neon with MSVC compiler for windows on arm64 target and contains following changes,

- Replace the dispatcher mechanism and use an explicit function selection using scalar types. This is required as MSVC intrinsics uses the same underlying type for multiple neon vector types and function selection using target vector type causes the wrong function to be called.

- Add a function to convert `Initializer_list<batch<T>>` to neon vector type as there are no constructors provided for the same operation in MSVC.

- NEON/NEON64 identification using MSVC specific flags

- Add a `_` to intrinsics wrapper functions. MSVC defines some intrinsics using macros and without the prefix the wrapper function names get replaced by the pre-processor.

### DIFF
--- a/include/xsimd/config/xsimd_config.hpp
+++ b/include/xsimd/config/xsimd_config.hpp
@@ -213,6 +213,9 @@
     #else
         #define XSIMD_WITH_NEON64 0
     #endif
+#elif defined(_MSC_VER) && defined(_M_ARM64)
+    #define XSIMD_WITH_NEON     1
+    #define XSIMD_WITH_NEON64   1
 #else
     #define XSIMD_WITH_NEON 0
     #define XSIMD_WITH_NEON64 0
--- a/include/xsimd/arch/xsimd_neon.hpp
+++ b/include/xsimd/arch/xsimd_neon.hpp
@@ -19,52 +19,53 @@
 
 #include "../types/xsimd_neon_register.hpp"
 #include "../types/xsimd_utils.hpp"
+#include "xsimd_neon_dispatcher.hpp"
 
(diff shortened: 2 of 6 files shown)


nsait-linaro:
> Can you rebase your PR on the master branch please? This would make the diff easier to read (and would solve the conflicts).

Yes sure

nsait-linaro:
The change might be a bit invasive. Let me know if you need help with the review. 

Another approach to support MSVC would be to wrap the vector types in a custom class so that all types like float32x4_t and int32x4_t can be distinguished by the dispatcher but it would require too many conversions from the wrapper class to the native class before passing onto intrinsic functions etc which can be done with the user-defined operator but I guess performance cost will be there.

include/xsimd/config/xsimd_config.hpp
```diff
@@ -213,6 +213,9 @@
     #else
         #define XSIMD_WITH_NEON64 0
     #endif
+#elif defined(_MSC_VER) && defined(_M_ARM64)
```
amyspark, line 216: There's a specific section for MSVC workarounds at the bottom. Also, from [MSVC docs](https://docs.microsoft.com/en-us/cpp/intrinsics/arm-intrinsics?view=msvc-160#top) it is implied that NEON support is always available. Would it be possible to drop the 64-bit check as well?

pdeep854:
i am able to build xsimd using MSVC on windows on arm64
refer https://github.com/xtensor-stack/xsimd/pull/1285

AntoinePrv:
This should be solved now.

PR #628, opened 2021-11-02
fix fast_cast selection

### LINKED ISSUE #626: fast_cast is not used
Currently, batch_cast always selects the fallback implementation, because `kernel::detail::conversion_type` is called with plain (non-batch types), but expects batch types.

Fixing this improves performance significantly; I'll open a PR soon.

tomjnixon:
Fixes #626 

I'm not sure about formatting; it would be helpful to have a clang-format config.

The value of a test for this is a bit questionable since you can't see what implementation will actually use, so feel free to skip that.

### DIFF
--- a/include/xsimd/arch/generic/xsimd_generic_details.hpp
+++ b/include/xsimd/arch/generic/xsimd_generic_details.hpp
@@ -136,7 +136,9 @@ namespace xsimd {
 
       template <class A, class From, class To>
       struct conversion_type_impl<A, From, To,
-                void_t<decltype(fast_cast(std::declval<const From&>(), std::declval<const To&>(), std::declval<const A&>()))>>
+                void_t<decltype(fast_cast(std::declval<const batch<From, A>&>(),
+                                          std::declval<const batch<To, A>&>(),
+                                          std::declval<const A&>()))>>
       {
           using type = with_fast_conversion;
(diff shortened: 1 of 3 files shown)


tomjnixon:
working on that failure; it does seem like neon fast_cast (via converting load) is broken somehow

tomjnixon:
OK, i think I'm done with this; there were a couple of additional issues.

On NEON, `fast_cast` called `bitwise_cast`, which is the wrong type of cast. For now I've removed it; obviously there should be an implementation of this, but it's not something I have a need for, and removing it doesn't make anything worse (since it wasn't called anyway).

It's not really relevant to this, but there were some hairy compile failures for calls to NEON `bitwise_cast` for type combinations that aren't supported by the defined intrinsics. I guess this is OK since it seems like every sensible combination is supported? It probably depends on whether you can construct a `batch<double, neon>` -- I'm not sure.

I'm not really keen on the implementation of that; it seems like since the code has a list of all the conversion functions anyway you might as well use macros to generate overloaded functions directly.

serge-sans-paille:
Thanks *a lot* for that patch. you're correct when stating that testing that one is a bit tricky...

tomjnixon:
No worries, thanks for merging it quickly.

PR #631, opened 2021-11-04
Enables support for FMA when AVX2 is detected on Windows

### LINKED ISSUE #630: xsimd doesn't detect FMA on MSVC
:wave:

I noticed that with MSVC, FMA is never detected because it relies on the `__FMA__` macro, which is GCC and Clang only.

However, from [this StackOverflow question](https://stackoverflow.com/questions/34265982/automatically-generate-fma-instructions-in-msvc) as well as [this MSDN post](blogs.msdn.com/b/vcblog/archive/2014/02/28/avx2-support-in-visual-studio-c-compiler.aspx), MSVC is indeed able to generate FMA instructions with `/arch:AVX2`.

Would it be possible to add support for it, or should it be force enabled from the outside?

(I cannot supply a Godbolt as an example, because it refuses to compile xsimd.)

JohanMabille:
Fixes #630

### DIFF
--- a/include/xsimd/arch/xsimd_fma3.hpp
+++ b/include/xsimd/arch/xsimd_fma3.hpp
@@ -12,7 +12,7 @@
 #ifndef XSIMD_FMA3_HPP
 #define XSIMD_FMA3_HPP
 
-#include "../types/xsimd_sse_register.hpp"
+#include "../types/xsimd_fma3_register.hpp"
 
 namespace xsimd
 {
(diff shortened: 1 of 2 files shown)


tomjnixon:
Should this not enable FMA5? I think with /arch:AVX2 MSVC can generate 256 bit FMA instructions.

JohanMabille:
Ah indeed.

serge-sans-paille:
Can you rebase on master branch now that the clang-format patch is in?

JohanMabille:
sure

Issue #633, opened 2021-11-04
dispatch assumes that target architectures are supported


tomjnixon:
First, sorry for the long issue, or if it's wildly off-base.

Compiler architecture flags change both the set of instructions available (through intrinsics) and  the set of instructions used by normal code. This means that in general it is not safe to run a program compiled with `-mavx2` (for example) on a CPU that doesn't support AVX2, even if any code using AVX2 intrinsics is disabled at run-time, because the compiler is free to generate them for normal code too.

I've demonstrated this here: https://gcc.godbolt.org/z/Mvf5GxGGe

The way I've implemented this before is to compile vector functions multiple times for different architectures (producing different symbols), and link them with the rest of the application and dispatch code which is compiled for the lowest supported architecture.

This seems messy, but I think it's the only way to implement run-time dispatch reliably. I guess it's been fine up to now because most people are running a CPU supporting AVX2+FMA and aren't enabling options to generate avx512, but that could change.

This way of doing things should work fine with xsimd, but there's a couple of issues:

- `dispatch` checks that all architectures are supported, but when compiled with the lowest supported architecture they will not all be. Could this check just be removed?

  The default architecture list only contains supported architectures, and the dispatched function would fail to compile anyway. The downside would be worse error messages if you pass an unsupported architecture to `dispatch`, but it's only as bad as manually trying to use a `batch` with an unsupported architecture, and that could probably be made better with a `static_assert` in `batch` instead.

- The documentation doesn't mention this. Explaining it might be a bit tricky and off-putting, especially without an example and/or cmake support, so maybe that should come first?

amyspark:
> This means that in general it is not safe to run a program compiled with -mavx2 (for example) on a CPU that doesn't support AVX2, even if any code using AVX2 intrinsics is disabled at run-time, because the compiler is free to generate them for normal code too.

The only thing that xsimd's `Arch` template parameter changes is the register size. The compiler is free to optimize it to match the `-march` target, as shown here: https://gcc.godbolt.org/z/r5qxcYrj3.

> The way I've implemented this before is to compile vector functions multiple times for different architectures (producing different symbols), and link them with the rest of the application and dispatch code which is compiled for the lowest supported architecture.

Vc does a similar thing, it's the way we use at Krita to runtime dispatch correctly: https://github.com/VcDevel/Vc/blob/1.4/cmake/VcMacros.cmake#L447

[...]

Issue #634, opened 2021-11-04
cpuid test for FMA is disabled


tomjnixon:
The `best` calculation for this should probably check for `avx2` and `fma`, as the fma5 register type requires both.

Issue #632, opened 2021-11-04
dispatch uses lowest available architecture if best available is not in the arch list


tomjnixon:
Currently, `available_architectures().best` is used to select the function to be called by `dispatch`. This is the version number of the best architecture supported by the CPU.

If the best architecture is not in the `arch_list` passed to dispatch, all [`Arch::version() == best_arch`](https://github.com/xtensor-stack/xsimd/blob/f8c1a032673bbae2222d061bcb0e102d1a9ba930/include/xsimd/config/xsimd_arch.hpp#L196) tests fail, and so the lowest available architecture is used.

I guess this should just be changed to `Arch::version() <= best_arch`.

I've added a test here which passes for me on AVX2, but probably shouldn't: https://github.com/tomjnixon/xsimd/commit/dispatch_issue

I've got a couple of other issues with dispatching that I'll open separately, as i think they are independent.

PR #635, opened 2021-11-04
Fix dispatch issues

tomjnixon:
See #632, #633 and #634.

This doesn't cover the documentation or cmake aspects of #633; i think that's a separate job.

I'm not totally sure about the consequences of "add explicit check for architecture support in batch"; it could make for more mysterious error messages if simd_register is used directly for some reason, but I think it'll be alright.

### DIFF
--- a/include/xsimd/types/xsimd_register.hpp
+++ b/include/xsimd/types/xsimd_register.hpp
@@ -20,7 +20,10 @@ namespace xsimd
     {
 
         template <class T, class Arch>
-        struct simd_register;
+        struct simd_register
+        {
+            static_assert(Arch::supported(), "usage of simd_register with unsupported architecture");
+        };
 
         template <class T, class A>
         struct has_simd_register : std::false_type
--- a/include/xsimd/config/xsimd_arch.hpp
+++ b/include/xsimd/config/xsimd_arch.hpp
@@ -193,7 +193,7 @@ namespace xsimd
             auto walk_archs(arch_list<Arch, ArchNext, Archs...>, Tys&&... args) -> decltype(functor(Arch {}, std::forward<Tys>(args)...))
             {
                 static_assert(Arch::supported(), "dispatching on supported architecture");
-                if (Arch::version() == best_arch)
+                if (Arch::version() <= best_arch)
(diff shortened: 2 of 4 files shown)


include/xsimd/types/xsimd_register.hpp
```diff
@@ -20,7 +20,9 @@ namespace xsimd
     {
 
         template <class T, class Arch>
-        struct simd_register;
+        struct simd_register
```
serge-sans-paille, line 23: I <3 all of your changes. Can you explain that one?

tomjnixon: Ha, yes. Without that, the `static_assert` in `batch` never fires, because `simd_register` is not defined for unsupported architectures (there's only a forward declaration, which can't be inherited from).

That's why i'm not sure about it, as it means that `simd_register` can be constructed for unsupported architectures. I couldn't think of any way around this except for making it's constructor private, but that feels messy.

serge-sans-paille: What about putting the same static_assert in `simd_register` then?

tomjnixon: Oh yeah, that works. I tried this originally but got confused.

The difference now from the original is if you try to construct a batch with a bad type; you used to get

```
invalid use of incomplete type ‘struct xsimd::types::simd_register<double*, xsimd::sse2>’
```

whereas now you get

```
no type named ‘register_type’ in ‘struct xsimd::types::simd_register<double*, xsimd::sse2>’
```
I think it's fine; `simd_register` is presumably an internal thing anyway.

tomjnixon:
In this batch of changes I've:

- improved avx512 detection (tested with intel SDE)
- updated commit message in "don't check for supported architectures in dispatch" to mention, not close issue #633, as that might need more work
- moved the architecture check from `batch` to `simd_register`

I've also implemented the per-architecture build thing mentioned in #633 in my application using this -- seems to work ok!

The only thing missing is a way to get the maximum required alignment from an `arch_list`; I'll put in a separate PR for that.

tomjnixon:
Cheers!

PR #640, opened 2021-11-09
add alignment helper to arch_list

tomjnixon:
This is a follow-on from #635; this is useful when using per-architecture compilation and dynamic dispatch, as the required alignment when allocating data can be greater than the alignment of the default architecture.

### DIFF
--- a/include/xsimd/config/xsimd_arch.hpp
+++ b/include/xsimd/config/xsimd_arch.hpp
@@ -61,6 +61,18 @@ namespace xsimd
         {
         };
 
+        template <typename T>
+        constexpr T max_of(T value)
+        {
+            return value;
+        }
+
+        template <typename T, typename... Ts>
+        constexpr T max_of(T head0, T head1, Ts... tail)
+        {
+            return max_of((head0 > head1 ? head0 : head1), tail...);
+        }
+
     } // namespace detail
 
     // An arch_list is a list of architectures, sorted by version number.
@@ -88,6 +100,12 @@ namespace xsimd
         static void for_each(F&& f)
         {
             (void)std::initializer_list<bool> { (f(Archs {}), true)... };
+        }
+
+        static constexpr std::size_t alignment()
+        {
+            // all alignments are a power of two
+            return detail::max_of(Archs::alignment()..., static_cast<size_t>(0));
         }
     };
 
--- a/include/xsimd/types/xsimd_generic_arch.hpp
+++ b/include/xsimd/types/xsimd_generic_arch.hpp
@@ -23,6 +23,7 @@ namespace xsimd
     {
         static constexpr bool supported() { return true; }
         static constexpr bool available() { return true; }
+        static constexpr std::size_t alignment() { return 0; }
         static constexpr bool requires_alignment() { return false; }
 
     protected:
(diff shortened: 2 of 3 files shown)


tomjnixon:
Hopefully fixed the MSVC issue...

This seems a bit more complicated than it needs to be, to work around the lack of `alignment()` for `generic`. Things like `alignas` ignore alignment values of 0, so perhaps it would be easier if all architectures had it?

This might make other things easier down the line too, e.g. `default_allocator` doesn't currently work with `generic` because it uses `alignment()` directly.

include/xsimd/config/xsimd_arch.hpp
```diff
@@ -89,6 +129,11 @@ namespace xsimd
         {
             (void)std::initializer_list<bool> { (f(Archs {}), true)... };
         }
+
+        static constexpr std::size_t alignment()
+        {
+            return detail::max_alignment<Archs...>::value;
```
serge-sans-paille, line 135: Side note: theoretically, it should probably be the Least common multiple instead. Because we only have power of twos, max is faster, but maybe worth adding a comment?

include/xsimd/config/xsimd_arch.hpp
```diff
@@ -61,6 +61,46 @@ namespace xsimd
         {
         };
 
+        // requires_alignment() as a struct, required because MSVC doesn't
+        // support directly using constexpr functions for SFINAE
```
serge-sans-paille, line 65: Following one of yourcomment, I do think it would be simpler to always provide an alignment() method, which defaults to ``0``so that this FINAE is not needed.

include/xsimd/config/xsimd_arch.hpp
serge-sans-paille: I think this could ultimately be written as:
```
return details::max_of(Archs::alignment()...);
```

tomjnixon: alignas() causes a compile error if you give it a value that isn't a power of two (or zero), so I don't think a check is really needed for this (as the rest of the library will fail to compile too).

I've implemented it as you suggested, though I think it's a bit messy to use a constexpr function for this, as you end up having to have a separate type variable (in a parameter pack) for each element, and you can't use temporary variables.

I've put a version using template structs instead here:

https://github.com/xtensor-stack/xsimd/compare/master...tomjnixon:arch_list_alignment_struct

What do you think?

include/xsimd/config/xsimd_arch.hpp
```diff
@@ -61,6 +61,18 @@ namespace xsimd
...
+        template <typename T>
+        constexpr T max_of(T value)
+        {
+            return value;
+        }
+
+        template <typename T, typename... Ts>
+        constexpr T max_of(T head, Ts... tail)
+        {
+            return head > max_of(tail...) ? head : max_of(tail...);
```
serge-sans-paille, line 73: What about writing this as

```
template <typename T0, typename T1, typename... Ts>
constexpr T max_of(T0 arg0, T1 arg1, Ts... tail) {
  return max_of((arg0 > arg1 ? arg0 : arg1), tail...);
}
```
that would avoid the double call of max_of. Looks good otherwise

tomjnixon: Thanks, I made that change.

serge-sans-paille:
Cool!

PR #642, opened 2021-11-16
improve code generated for AVX2 signed integer comparisons

tomjnixon:
Previously, AVX2 `gt` was implemented, but `lt` fell back to the AVX `lt`, which is implemented with SSE2 instructions.

generic_logical has the following mappings:

 - `gt` -> `lt`
 - `ge` -> `le`
 - `le` -> `lt || eq`

so it seems best to just implement `eq`, `lt`, and `le` if it's available.

I've had a brief look at the SSE and NEON comparison code and they don't seem to be affected by this.

It's tempting to go through and delete all the `gt` and `ge` implementations as they don't seem to add anything, but I feel like I might be missing something.

Here's an example showing the issue: https://gcc.godbolt.org/z/Gd3K9WfTG

### DIFF
--- a/include/xsimd/arch/xsimd_avx.hpp
+++ b/include/xsimd/arch/xsimd_avx.hpp
@@ -569,44 +569,6 @@ namespace xsimd
             return _mm256_floor_pd(self);
         }
 
-        // ge
-        template <class A>
-        inline batch_bool<float, A> ge(batch<float, A> const& self, batch<float, A> const& other, requires_arch<avx>)
-        {
-            return _mm256_cmp_ps(self, other, _CMP_GE_OQ);
-        }
-        template <class A>
-        inline batch_bool<double, A> ge(batch<double, A> const& self, batch<double, A> const& other, requires_arch<avx>)
-        {
-            return _mm256_cmp_pd(self, other, _CMP_GE_OQ);
-        }
-        template <class A, class T, class = typename std::enable_if<std::is_integral<T>::value, void>::type>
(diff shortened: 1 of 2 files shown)


tomjnixon:
I've put some comparisons of the assembly before and after here: https://gist.github.com/tomjnixon/ad1c8bdb958a3f04385778b79013d06f

This seems to pessimise `ge` a bit on avx when compiling with clang, though i can't see why that would be; it looks like there's an optimisation swapping `x > y || x == y` for `!(x < y)` that doesn't happen for other combinations. I guess might be worth adding this explicitly, or swapping the fallback `le` implementation to use this in general, as I think it's usually better on X86

Clang also generates some odd left and right shift sequences after some comparison operations, but that's not really relevant here.

serge-sans-paille:
This all looks good, thanks for tracking this one!

serge-sans-paille:
> I've put some comparisons of the assembly before and after here: https://gist.github.com/tomjnixon/ad1c8bdb958a3f04385778b79013d06f
> 
> This seems to pessimise `ge` a bit on avx when compiling with clang, though i can't see why that would be; it looks like there's an optimisation swapping `x > y || x == y` for `!(x < y)` that doesn't happen for other combinations. I guess might be worth adding this explicitly, or swapping the fallback `le` implementation to use this in general, as I think it's usually better on X86

+1 fore adding explicit `le` that would avoid fallingback to `lt` || `eq`. Can you do this?

PR #645, opened 2021-11-20
Feature/noexcept

serge-sans-paille:


include/xsimd/arch/generic/xsimd_generic_arithmetic.hpp
```diff
@@ -27,7 +27,7 @@ namespace xsimd
 
         // bitwise_lshift
         template <class A, class T, class /*=typename std::enable_if<std::is_integral<T>::value, void>::type*/>
-        inline batch<T, A> bitwise_lshift(batch<T, A> const& self, batch<T, A> const& other, requires_arch<generic>)
+        inline batch<T, A> bitwise_lshift(batch<T, A> const& self, batch<T, A> const& other, requires_arch<generic>) noexcept
         {
             return detail::apply([](T x, T y)
                                  { return x << y; },
```
marzer, line 33: I think for MSVC to truly benefit here you'd need to also apply it to any lambdas too, e.g.
```cpp
return detail::apply([](T x, T y) noexcept
                                 { return x << y; },
```

(unlike `constexpr`, lambdas [aren't automatically `noexcept`](https://www.reddit.com/r/cpp_questions/comments/lusx7q/are_lambdas_automatically_noexcept/))

serge-sans-paille: Oh Joy :-)

include/xsimd/xsimd.hpp
```diff
@@ -18,6 +18,14 @@
 #define XSIMD_NO_DISCARD
 #endif
 
+#if defined(__GNUC__)
+#define XSIMD_INLINE inline __attribute__((always_inline))
+#elif defined(_WIN32)
+#define XSIMD_INLINE inline __forceinline
+#else
+#define XSIMD_INLINE inline
+#endif
```
marzer, line 27: `_WIN32` is probably the wrong thing to check for `__forceinline` support since compilers will define that for simply indicating a Windows target (and even some misguided developers will define for general windows compatibility reasons), regardless of whether support for `__forceinline` is available. An MSVC-specific macro, like `_MSC_VER`, is probably what you want to check for instead.

Additionally you may want to guard this whole construct inside an `#ifndef XSIMD_INLINE` block, as some people with more specialized needs may not want everything force-inlined and will want to override this behaviour...

amyspark: > An MSVC-specific macro, like `_MSC_VER`, is probably what you want to check for instead.

This is correct, MSYS2-based compilers just work by accident-- because they get checked first.

> Additionally you may want to guard this whole construct inside an #ifndef XSIMD_INLINE block, as some people with more specialized needs may not want everything force-inlined and will want to override this behaviour...

In the case of GCC and Clang, forcing the inlining isn't even necessary, it is automatically done on `-O2` and higher. So you may just force it on MSVC.

serge-sans-paille:
Switched to use `_MSC_VER `. Not providing users the capability to change `XSIMD_INLINE` behavior as of now, I'm not convinced by this use case.

amyspark:
@serge-sans-paille, there's a performance drop between 8.0.3 (+ an application of `s/inline /__forceinline`) and this PR. I think you may have overinlined functions; so far I know that f9d3b5dbea6f6995af64b5e3232f5d9071f0cbcf introduces this drop, and the application of inlining to the friend batch functions does the rest.

I will let you know when/if I am able to find the exact set of offenders.

serge-sans-paille:
Let's move the inline discussion to #646 and keep that PR for noexcept then. I think there's a consensus on `noexcept` doesn't hurt.

PR #646, opened 2021-11-22
Feature/all inline

serge-sans-paille:


amyspark:
The performance bug has been reported to MSVC [here](https://developercommunity.visualstudio.com/t/2x-performance-loss-when-using-__forcein/1592199).

tomjnixon:
@amyspark having a look at your benchmark, have you had a look at the assembly before/after to see what's going on?

Random speculation: inlining xsimd functions could be causing `AlphaDarkenOp::operator()` to not get inlined into the `testCompositionSpeed` loop (if inlining is just based on code size for example), which would be terrible for performance as there's a lot of code in there which would otherwise get pulled out of the loop. The same could well apply to some of the other non-force-inline functions you're using.

amyspark:
@serge-sans-paille upon further review, it seems that, instead of e.g. shifting a register right then using the result, MSVC spills the register on the stack, loads it, shifts, pushes and then pops it for the next operation.

amyspark:
A tad curious: would it be possible to make a test branch that replaces all constant pass-by-references with pass-by-values? IIRC MSVC was quite sensitive to references and doesn't realise they can be inlined without the need for `vmovups` and the like.

serge-sans-paille:
> A tad curious: would it be possible to make a test branch that replaces all constant pass-by-references with pass-by-values? IIRC MSVC was quite sensitive to references and doesn't realise they can be inlined without the need for `vmovups` and the like.

Before investigating, can *you* do test that for your test case and tell us if it changes something? If so I'm ok to do the change

marzer:
FWIW when switching our project to this branch we saw speedups of 5% - 10% across the board (targeting AVX, compiled with MSVC 19.31 (VS2022)).

serge-sans-paille:
Obsoleted by #1014

Issue #608, opened 2021-10-15
How to write code that agnostically targets SSE, NEON and AVX in xsimd 8


boris-kuz:
Hey everyone!

We noticed that the second template argument of `xsimd::batch` was changed from the width to the target architecture from version 7.x to 8.
This raises a few questions:

1. Does the runtime dispatching still work? The target architecture is specified at compile time, so how is this compatible with the runtime-dispatching approach?
2. Suppose you have a program that is written with batches of width 4 in mind. How do you make sure this code correctly compiles and runs on platforms supporting SSE4.2, NEON (ARM) and AVX? If you rely on the automatic architecture detection, switching from SSE4.2 to AVX will break the code (when using `xsimd::hadd` for example). If you try to circumvent this by specifying the architecture by saying `xsimd::batch<float, xsimd::sse4_2>`, it won't compile on ARM/with NEON. So what would the approach here be?

Thanks!

serge-sans-paille:
The following snippet extracts the best type to hold 4 floats among the available architectures (and dumps the architecture on stdout)

```
#include <xsimd/xsimd.hpp>

namespace details {
template<class ArchList>
struct batch4f;

template<>
struct batch4f<xsimd::arch_list<>> {
  using type = void;
};

template<class Arch, class... Archs>
struct batch4f<xsimd::arch_list<Arch, Archs...>> {
  using type = typename std::conditional<
    xsimd::batch<float, Arch>::size == 4,
    xsimd::batch<float, Arch>,
    typename batch4f<xsimd::arch_list<Archs...>>::type>::type;
};
}

using batch4f = typename details::batch4f<xsimd::supported_architectures>::type;

#include<iostream>
int main() {
  std::cout << batch4f::arch_type::name() << std::endl;
  return 0;
}
```

@JohanMabille It may be worth providing such aliases in the core library.

boris-kuz:
Thanks, this looks great! Correct me if I'm wrong though - this only resolves the type at compile time, right?
The doc states 

> "xsimd makes it easy to write a single algorithm, generate one version of the algorithm per micro-architecture and pick the best one at runtime, based on the running processor capability."

So how does the runtime dispatching come into play here then? I don't really understand the point or the mechanics of specifying an architecture at compile them if the library is supposed to pick the "best intrinsics" at runtime?

JohanMabille:
> @JohanMabille It may be worth providing such aliases in the core library.

Agreed.

[...]

PR #647, opened 2021-11-22
added arch-agnostic fixed-size type aliases and tests

boris-kuz:
Taken from @serge-sans-paille mentioned here: https://github.com/xtensor-stack/xsimd/issues/608


serge-sans-paille:
@boris-kuz I was not very happy with the naming convention for the fixed-size batch aliases, so decided that it should be a user thing. Alternate implementation (more generic but that gives end-user the choice over the type alias name) in #674
Would that be ok for you?

boris-kuz:
Oh yes, that implementation is much nicer! Thanks serge!

PR #651, opened 2021-11-29
FMA5 defined with AVX instead of FMA3

edponce:
Updates the configuration checks for FMA-enabled ISA.
It allows the use of FMA3 with SSE4.2 or AVX, and FMA5 with AVX2.

### DIFF
--- a/include/xsimd/config/xsimd_config.hpp
+++ b/include/xsimd/config/xsimd_config.hpp
@@ -123,7 +123,7 @@
  */
 #ifdef __FMA__
 
-#if defined(__AVX2__)
+#ifdef __AVX2__
 #define XSIMD_WITH_FMA5 1
 #else
 #define XSIMD_WITH_FMA5 0
@@ -241,13 +241,13 @@
 #if !defined(__clang__) && (defined(_M_X64) || (defined(_M_IX86_FP) && _M_IX86_FP >= 2))
 #undef XSIMD_WITH_SSE4_2
 #define XSIMD_WITH_SSE4_2 1
-#undef XSIMD_WITH_FMA3
-#define XSIMD_WITH_FMA3 1
 #endif
 
 #if XSIMD_WITH_SSE4_2
 #undef XSIMD_WITH_SSE4_1
 #define XSIMD_WITH_SSE4_1 1
+#undef XSIMD_WITH_FMA3
+#define XSIMD_WITH_FMA3 1
 #endif
 
 #if XSIMD_WITH_SSE4_1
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
The problem with fma3 and fma4 is that these instruction sets impact both 128 and 256 bit registers, which is incompatible with xsimd model. So we defined two "virtual" instructions sets, that we probably named badly: one for SSE+FMA3 and one for AVX+FMA. Maybe we should name them differently, like

`XSIMD_SSE_FMA3` and `XSIMD_AVX_FMA3` ?

edponce:
Currently, [xsimd defines `fm5 = AVX2 + FMA`](https://github.com/xtensor-stack/xsimd/blob/master/include/xsimd/types/xsimd_fma5_register.hpp#L25) but the [configuration checks for `AVX + FMA`](https://github.com/xtensor-stack/xsimd/blob/master/include/xsimd/config/xsimd_config.hpp#L124-L127).

In terms of ISA features, there is FMA (aka FMA3) and FMA4. Based on this, I think the term FMA5 should be deprecated. FMA features are independent of SIMD ISA so I understand it is not trivial to adapt xsimd to it. AFAIK, FMA4 is not considered by xsimd, so having a generic `xsimd::fma3` that adapts to any SIMD ISA seems reasonable.

The configuration macros need to be updated and consider the cases where `-mfma` flag enables `__AVX__` automatically.

edponce:
I will update this PR based on previous discussion.

serge-sans-paille:
> In terms of ISA features, there is FMA (aka FMA3) and FMA4. Based on this, I think the term FMA5 should be deprecated. 
Correct

> FMA features are independent of SIMD ISA so I understand it is not trivial to adapt xsimd to it. Having a generic `xsimd::fma3` that adapts to any SIMD ISA seems reasonable.

I'm glad we have common ground there ;-)

> The configuration macros need to be updated and consider the cases where `-mfma` flag enables `__AVX__` automatically.
Can you be more explicit on that one?

edponce:
At least in Clang and GCC compilers, using `-mfma` flag automatically defines `__AVX__` even if `-mavx` was not set explicitly.
For example, these set of flags `-msse4.2 -mfma` define `__SSE4_2__, __FMA__, __AVX__`, so [these checks](https://github.com/xtensor-stack/xsimd/blob/master/include/xsimd/config/xsimd_config.hpp#L107-L134) need to be reviewed alongside the FMA3 consolidation.

serge-sans-paille:
Obsoleted by #660

PR #657, opened 2021-12-16
Fix xsimd::transform for complex types.

### LINKED ISSUE #652: `xsimd::transform` no longer works with complex types in xsimd 8
`xsimd::transform` no longer works in xsimd 8. Here is a short example of code that previously worked in xsimd 7 but no longer compiles:

```cpp
std::vector<std::complex<float>, xsimd::aligned_allocator<std::complex<float>>> vec(100);

xsimd::transform(vec.begin(), vec.end(), vec.begin(),
    [&](const auto& x) { return x; });
```

One fix is to add non-member `store_aligned` and `store_unaligned` definitions. For example, adding the following definitions in `xsimd/types/xsimd_batch.hpp` fixes compilation:

```cpp
template<class A, class T>
void store_aligned(std::complex<T> *mem, batch<std::complex<T>, A> const& self)
{
    self.store_aligned(mem);
}

template<class A, class T>
void store_unaligned(std::complex<T> *mem, batch<std::complex<T>, A> const& self)
{
    self.store_unaligned(mem);
}
```

Altering the definition of `xsimd::transform` to use the static member `store_*` functions would also work, but having the definitions above is probably more generally useful.

I'm not sure where the right place to put this code is, or even if it's the right fix, so I hope you will excuse the lack of a pull request.

Thanks!

mainland:
This fixes #652. New tests are included.

### DIFF
--- a/include/xsimd/stl/algorithms.hpp
+++ b/include/xsimd/stl/algorithms.hpp
@@ -47,7 +47,7 @@ namespace xsimd
             for (std::size_t i = align_begin; i < align_end; i += simd_size)
             {
                 batch_type batch = batch_type::load_aligned(&first[i]);
-                xsimd::store_aligned(&out_first[i], f(batch));
+                f(batch).store_aligned(&out_first[i]);
             }
 
             for (std::size_t i = align_end; i < size; ++i)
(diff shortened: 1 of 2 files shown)


serge-sans-paille:
This looks good. thanks for the extra testing!

Issue #622, opened 2021-10-31
Document to_float and to_int and support casts to/from unsigned types


amyspark:
Hi!

I came upon this library while researching replacements for Vc. So far it's worked great, minus a few kinks here and there. However, the lack of intrinsics-backed casts produces a ~5x worse compute times on MSVC between Vc and xsimd-based implementations, even while accounting for hidden casts inside Vc. This performance hit can go as far as 100x if `/Ob1` or lower are used.

As an example:

```
template <bool aligned, class U = typename std::conditional<aligned, xsimd::aligned_mode, xsimd::unaligned_mode>::type>
static inline float_v fetch_alpha_32(const void *data)
{
  uint_v data_i =
      uint_v::load(reinterpret_cast<const uint_v::value_type *>(data), U{});
  return xsimd::batch_cast<float>(data_i >> 24); // this conversion goes through batch_cast(with_slow_conversion)
}
```

Instead of doing an aligned store, a compiler-managed copy, and an aligned load, it could be something as simple as `_mm256_cvtepi32_ps(self)`.

serge-sans-paille:
mmmh, actually, `xsimd` can generate call to `_mm256_cvtepi32_ps`, but only when converting *signed* integers to float, and not *unsigned* integers. Which is definitively the semantic of [`_mm256_cvtepi32_ps`](https://www.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/compiler-reference/intrinsics/intrinsics-for-intel-advanced-vector-extensions/intrinsics-for-conversion-operations-2/mm256-cvtepi32-ps.html) but we could do better. Let me try to fix that.

Meanwhile, if you can force the cast of `data_i >> 24` to `int_32` is possible.

serge-sans-paille:
(And note that although usage of `/ObN` family option is arguable, #620 fixes interactions of `xsimd` with `/Ob1`)

amyspark:
[...]

PR #667, opened 2022-01-21
Provide better  conversion functions

serge-sans-paille:
Related to #622


serge-sans-paille:
cc @JohanMabille I'm very happy of this kind of update that removes redundant code :-)

PR #670, opened 2022-01-27
Only use _mm256_unpacklo_epi<N> on AVX2

### LINKED ISSUE #668: AVX zip_lo/hi incorrectly uses AVX2 instructions
👋  MWE:

```cpp
#include <xsimd/xsimd.hpp>

using float_v = xsimd::batch<float>;
using uint_v = xsimd::batch<unsigned int>;
using int_v = xsimd::batch<int>;

std::pair<int_v, int_v> interleave(const int_v &a, const int_v &b)
{
    return {xsimd::zip_lo(a, b), xsimd::zip_hi(a, b)};
}
```

When built with `c++ -O2 -mavx` (clang 13 on MSYS; note it's *not* AVX2): 

```
In file included from C:/benchmark-xsimd/main2.cpp:1:
In file included from C:/include/xsimd/xsimd.hpp:29:
In file included from C:/include/xsimd/config/../types/../types/xsimd_batch.hpp:384:
In file included from C:/include/xsimd/config/../types/../arch/xsimd_isa.hpp:48:
C:/include/xsimd/arch/./xsimd_avx.hpp:1184:24: error: always_inline function '_mm256_unpacklo_epi32' requires target feature 'avx2', but would be inlined into function 'zip_lo' that is compiled without support for 'avx2'
                return _mm256_unpacklo_epi32(self, other);
                       ^
C:/include/xsimd/arch/./xsimd_avx.hpp:1154:24: error: always_inline function '_mm256_unpackhi_epi32' requires target feature 'avx2', but would be inlined into function 'zip_hi' that is compiled without support for 'avx2'
                return _mm256_unpackhi_epi32(self, other);
```

Checked in xsimd 8.0.3, and also in commit 315f7c11751bab0771478f48c4197b4cf9b50404.

serge-sans-paille:
As a workaround, rely on cast from float to support epi32 and epi64 on AVX.

Should fix #668

### DIFF
--- a/include/xsimd/arch/xsimd_avx.hpp
+++ b/include/xsimd/arch/xsimd_avx.hpp
@@ -1146,6 +1146,37 @@ namespace xsimd
         {
             switch (sizeof(T))
             {
+            case 1:
+            case 2:
+            {
+                // extract low word
+                __m128i self_lo = _mm256_extractf128_si256(self, 0);
+                __m128i self_hi = _mm256_extractf128_si256(self, 1);
+                // extract high word
+                __m128i other_lo = _mm256_extractf128_si256(other, 0);
+                __m128i other_hi = _mm256_extractf128_si256(other, 1);
+
+                // interleave
+                __m128i res_lo, res_hi;
+                if (sizeof(T) == 1)
+                {
+                    res_lo = _mm_unpackhi_epi8(self_lo, other_lo);
+                    res_hi = _mm_unpackhi_epi8(self_hi, other_hi);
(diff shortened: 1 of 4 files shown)


amyspark:
@serge-sans-paille I think you can still do `zip_lo` and  `zip_hi` on AVX? Though with some elbow grease, e.g. https://godbolt.org/z/fE49j6a7s

serge-sans-paille:
@amyspark thanks for the hint, and... done ;-)

PR #674, opened 2022-02-16
Provide type utility to explicitly require a batch of given size and …

### LINKED ISSUE #608: How to write code that agnostically targets SSE, NEON and AVX in xsimd 8

serge-sans-paille:
…type

Through xsimd::make_sized_batch<T, N>

Fix #608

### DIFF
--- a/.azure-pipelines/azure-pipelines-win.yml
+++ b/.azure-pipelines/azure-pipelines-win.yml
@@ -55,8 +55,7 @@ jobs:
           setlocal EnableDelayedExpansion
           call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x86_amd64
           mkdir build & cd build
-          cmake -G Ninja ^
-                -DCMAKE_BUILD_TYPE=Release ^
+          cmake -DCMAKE_BUILD_TYPE=Release ^
                 -DCMAKE_C_COMPILER=clang-cl ^
                 -DCMAKE_CXX_COMPILER=clang-cl ^
                 -DDOWNLOAD_GTEST=ON ^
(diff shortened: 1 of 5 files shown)


cyb70289:
We have some code like
```
#ifdef ARROW_HAVE_NEON
  using simd_batch = xsimd::batch<int8_t, xsimd::neon64>;
#else
  using simd_batch = xsimd::batch<int8_t, xsimd::sse4_2>;
#endif
```
Will this PR simplify the code to
`using simd_batch = xsimd::make_sized_batch_t<int8_t, 16>`?

serge-sans-paille:
> Will this PR simplify the code to `using simd_batch = xsimd::make_sized_batch_t<int8_t, 16>`?

yep. And then depending on available architectures `simd_batch` will hold an actual `batch` (could also be `batch<int8_t, sse2>` but never batch<int8_t, avx>`) or `void` (if no support is found, and you can `static_assert` on that case if you want to be safe).

PR #681, opened 2022-03-02
Document batch_constant

### LINKED ISSUE #680: Documentation : batch_constant
Hi xsimd team.

I want to make a little suggestion. I had to use the swizzle function lately for a project, and I had a hard time creating the `batch_constant` needed to perform the swizzle. I had to refer to the test file [here](https://github.com/xtensor-stack/xsimd/blob/master/test/test_swizzle.cpp). It would be imo a good idea to add an example in the documentation.

My example was for a rotation, which is often used, so it could be a good example maybe you can reuse the code :
```
namespace xs = xsimd;
using b_type = xs::batch<float, xs::avx2>;


template <class T>
struct as_index
{
    using type = xsimd::as_unsigned_integer_t<T>;
};


struct Rot
{
    static constexpr unsigned get(unsigned i, unsigned n)
    {
        return (i + n - 1) % n;
    }
};

int main()
{
b_type a;
b_type b;
// ...
a = xs::swizzle(b, xsimd::make_batch_constant<typename as_index<b_type>::type, Rot>());
}
```

BR and thanks for your project.

serge-sans-paille:
Fix #680

### DIFF
--- a/docs/Doxyfile
+++ b/docs/Doxyfile
@@ -2,6 +2,7 @@ PROJECT_NAME      = "xsimd"
 XML_OUTPUT        = xml
 INPUT             = ../include/xsimd/types/xsimd_api.hpp \
                     ../include/xsimd/types/xsimd_batch.hpp \
+                    ../include/xsimd/types/xsimd_batch_constant.hpp \
                     ../include/xsimd/config/xsimd_arch.hpp \
                     ../include/xsimd/config/xsimd_config.hpp \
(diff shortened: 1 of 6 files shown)


serge-sans-paille:
@bonben does the new documentation looks reasonable to you (at least its Doxygen version :-/) ?

serge-sans-paille:
@JohanMabille same remark on documentation rerendering for that one ;-)

PR #682, opened 2022-03-04
Fix some ambiguous doxygen reference

serge-sans-paille:
thanks to @klagrassa and her comrade for spotting this.


serge-sans-paille:
@JohanMabille once merged, a rerendering of the RTD is needed (it's currently broken for several functions)

PR #688, opened 2022-03-14
Make xsimd::pow implementation resilient to FE_INVALID

### LINKED ISSUE #684: pow broken for 0.
Hi!

Thanks for the fantastic library! I'm working on getting xsimd integrated into our code to make sure we won't have any bugs. One thing I've come across is that the implementation of `pow` doesn't work for `0.` when floating point exceptions are enabled. Specifically, I'm doing:

```cpp
const xs::batch<double> a{3., 2., 1., 0.};
const xs::batch<double> b{xs::batch<double>::broadcast(2.)};
const xs::batch<double> c = pow(a, b);
```

As far as I can tell this is caused by the use of `log(abs(self))` in https://github.com/xtensor-stack/xsimd/blob/1cc996ef1c2a0b696acb21d65bc9ed67516a4e07/include/xsimd/arch/generic/xsimd_generic_math.hpp#L1899

This specifically triggers `FE_INVALID`. A simple fix seems to be:

```
batch_type z = exp(other * log(abs(self)
                              + std::numeric_limits<T>::min()));
```

This also works if `FE_UNDERFLOW` is enabled. Some specialization for floating point types might be necessary if the same code is used for integer exponentiation. Of course, this assumes `other` is positive. If not, then I think it should report `nan`.

I'll report any more if I find them (and try to at least suggest a solution!).

Thank you so much!

serge-sans-paille:
Fix #684

### DIFF
--- a/test/test_power.cpp
+++ b/test/test_power.cpp
@@ -14,6 +14,8 @@
 
 #include "test_utils.hpp"
 
+#include <cfenv>
+
 template <class B>
 class power_test : public testing::Test
 {
@@ -24,6 +26,7 @@ protected:
     using vector_type = std::vector<value_type>;
 
     size_t nb_input;
+    vector_type zlhs_input;
     vector_type lhs_input;
     vector_type rhs_input;
     vector_type expected;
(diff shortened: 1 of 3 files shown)


nilsdeppe:
This is great! Thank you so much for the quick fix. We really appreciate it!

serge-sans-paille:
Unfortunately, there' seems to be another issue lying around :'(

nilsdeppe:
Means more debugging, but uncovering bugs is something I always try to look at optimistically! It makes your resulting code better---something very welcome :D

test/test_power.cpp
```diff
@@ -61,6 +63,27 @@ class power_test : public testing::Test
...
+            std::transform(lhs_input.cbegin(), lhs_input.cend(), rhs_input.cbegin(), expected.begin(),
+                           [](const value_type& l, const value_type& r)
+                           { return std::pow(0 * l, r); });
+            batch_type lhs_in, rhs_in, out;
+            for (size_t i = 0; i < nb_input; i += size)
+            {
+                detail::load_batch(lhs_in, lhs_input, i);
+                detail::load_batch(rhs_in, rhs_input, i);
+                // Test with FE_INVALID...
+                feenableexcept(FE_INVALID);
```
JohanMabille, line 77: This is a glibc extension which is not available on Windows (thus the failure)

serge-sans-paille:
Finally! Greeeeeeen!

nilsdeppe:
🎉 thank you so much :D

PR #690, opened 2022-03-27
Provide xsimd::insert to modify a single value from a vector

### LINKED ISSUE #658: Indexing and replacing a single batch element in xsimd8.
Hi all, 
Since version 8, I am missing the handful `operator[]` functionality. In version 7, apart from providing a way of indexing batches, it would also provide a l-value, so you could write something like, 
```
xsimd::batch<int, 4> vInt4{1,2,3,4};
vInt4[0] = 2;
```
I wonder what's the best way to get something similar in version 8. As suggested in the migration guide, one could store the register into an array, modify the array as desired, in this case replacing a single element of the array, and then loading the modified array into the batch again.
That would work, but I am wondering if there is a way to get a `l-value` with `xsimd::batch<T>::get(size_t)`.

Thank you!!

serge-sans-paille:
Fix #658

### DIFF
--- a/include/xsimd/arch/xsimd_avx.hpp
+++ b/include/xsimd/arch/xsimd_avx.hpp
@@ -25,6 +25,10 @@ namespace xsimd
     {
         using namespace types;
 
+        // fwd
+        template <class A, class T, size_t I>
+        inline batch<T, A> insert(batch<T, A> const& self, T val, index<I>, requires_arch<generic>) noexcept;
+
         namespace detail
         {
             inline void split_avx(__m256i val, __m128i& low, __m128i& high) noexcept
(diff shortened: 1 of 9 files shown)


serge-sans-paille:
I still need to check if set1ps + shuffle is more efficient than broadcast + blend. @JohanMabille I provided a generic version so the arm test should pass, but there may be a proper neon way to do that

amyspark:
@serge-sans-paille gather/scatter, IIRC, is SVE's purview.

amyspark:
@serge-sans-paille Vc uses broadcast+blend on sse4.1, iirc. Can't compare with ARM because Vc doesn't support it.

serge-sans-paille:
@amyspark : yeah, I did a few test and broadcast + blend is better than set1 + suffle (sometimes two shuffles are needed depending on the position)

serge-sans-paille:
Ready for merging, cc @JohanMabille

include/xsimd/arch/xsimd_avx.hpp
```diff
@@ -680,6 +684,25 @@ namespace xsimd
             return _mm256_add_pd(tmp1, tmp2);
         }
 
+        // insert
+        template <class A, class T, size_t I, class = typename std::enable_if<std::is_integral<T>::value, void>::type>
+        inline batch<T, A> insert(batch<T, A> const& self, T val, index<I> pos, requires_arch<avx>) noexcept
+        {
+            switch (sizeof(T))
+            {
+#ifndef _WIN32
```
amyspark, line 693: Why doesn't this apply on Windows? Did you intend `_MSC_VER` @serge-sans-paille ?

serge-sans-paille: > Why doesn't this apply on Windows? Did you intend `_MSC_VER` @serge-sans-paille ?

This was failing using the windows compiler, and also with clang-cl due to invalid intrinsics, if I recall correctly. Feel free to propose a patch and investigate :-)

PR #695, opened 2022-03-30
Make error_gamma_test/arm_double.gamma pass on Apple M1

Yuhta:


test/test_utils.hpp
```diff
@@ -283,7 +283,7 @@ namespace detail
                 return utils::isinf(rhs) && (lhs * rhs > 0) /* same sign */;
             }
 
-            T relative_precision = 2048 * std::numeric_limits<T>::epsilon();
```
serge-sans-paille, line 286: I'd rather have this check specific to your target, can you update you patch to be specific to APPLE M1?
If you feel brave, setting up an extra CI test for that target would be great...

PR #698, opened 2022-04-05
Test broader clang versions (from version 6 to version 14)

emmenlau:


emmenlau:
Well, I wanted to only add more versions of clang to the build. But in the course of that, I found quite a number of issues and errors in the build scripts. I hope they are all fixed now, please review?

.azure-pipelines/azure-pipelines-linux-clang.yml
```diff
@@ -14,35 +14,41 @@ jobs:
           arm_arch_target: 'armv8-a'
         clang_6:
           llvm_version: '6.0'
-        clang_7:
```
serge-sans-paille, line 17: Why disabling this one and the one below?

emmenlau: There was no specific reason other than that I meant to be "careful with resources". The total number of build jobs remains the same, and I removed only intermediate versions, in the hope to still cover a broad range of clang versions. If you do not mind the extra CPU hours I can add the jobs back?

emmenlau: But personally I would rather add more diverse compilers like an Android build or other diverse platforms...

serge-sans-paille: +1, just wanted to know the motivation. Thanks for the cleanup!

emmenlau: Great!

emmenlau:
Hey @serge-sans-paille , thanks a lot for considering this PR so quickly! I think (and hope) it improves the CI a bit.

A related question: It seems at least possible that the other related projects (xtl, xtensor, xtensor-blas, etc) suffer from similar problems?! I did not check. But there where a number of typical copy-paste-errors in the azure pipeline yml, so its likely that other projects have similar code? Is that something you could check for, or should I propose something?

serge-sans-paille:
I'm not involved in other projects, so if you feel like checking their CI
config, please go ahead :-)

PR #702, opened 2022-04-07
Enable riscv64 in build system

tekkamanninja:
Signed-off-by: Wei Fu <wefu@redhat.com>


serge-sans-paille:
Thanks for the PR! What are you trying to achieve, this does not add any actual support, it will just allow for the scalar version, right?

tekkamanninja:
> Thanks for the PR! What are you trying to achieve? This does not add any actual support, it will just allow for the scalar version, right?

Hi ,
We are trying to build it on RISC-V (for Fedora 36), but this package failed because of that it miss  a branch for riscv64 :-)
Once we add this patch, we built this on risc64 successfully 

so maybe I should say: "enable riscv64 build"?

would you mind to correct me on this  :-)

serge-sans-paille:
Thanks for the details! Yeah, please amend your commit to reflect that aspect, other wise looks good :-)

serge-sans-paille:
Thanks o/

PR #704, opened 2022-04-08
Removed useless overload

JohanMabille:


serge-sans-paille:
Indeed, it's the same as the generic one. Less code, more happy developers \o/

PR #706, opened 2022-04-10
Add gather/scatter functions to xsimd::batch

### LINKED ISSUE #106: scatter and gather instructions
It could be interesting to support intels scatter and gather instructions (however, scatter does seem to be available only since AVX512).

amyspark:
Hi!

This PR is to propose adding five new functions to xsimd. I've used these in the (still WIP) port from Vc to xsimd, which you can see [here](https://invent.kde.org/graphics/krita/-/merge_requests/1404).

These functions are *(crossing out already upstreamed ones, check the mentions)*:

- ~~`xsimd::reciprocal`: calculates the approximate reciprocal of the SIMD register.~~
- ~~`xsimd::none`: a typo-free way to say `!xsimd::any` more clearly.~~
- `xsimd::gather` and `xsimd::scatter`: for <AVX512F, templates gathering using the new `xsimd::insert` function. (It will also take advantage of #691 when it's fixed)
- ~~`xsimd::round_to_even`: uses intrinsics to achieve fast rounding to even *integers*.~~

~~These do not include tests yet, as I'd like first to gather some feedback, especially regarding the approach I've followed and potential performance issues.~~ I've added the relevant test cases.

Thanks in advance for all your comments!

Fixes #106

### DIFF
--- a/include/xsimd/arch/generic/xsimd_generic_memory.hpp
+++ b/include/xsimd/arch/generic/xsimd_generic_memory.hpp
@@ -89,8 +89,8 @@ namespace xsimd
         gather(U const* src, batch<V, A> const& index,
                kernel::requires_arch<generic>) noexcept
         {
-            static_assert(batch<T, A>::size <= batch<V, A>::size,
-                          "Not enough indexes for gathering!");
+            static_assert(batch<T, A>::size == batch<V, A>::size,
+                          "Index and destination sizes must match");
             return kernel::detail::gather<T, A, U, V>(
                 src, index,
                 ::xsimd::detail::make_index_sequence<xsimd::batch<T, A>::size>());
@@ -101,8 +101,8 @@ namespace xsimd
         gather(U const* src, batch<V, A> const& index,
                kernel::requires_arch<generic>) noexcept
         {
-            static_assert(batch<T, A>::size <= batch<V, A>::size,
-                          "Not enough indexes for gathering!");
+            static_assert(batch<T, A>::size == batch<V, A>::size,
+                          "Index and destination sizes must match");
             const auto dst = kernel::detail::gather<U, A, U, V>(
                 src, index,
                 ::xsimd::detail::make_index_sequence<batch<U, A>::size>());
@@ -203,8 +203,8 @@ namespace xsimd
                 batch<V, A> const& index,
                 kernel::requires_arch<generic>) noexcept
         {
-            static_assert(batch<T, A>::size <= batch<V, A>::size,
-                          "Not enough indexes for scattering!");
+            static_assert(batch<T, A>::size == batch<V, A>::size,
+                          "Source and index sizes must match");
             kernel::detail::scatter<T, A, U, V>(
                 src, dst, index,
                 ::xsimd::detail::make_index_sequence<batch<T, A>::size>());
@@ -216,8 +216,8 @@ namespace xsimd
                 batch<V, A> const& index,
                 kernel::requires_arch<generic>) noexcept
         {
-            static_assert(batch<T, A>::size <= batch<V, A>::size,
-                          "Not enough indexes for scattering!");
+            static_assert(batch<T, A>::size == batch<V, A>::size,
+                          "Source and index sizes must match");
             const auto tmp = batch_cast<U>(src);
             kernel::detail::scatter<U, A, U, V>(
                 tmp, dst, index,
--- a/include/xsimd/arch/generic/xsimd_generic_memory.hpp
+++ b/include/xsimd/arch/generic/xsimd_generic_memory.hpp
@@ -58,6 +58,12 @@ namespace xsimd
         // gather
         namespace detail
         {
+            template <class T, class U, class B>
+            using sizes_match_t = typename std::enable_if<sizeof(T) == sizeof(U), B>::type;
+
+            template <class T, class U, class B>
+            using sizes_mismatch_t = typename std::enable_if<sizeof(T) != sizeof(U), B>::type;
+
             template <typename T, typename A, typename U, typename V, size_t I>
             inline batch<T, A> gather(U const* src, batch<V, A> const& index,
                                       ::xsimd::detail::index_sequence<I>) noexcept
@@ -66,29 +72,42 @@ namespace xsimd
                               ::xsimd::index<I>());
             }
 
-            template <typename T, typename A, typename U, typename V, size_t I,
-                      size_t... Is,
-                      typename = typename std::enable_if<sizeof...(Is) != 0>::type>
+            template <typename T, typename A, typename U, typename V, size_t I0, size_t I1, size_t... Is>
             inline batch<T, A>
             gather(U const* src, batch<V, A> const& index,
-                   ::xsimd::detail::index_sequence<I, Is...>) noexcept
+                   ::xsimd::detail::index_sequence<I0, I1, Is...>) noexcept
             {
                 const auto test = kernel::detail::gather<T, A, U, V>(
-                    src, index, ::xsimd::detail::index_sequence<Is...>());
-                return insert(test, static_cast<T>(src[index.get(I)]),
-                              ::xsimd::index<I>());
+                    src, index, ::xsimd::detail::index_sequence<I1, Is...>());
+                return insert(test, static_cast<T>(src[index.get(I0)]),
+                              ::xsimd::index<I0>());
             }
         } // namespace detail
 
         template <typename A, typename T, typename U, typename V>
-        inline batch<T, A> gather(U const* src, batch<V, A> const& index,
-                                  kernel::requires_arch<generic>) noexcept
+        inline typename detail::sizes_mismatch_t<T, U, batch<T, A>>
+        gather(U const* src, batch<V, A> const& index,
+               kernel::requires_arch<generic>) noexcept
         {
             static_assert(batch<T, A>::size <= batch<V, A>::size,
-                          "Need matching sizes for indexing!");
+                          "Not enough indexes for gathering!");
             return kernel::detail::gather<T, A, U, V>(
                 src, index,
                 ::xsimd::detail::make_index_sequence<xsimd::batch<T, A>::size>());
+        }
+
+        template <typename A, typename T, typename U, typename V>
+        inline typename detail::sizes_match_t<T, U, batch<T, A>>
+        gather(U const* src, batch<V, A> const& index,
+               kernel::requires_arch<generic>) noexcept
+        {
+            static_assert(batch<T, A>::size <= batch<V, A>::size,
+                          "Not enough indexes for gathering!");
+            const auto dst = kernel::detail::gather<U, A, U, V>(
+                src, index,
+                ::xsimd::detail::make_index_sequence<batch<U, A>::size>());
+
+            return batch_cast<T>(dst);
         }
 
         // insert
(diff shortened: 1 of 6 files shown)


include/xsimd/types/xsimd_api.hpp
```diff
@@ -1922,6 +1990,20 @@ namespace xsimd
...
+     * @ingroup batch_bool_reducers
+     *
+     * Return true if none of the boolean values in the batch is true,
+     * false otherwise.
+     * @param x the batch to reduce.
+     * @return a boolean scalar.
+     */
+    template <class T, class A>
+    inline bool none(batch_bool<T, A> const& x) noexcept
+    {
```
serge-sans-paille, line 2003: This one looks good to me.

include/xsimd/types/xsimd_api.hpp
```diff
@@ -1462,6 +1507,29 @@ namespace xsimd
...
+     * @ingroup batch_data_transfer
+     *
+     * Scatter elements from the batch \c src at addresses starting at \c dst
+     * and offset by each element in \c index. The scale is always 1.
+     * @param src Batch of elements to scatter.
+     * @param dst Destination address
+     * @param index Indexes in which to store the elements to.
+     */
+    template <typename T, typename A, typename U, typename V>
+    inline void scatter(batch<T, A> const& src, U* dst,
```
serge-sans-paille, line 1527: I think You should use ``T == U`` here. The scatter intrinscis you're using don't support another configuration anyway

include/xsimd/types/xsimd_api.hpp
```diff
@@ -1431,6 +1461,21 @@ namespace xsimd
...
+     * Computes the batch of nearest integer values to scalars in \c x (in
+     * floating point format), rounding halfway cases to even.
+     * @param x batch of floating point values.
+     * @return the batch of nearest even integer values.
+     */
+    template <class T, class A>
+    inline batch<as_integer<T>, A>
+    round_to_even(batch<T, A> const& x) noexcept
+    {
+        return kernel::round_to_even(x, A {});
```
serge-sans-paille, line 1476: That's a surprising function, I'm curious about its usage?

include/xsimd/types/xsimd_api.hpp
```diff
@@ -1375,6 +1390,21 @@ namespace xsimd
...
+     * @ingroup batch_arithmetic
+     *
+     * Computes the approximate reciprocal of the batch \c x.
+     * The maximum relative error for this approximation is
+     * less than 1.5*2^-12.
+     * @param x scalar or batch of scalars
+     * @return the reciprocal.
+     */
+    template <class A>
+    inline batch<float, A> reciprocal(batch<float, A> const& x) noexcept
```
serge-sans-paille, line 1403: no `double` version? Even if it means using the default algorithm, for genericity I think we need one.

include/xsimd/types/xsimd_api.hpp
```diff
@@ -758,6 +758,21 @@ namespace xsimd
...
+     *
+     * Create a new batch gathering elements starting at address \c src and
+     * offset by each element in \c index. The scale is always 1.
+     * @param src Starting address.
+     * @param index Indexes of the elements to gather.
+     * @return a batch containing the gathered elements.
+     */
+    template <typename T, typename A, typename U, typename V>
+    inline batch<T, A> gather(U const* src, const batch<V, A>& index) noexcept
+    {
```
serge-sans-paille, line 772: I think you should have `V == U` here, much like scatter.

include/xsimd/arch/xsimd_neon64.hpp
```diff
@@ -592,6 +592,18 @@ namespace xsimd
...
+        /**************
+         * reciprocal *
+         **************/
+
+        template <class A>
+        inline batch<double, A>
+        reciprocal(const batch<double, A>& x,
+                   kernel::requires_arch<neon64>) noexcept
+        {
+            return vrecpeq_f64(x);
```
serge-sans-paille, line 604: ok, so SSE only has float specialization, while neon has the double one too, this should be reflected in the `xsimd_api.h` file

include/xsimd/arch/generic/xsimd_generic_memory.hpp
```diff
@@ -120,6 +156,42 @@ namespace xsimd
...
+                    src, dst, index, ::xsimd::detail::index_sequence<Is...>());
+            }
+        } // namespace detail
+
+        template <typename A, typename T, typename U, typename V>
+        inline void scatter(batch<T, A> const& src, U* dst,
+                            batch<V, A> const& index,
+                            kernel::requires_arch<generic>) noexcept
+        {
+            static_assert(batch<T, A>::size <= batch<V, A>::size,
```
serge-sans-paille, line 188: Given that the `get` operator requires a store for each get, I'd rather perform a single aligned store to scratch memory, then perform the scatter sequentially. that being said, I'd also love *not* needing that generic implementation if we have a specialization for each instruction set.

include/xsimd/arch/generic/xsimd_generic_memory.hpp
```diff
@@ -55,6 +55,42 @@ namespace xsimd
...
+            template <typename T, typename A, typename U, typename V, size_t I,
+                      size_t... Is,
+                      typename = typename std::enable_if<sizeof...(Is) != 0>::type>
+            inline batch<T, A>
+            gather(U const* src, batch<V, A> const& index,
+                   ::xsimd::detail::index_sequence<I, Is...>) noexcept
+            {
+                const auto test = kernel::detail::gather<T, A, U, V>(
+                    src, index, ::xsimd::detail::index_sequence<Is...>());
+                return insert(test, static_cast<T>(src[index.get(I)]),
```
serge-sans-paille, line 78: interesting approach to use insert here. Unfortunately for some arch we don't have a specialized version of insert and I fear it would be slower than the naive *fill a temporary buffer then aligned load* approach. Probably needs benchmarking.

serge-sans-paille (commented):
Thanks a lot for this PR, overall very interesting stuff in there :-) I think you'd better split it, one function per PR, for easier review though.

include/xsimd/arch/generic/xsimd_generic_memory.hpp
amyspark: This Godbolt compares the codegen of the two: https://godbolt.org/z/GM7cYeahe

And this one uses the two alternatives in context (deinterleaving of packed pixels): https://godbolt.org/z/eKGhEqWYq

include/xsimd/arch/xsimd_neon64.hpp
amyspark: There's an AVX512F-provided intrinsic, but it requires AVX512VL as well: https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=mm_rcp1&ig_expand=5571

I've patched it by requiring SFINAE if T is not a floating point type, is that OK with you?

include/xsimd/types/xsimd_api.hpp
amyspark: I've patched it by requiring SFINAE if T is not a floating point type, is that OK with you?

include/xsimd/types/xsimd_api.hpp
amyspark: We (Krita) need the source and destination types to be separate, because we need to perform uint8/16<->float32 conversion when gathering/scattering integer depth pixels. These cases are obviously covered by the default, but when the types do match we can let the user take advantage of AVX512F intrinsics.

include/xsimd/types/xsimd_api.hpp
amyspark: I think I made a mistake with the naming there ^^

The purpose of this function is to round floating-point numbers to integer-typed integers. Currently, `to_int` truncates (cvttps), and `round` makes an horrifying number of compares, moves, and blends, and returns a batch of floating points. See https://godbolt.org/z/q4rc6dvW3.

serge-sans-paille:
@amyspark & @JohanMabille once all these extensions are merged, I suggest we do a `8.1.0` release

amyspark:
> @amyspark & @JohanMabille once all these extensions are merged, I suggest we do a `8.1.0` release

That would be great! It would make version selection very easy, given the extensions I've upstreamed.

amyspark:
I've added the tests now. Do let me know if you'd like the `scale` parameter to be added.

include/xsimd/arch/generic/xsimd_generic_memory.hpp
serge-sans-paille: proof by benchmark, perfect :-)

include/xsimd/types/xsimd_api.hpp
serge-sans-paille: Well, I question the API then. When I read *scatter this register of `uint32_t` to this pointer of `float`*, what kind of conversion happen? Is it a `reinterpret_cast` or a `static_cast`?
If it's a `reinterpret_cast`, then I'd rather have a scatter to `uint32_t*` followed, in client code, by the `reinterpret_cast`.
If it's a `static_cast` well, that's unsupported by the intrinsic and I'd rather have a `batch_cast` followed by a `scatter`.

Just my views on the problem, I'm curious to read your thoughts on this.

amyspark: > Well, I question the API then. When I read scatter this register of uint32_t to this pointer of float, what kind of conversion happen? Is it a reinterpret_cast or a static_cast?

In all cases, where `T` and `U` don't match, it's because we intend xsimd to apply a `static_cast`. (These semantics are taken straight out of Vc's gather, [here](https://vcdevel.github.io/Vc-1.4.1/classVc_1_1Vector.html#aa5788937fe94e616e4587fff8343de53). Same applies for [scatter](https://vcdevel.github.io/Vc-1.4.1/classVc_1_1Vector.html#a877b3f1854117429aca9b4a84561cdbb).)

>  If it's a static_cast well, that's unsupported by the intrinsic and I'd rather have a batch_cast followed by a scatter.

That's not possible (AFAIK). Suppose you need to gather 4 `uint8_t`s into a SSE `batch<float>` (e.g. to calculate a sRGB gaussian mask). `batch_cast` will catch you because of the mismatching vector sizes:

https://github.com/xtensor-stack/xsimd/blob/ce48f25f061dcb2254eb2e75777a792dd73ee9ba/include/xsimd/arch/generic/xsimd_generic_math.hpp#L68

I suppose I could add another instance for where the vector sizes do match?

amyspark: Here's a sampler of four alternatives: template v scratch memory with static_cast v batch_cast. https://godbolt.org/z/zvMdqfzjv

Can't check with Google Benchmark because `DoNotOptimize` seems to not be working correctly at the moment: https://godbolt.org/z/hPbrjfMs9

amyspark: Alternatively, here are sample runs of llvm-mca on the individual alternatives: https://godbolt.org/z/bvjhvcGG4

include/xsimd/arch/generic/xsimd_generic_memory.hpp
```diff
@@ -55,6 +55,42 @@ namespace xsimd
...
+            inline batch<T, A> gather(U const* src, batch<V, A> const& index,
+                                      ::xsimd::detail::index_sequence<I>) noexcept
+            {
+                return insert(batch<T, A> {}, static_cast<T>(src[index.get(I)]),
+                              ::xsimd::index<I>());
+            }
+
+            template <typename T, typename A, typename U, typename V, size_t I,
+                      size_t... Is,
+                      typename = typename std::enable_if<sizeof...(Is) != 0>::type>
```
serge-sans-paille, line 71: Instead of an `enable_if` here, you should just use `size_t I0, size_t I1, size_t... Is`

include/xsimd/arch/generic/xsimd_generic_memory.hpp
```diff
@@ -55,6 +55,42 @@ namespace xsimd
...
+                return insert(test, static_cast<T>(src[index.get(I)]),
+                              ::xsimd::index<I>());
+            }
+        } // namespace detail
+
+        template <typename A, typename T, typename U, typename V>
+        inline batch<T, A> gather(U const* src, batch<V, A> const& index,
+                                  kernel::requires_arch<generic>) noexcept
+        {
+            static_assert(batch<T, A>::size <= batch<V, A>::size,
```
serge-sans-paille, line 87: why `<=` and not `==` ?

include/xsimd/arch/generic/xsimd_generic_memory.hpp
```diff
@@ -120,6 +156,42 @@ namespace xsimd
...
+            inline void scatter(batch<T, A> const& src, U* dst,
+                                batch<V, A> const& index,
+                                ::xsimd::detail::index_sequence<I>) noexcept
+            {
+                dst[index.get(I)] = static_cast<U>(src.get(I));
+            }
+
+            template <typename T, typename A, typename U, typename V, size_t I,
+                      size_t... Is,
+                      typename = typename std::enable_if<sizeof...(Is) != 0>::type>
```
serge-sans-paille, line 172: same here

include/xsimd/types/xsimd_batch.hpp
```diff
@@ -537,6 +542,34 @@ namespace xsimd
         return load_unaligned(mem);
     }
 
+    /**
+     * Create a new batch gathering elements starting at address \c src and
+     * offset by each element in \c index. The scale is always 1.
+     * @param src Starting address.
```
serge-sans-paille, line 548: please add a note here on the fact that if `T != U`, then a `static_cast` is performed

serge-sans-paille (commented):
Thanks a lot for the strong arguments you provided. Do you plan to ptopose a neon implementation (if that's possible, I don't know if this instruction set supports scatter /gather)

include/xsimd/arch/generic/xsimd_generic_memory.hpp
amyspark: Vc makes it possible to use `int8_t` or `int16_t` indexes with a `float*` or `double*`, so I did not want to restrain user choice here. But do let me know if you want it restricted here.

amyspark:
> Thanks a lot for the strong arguments you provided. Do you plan to ptopose a neon implementation (if that's possible, I don't know if this instruction set supports scatter /gather)

Not directly, as gather/scatter is the purview of SVE, unfortunately. However, the templating of these two functions was made possible by your vectorization of `insert`, so that is why I first proposed #708 before moving on to this PR.

include/xsimd/arch/generic/xsimd_generic_memory.hpp
serge-sans-paille: Well, you're checking the `size` so the number of elements, I think they *must* match.

include/xsimd/types/xsimd_batch.hpp
```diff
@@ -537,6 +542,40 @@ namespace xsimd
...
+        static_assert(std::is_convertible<T, U>::value, "Can't convert from src to this batch's type!");
+        return kernel::gather<A, T>(src, index, A {});
+    }
+
+    /**
+     * Scatter elements from the batch \c src at addresses starting at \c dst
+     * and offset by each element in \c index.
+     * If \c T is not of the same size as \c U, a \c static_cast is performed
+     * at element scatter time.
+     * @param src Batch of elements to scatter.
```
serge-sans-paille, line 567: src is unused here

serge-sans-paille:
Thanks a lot for this great addition to xsimd!

PR #707, opened 2022-04-10
insert: allow intrinsics to be used for Windows

amyspark:
This PR adjusts the check that was added in #690 to:

- account for MSVC only (_WIN32 also implies MSYS compilers, such as GCC and Clang)
- block usage of the intrinsics only on MSVS 2015 and earlier, which lack the definition.


serge-sans-paille:
That's much better, thanks for your time!

PR #710, opened 2022-04-11
Add xsimd::reciprocal

amyspark:
:wave:

As requested in #706, this is the first function from my Krita work.

I've added tests to account for the documented precision in Intel Intrinsics Manual.


serge-sans-paille:
Great addition! Thanks a bunch o/

PR #712, opened 2022-04-11
Add xsimd::none

amyspark:
👋 

This PR is once again an offshoot of #706. This one adds `xsimd::none`, which completes the trifecta of `any/all/none` boolean checking functions.


serge-sans-paille:
Thanks! The trifecta is complete now :-)

PR #715, opened 2022-04-12
Bda improve cppstd in tests

serge-sans-paille:


serge-sans-paille:
cc @JohanMabille what do you think of that change? Moar CI looks good to me.

JohanMabille:
LGTM. Thanks @emmenlau !

PR #719, opened 2022-04-18
Azure Pipelines: fix targeting clang-cl

amyspark:
👋 

Currently, the clang-cl build in Azure Pipelines is a dud-- it still uses MSVC 2022 because CMake cannot find the conda-installed compiler. This MR changes the workflow definition to use ClangCL as included in MSVS and removes some unused steps.


serge-sans-paille:
I'd like @JohanMabille to confirm this one is ok. Simpler configuration looks good to me :-)

JohanMabille:
Thanks for the simplification!

Issue #722, opened 2022-04-29
assert vs. static_assert in initializer list constructor


jatinchowdhury18:
I noticed that the initializer list constructor for `xsimd::batch` throws an assertion if the initializer list size does not match the batch size. Since both the batch size and the list size are known at compile-time, would it be possible to make this a `static_assert`? I was recently migrating some code that had been using a different SIMD wrapper, and having the `static_assert` would have helped me to catch some mistakes a bit earlier in the process.

Btw, thanks to everyone who works to make XSIMD awesome! I've been using the library regularly for a few years now, and absolutely love it :).

serge-sans-paille:
xsimd is C++11, and in C++1, `std::initializer_list<T>::size`` is not constexpr, so no ``static_assert`` :-/

see https://en.cppreference.com/w/cpp/utility/initializer_list/size for reference

serge-sans-paille:
That being said, we could check the C++ version and use a ``static_assert`` if legal, I'll do that :-)

jatinchowdhury18:
Ah, that makes sense. For some reason I had it as `constexpr` since C++11 in my head. Having a compile time check when possible would be fantastic! Thanks again :).

PR #723, opened 2022-04-30
Harmonize std::initializer_list check

serge-sans-paille:
Only since C++14, see #722

### DIFF
--- a/include/xsimd/types/xsimd_batch.hpp
+++ b/include/xsimd/types/xsimd_batch.hpp
@@ -22,6 +22,17 @@
 namespace xsimd
 {
 
+    namespace details
+    {
+        template <class B>
+        void check_batch_init(B const&, std::initializer_list<typename B::value_type> data)
+        {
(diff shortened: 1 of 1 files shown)


include/xsimd/types/xsimd_batch.hpp
```diff
@@ -22,6 +22,17 @@
 namespace xsimd
 {
 
+    namespace details
+    {
+        template <class B>
+        void check_batch_init(B const&, std::initializer_list<typename B::value_type> data)
+        {
+            (void)data;
+            assert(data.size() == B::size && "consistent initialization");
```
JohanMabille, line 31: It should be possible to do a static assert on `B::size` since it's constexpr.

serge-sans-paille: It's not possible because `data.size()` is not constexpr in C++11. It is in C++14 but anyway data here is not in a `constexpr` context so no :-/

PR #724, opened 2022-05-02
Support slide_left and slide_right for SSE2, AVX and AVX2

serge-sans-paille:


serge-sans-paille:
As green as a clover field \o/ @JohanMabille that's ready for review, and as usual, Neon part is for you, looks like https://developer.arm.com/architectures/instruction-sets/intrinsics/#q=vextq_s8 is the way to go there

PR #725, opened 2022-05-02
Fix CI setup

serge-sans-paille:
Do not confuse $() and ${}


serge-sans-paille:
cc @emmenlau I think we messed up at some point here :-/

emmenlau:
Oh no!

emmenlau:
I'm very sorry that the fix did not work as expected. I can at least say that the old code was certainly broken, as it was printing the variables to the screen instead of evaluating them. But it seems to be unknown how to correctly evaluate variables in this CI? That's quite unlucky...

PR #727, opened 2022-05-03
Added complex to swizzle tests

JohanMabille:


JohanMabille:
@serge-sans-paille this is ready.

serge-sans-paille:
thanks for the quick fix!

PR #726, opened 2022-05-03
Fix build on macOS and Arm

amyspark:
Hi @serge-sans-paille @emmenlau,

Our Mac dev ran into the same issues as you wrt my changes to gather/scatter and insert on Arm. This MR is a cleanup of yours, which I've used to:

- add a macOS universal build workflow
- fix the tests on M1
- and fix the QEMU runs on Azure Pipelines.

Fixes #694
See #721
Closes #725

Issue #721, opened 2022-04-20
Releasing 8.1.0


serge-sans-paille:
Just a meta-bug to track regression / wish list etc before 8.1.0 release

serge-sans-paille:
cc @amyspark @emmenlau and @JohanMabille

emmenlau:
Thumbs up from my side, I (personally) do not have anything urgent in the pipeline. However I see there are a number of open issues regarding arm, Android and/or Apple M1 support. I can think that many of them are related in some form? Is that something that would be easy to address or nothing for the 8.1.0 release?

amyspark:
:+1:  from my side. Awaiting checks from our Mac and Android devs.

amyspark:
Heads up, I've got a report that AppleClang doesn't completely like the Arm optimizations I made in #708 and #710. I'm preparing the MR and the changes to CI to prevent it from happening again.

JohanMabille:
swizzle fixed, I think we can release when the `nearbyint_as_int` is fixed.

amyspark:
@JohanMabille I'm working on permutes/shuffles/interleaves. It'll take a while, so I'll leave it for the next release.

JohanMabille:
ok.

emmenlau:
Thanks a lot guys, @JohanMabille two thumbs up for the release from my side!

JohanMabille:
xsimd 8.1.0 is out!

PR #728, opened 2022-05-03
Fix nearbyint_as_int test on Arm

amyspark:
This PR fixes the semantics of nearbyint_as_int on Arm. I made a typo in the AVX512DQ version too.

You'll notice that I added a guard around the NEON64 version; I've seen online that the codegen of GCC simply doesn't return the expected results, but Clang's does.

This PR depends on #726, which is why you'll see extraneous commits. They'll be gone once that one is merged.

See #721


serge-sans-paille:
I think you now only need to cherry-pick a few commits from this one :-)

serge-sans-paille:
Thanks o/

PR #730, opened 2022-05-09
Slight optimization to gather/scatter + related fixes

amyspark:
:wave: 

This introduces small fixes and quality-of-life optimizations I made as part of my debugging.

- Made the tests buildable on MSYS Clang
- Arm's type size SFINAE have been factored out for better availability
- Changed the internals of gather/scatter to take a single index in the template (the current element).

The last change includes a small refactor to move the `get` implementation inside `generic_memory`, thus making it style-wise more compliant and mistake-proof.

### DIFF
--- a/include/xsimd/types/xsimd_utils.hpp
+++ b/include/xsimd/types/xsimd_utils.hpp
@@ -196,6 +196,34 @@ namespace xsimd
         return res;
     }
 
+    namespace kernel
+    {
+        namespace detail
+        {
+            /**************************************
+             * enabling / disabling metafunctions *
+             **************************************/
+
+            template <class T>
+            using enable_integral_t = typename std::enable_if<std::is_integral<T>::value, int>::type;
+
+            template <class T, size_t S>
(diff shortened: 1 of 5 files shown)


include/xsimd/types/xsimd_utils.hpp
```diff
@@ -196,6 +196,34 @@ namespace xsimd
...
     }
 
+    namespace kernel
+    {
+        namespace detail
+        {
+            /**************************************
+             * enabling / disabling metafunctions *
+             **************************************/
+
```
serge-sans-paille, line 206: we should probably define a `void_t` type here, and use it for all the following `enable_..._t` types

serge-sans-paille:
Small nits, otherwise LGTM

include/xsimd/types/xsimd_utils.hpp
amyspark: Haven't added it because I can't fathom yet how `void_t` works 😅

serge-sans-paille:
Thanks for the great flow of contributions!

PR #732, opened 2022-05-11
Added Serge Guelton to the license file

JohanMabille:


serge-sans-paille:
Bravely merging that one. It's a real pleasure to contribute to that project

PR #736, opened 2022-05-12
Adding some complex batch operations

jatinchowdhury18:
I had a few complex operations based on the `xsimd` implementations and thought they might make useful additions to the library. If not, that's totally fine as well!

The operations are:
- `xsimd::pow` for the case where only one of the arguments is complex
- `xsimd::polar` (basically the same as `std::polar`)
- `xsimd::mul_real` as a faster way of doing `(complex_batch_1 * complex_batch_2).real();`, and corresponding `xsimd::mul_imag`

### DIFF
--- a/include/xsimd/arch/generic/xsimd_generic_math.hpp
+++ b/include/xsimd/arch/generic/xsimd_generic_math.hpp
@@ -1953,6 +1953,36 @@ namespace xsimd
             return select(absa == ze, cplx_batch(ze), cplx_batch(r * cos(theta), r * sin(theta)));
         }
 
+        template <class A, class T>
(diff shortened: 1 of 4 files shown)


serge-sans-paille:
Hey, thanks for contributing! It would be great to split each operation in individual PRs so that we can review them individiually. Some early thoughts:
- `xsimd::pow`: yes!
- `xsimd::polar`: yes!
- `xsimd::mul_real`: probably better to have an overload of `xsimd::mul` that accepts a complex and a scalar argument.

jatinchowdhury18:
Sure thing! I'll close this PR for now.

PR #739, opened 2022-05-12
Add xsimd::polar()

jatinchowdhury18:
Intended to be the equivalent of `std::polar()`

### DIFF
--- a/include/xsimd/arch/generic/xsimd_generic_math.hpp
+++ b/include/xsimd/arch/generic/xsimd_generic_math.hpp
@@ -970,13 +970,6 @@ namespace xsimd
             return { r * sincosTheta.second, r * sincosTheta.first };
         }
 
-        template <class A, class T>
-        inline batch<std::complex<T>, A> polar(const batch<T, A>& theta, requires_arch<generic>) noexcept
-        {
-            auto sincosTheta = sincos(theta);
-            return { sincosTheta.second, sincosTheta.first };
-        }
-        
         // fdim
         template <class A, class T>
         inline batch<T, A> fdim(batch<T, A> const& self, batch<T, A> const& other, requires_arch<generic>) noexcept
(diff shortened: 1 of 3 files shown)


include/xsimd/arch/generic/xsimd_generic_math.hpp
```diff
@@ -962,6 +962,21 @@ namespace xsimd
...
+        // polar
+        template <class A, class T>
+        inline batch<std::complex<T>, A> polar(const batch<T, A>& r, const batch<T, A>& theta, requires_arch<generic>) noexcept
+        {
+            auto sincosTheta = sincos(theta);
+            return { r * sincosTheta.second, r * sincosTheta.first };
+        }
+
+        template <class A, class T>
+        inline batch<std::complex<T>, A> polar(const batch<T, A>& theta, requires_arch<generic>) noexcept
```
serge-sans-paille, line 974: shouldn't this be `r` instead of `theta`?

serge-sans-paille: And then the formula is different, of course. But that would match the C++ api, as https://en.cppreference.com/w/cpp/numeric/complex/polar

include/xsimd/types/xsimd_api.hpp
```diff
@@ -1334,6 +1334,33 @@ namespace xsimd
...
+    {
+        return kernel::polar<A>(r, theta, A {});
+    }
+
+    /**
+     * @ingroup batch_math
+     *
+     * Returns a complex batch with phase angle \c theta.
+     * @param theta The phase angle of the desired complex result.
+     * @return exp(i * \c theta).
```
serge-sans-paille, line 1356: This is not the standard API, we try to stick to `std::polar`, even if that misses an optimization

serge-sans-paille (changes requested):
Thanks for the contribution! I have one major concern considering the API, but that's trivial to fix.

jatinchowdhury18:
Ah, good call, I hadn't realized that my "phase-only" implementation was not part of the standard library. I've removed the phase-only implementation, and tweaked the default arguments of the other implementation to match the standard API.

serge-sans-paille:
Thanks for this patch o/

PR #740, opened 2022-05-12
Add real and imaginary complex multiplication

jatinchowdhury18:
I often run into cases where I need to multiply two complex batches, but only take the real or imaginary part, e.g.
```cpp
batch<std::complex<float>> a;
batch<std::complex<float>> b;
auto y = real(a * b);
```
In this case, doing the full multiply operation is somewhat wasteful, since half of the operation ends up getting thrown away.

This PR adds `xsimd::mul_real()` and `xsimd::mul_imag()` to compute this result without the wasted computations.

### DIFF
--- a/include/xsimd/arch/generic/xsimd_generic_math.hpp
+++ b/include/xsimd/arch/generic/xsimd_generic_math.hpp
@@ -1780,6 +1780,19 @@ namespace xsimd
                                  self, other);
         }
 
+        // mul_real/imag
+        template <class A, class T>
+        inline batch<T, A> mul_real(const batch<std::complex<T>, A>& a, const batch<std::complex<T>, A>& b, requires_arch<generic>) noexcept
+        {
+            return (a.real() * b.real()) - (a.imag() * b.imag());
+        }
+
+        template <class A, class T>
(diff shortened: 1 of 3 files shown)


jatinchowdhury18:
It looks like the CI failure here is coming from a failure to install the build toolchain for one of the Linux GCC versions? I'm not sure if that could have been caused by one of the changes in the PR here.

JohanMabille:
THe failure was due to a network error. Thanks for this feature!

serge-sans-paille:
Sorry for reviving this, but I'd like to obsolete this API:

According to https://godbolt.org/z/71v9bj36x a compiler is well able to optimize the construction `xsimd::real(x * y)` and we don't want to optimize each interaction between xsimd function at API level, this paves the way for API burst.

JohanMabille:
Indeed, if all recent compilers can optimize away useless comptations, I would be in favor of removing these APIs.

serge-sans-paille:
GCC, clang, Intel, MSVC: https://godbolt.org/z/ThExrqeMY

jatinchowdhury18:
Cool, this is really good to know! I'll make a separate PR reverting this change.

PR #741, opened 2022-05-13
Accelerate gathers

amyspark:
:wave: 

I finished the xsimd port of our app, and I discovered that Vc accelerates narrowing gathers of particular types. xsimd doesn't have these primitives, so I looked at the assembly and ported them here.

This PR also removes dead code that got introduced by mistake in #730 .

(Setting as draft because I'd like to check if I can add choices in SSE-land, as well as widening options.)

### DIFF
--- a/include/xsimd/arch/xsimd_avx2.hpp
+++ b/include/xsimd/arch/xsimd_avx2.hpp
@@ -318,7 +318,7 @@ namespace xsimd
                                   kernel::requires_arch<avx2>) noexcept
         {
             // scatter for this one is AVX512F+AVX512VL
-            return _mm256_i32gather_epi32(src, index, sizeof(T));
+            return _mm256_i32gather_epi32(reinterpret_cast<const int32_t*>(src), index, sizeof(T));
         }
 
         template <class T, class A, class U, detail::enable_sized_integral_t<T, 8> = 0, detail::enable_sized_integral_t<U, 8> = 0>
@@ -326,7 +326,8 @@ namespace xsimd
                                   kernel::requires_arch<avx2>) noexcept
         {
             // scatter for this one is AVX512F+AVX512VL
-            return _mm256_i64gather_epi64(src, index, sizeof(T));
+            // Note: GCC for some reason defines this with long long int
+            return _mm256_i64gather_epi64(reinterpret_cast<const long long int*>(src), index, sizeof(T));
         }
 
         template <class A, class U,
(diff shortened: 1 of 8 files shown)


include/xsimd/arch/xsimd_avx512f.hpp
```diff
@@ -740,40 +740,57 @@ namespace xsimd
...
-        template <class A, class T,
-                  typename std::enable_if<std::is_same<uint64_t, T>::value || std::is_same<int64_t, T>::value,
-                                          void>::type>
-        inline batch<T, A> gather(T const* src, batch<int64_t, A> const& index,
+        template <class T, class A, class U, detail::enable_sized_integral_t<T, 8> = 0, detail::enable_sized_integral_t<U, 8> = 0>
+        inline batch<T, A> gather(batch<T, A> const&, T const* src, batch<U, A> const& index,
                                   kernel::requires_arch<avx512f>) noexcept
         {
-            return _mm512_i64gather_epi64(index, src, sizeof(T));
+            return _mm512_i64gather_epi64(index, reinterpret_cast<const long long int*>(src), sizeof(T));
```
serge-sans-paille, line 754: This cast looks suspicious. Looks like you want a cast to `const void *` instead : https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm512_i64gather_epi64&expand=4295&ig_expand=4021

serge-sans-paille:
One minor nit and good to go!

include/xsimd/arch/xsimd_avx512f.hpp
amyspark: @serge-sans-paille yep, GCC and Clang disagree on this. See the current PR checks.

amyspark: Update: it did work :D

include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -313,44 +313,62 @@ namespace xsimd
...
+            return _mm256_i64gather_epi64(reinterpret_cast<const long long int*>(src), index, sizeof(T));
         }
 
-        template <class A>
-        inline batch<float, A> gather(float const* src,
-                                      batch<int32_t, A> const& index,
+        template <class A, class U,
+                  detail::enable_sized_integral_t<U, 4> = 0>
+        inline batch<float, A> gather(batch<float, A> const&, float const* src,
+                                      batch<U, A> const& index,
```
bernhardmgruber, line 335: Hi! This actually fixes a bug in the latest release. Before this change, this overload of `xsimd::kernel::gather` had only 1 template parameter. And since `xsimd::batch::gather` calls `kernel::gather<A, T>(src, index, A {});` (2 template arguments), this overload is always a mismatch and thus never selected. In my case, this resulted in `xsimd::kernel::gather` always falling back to the generic implementation on AVX2+FMA3.

I was about to send a bugfix, but then noticed that this issue is fixed on master with this PR. Thank you!

amyspark: Apologies there! I noticed that bug only when disassembling Krita a few months ago.

I think this could be sped up once a patch release is cut? @serge-sans-paille ?

bernhardmgruber: No need to apologies, you are doing great work! :)

PR #742, opened 2022-05-13
Do not run tests on unsupported architectures

serge-sans-paille:


serge-sans-paille:
Validation passes on all Fedora architectures, see https://src.fedoraproject.org/rpms/xsimd/pull-request/10

PR #744, opened 2022-05-17
Fix benchmark build on armv7a

amyspark:
👋 

This PR fixes building the benchmarks on armv7a native compilers.


JohanMabille:
Thanks!

PR #746, opened 2022-05-18
Fix usage of unincluded std::array in avx512bw

amyspark:
:wave:

This PR is to fix an oversight in #724, in recent enough Clang versions `#include <array>` isn't implied by `type_traits`.

I tested this with MSYS Clang 14.0.3 on the CLANG32 and CLANG64 ABIs.


serge-sans-paille:
Thanks!

PR #755, opened 2022-05-30
Made a minor docs syntax fix.

wermos:
These two things weren't showing up properly on the docs.


JohanMabille:
Thanks! No need to wait for the CI, this is documentation only.

PR #760, opened 2022-06-01
Made it so that clang also warns on unused results.

wermos:
The proper way to check if the compiler is Clang or not, is to [check if `__clang__` is defined](https://stackoverflow.com/a/2658480/12591388). I updated the preprocessor directives around `XSIMD_NO_DISCARD` to check for `__clang__` as well.

### DIFF
--- a/include/xsimd/xsimd.hpp
+++ b/include/xsimd/xsimd.hpp
@@ -12,7 +12,7 @@
 #ifndef XSIMD_HPP
 #define XSIMD_HPP
 
-#if defined(__GNUC__)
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
Thanks!

wermos:
You're right. I thought of this after sending in the PR. I figured that I could send it as another PR after this one is accepted. But I'll add this stuff to this PR itself.

Apart from the `[[nodiscard]]` attribute, it seems that MSVC doesn't really have any way to emit warnings on discarding the result of a function, so I think adding this feature will make programming on Windows a lot easier.

PR #761, opened 2022-06-01
Added support for `[[nodiscard]]`

wermos:
As discussed in #760 , I have added some preprocessor code to detect `[[nodiscard]]` support if the compiler has it, and use `[[nodiscard]]` in those cases.

### DIFF
--- a/include/xsimd/xsimd.hpp
+++ b/include/xsimd/xsimd.hpp
@@ -12,7 +12,15 @@
 #ifndef XSIMD_HPP
 #define XSIMD_HPP
 
-#if defined(__GNUC__) || defined(__clang__)
+#if defined(__has_cpp_attribute)
+// if this is true, then the compiler supports feature test macros
+#if __has_cpp_attribute(nodiscard) >= 201603L
+// if this is true, then the compiler supports [[nodiscard]] without a message
+#define XSIMD_NO_DISCARD [[nodiscard]]
+#endif 
+#elif __cplusplus >= 201703L // this means we are using C++17 or higher
+#define XSIMD_NO_DISCARD [[nodiscard]]
+#elif defined(__GNUC__) || defined(__clang__) // this means we are using GCC or Clang
 #define XSIMD_NO_DISCARD __attribute__((warn_unused_result))
 #else
 #define XSIMD_NO_DISCARD


marzer:
Just saw that there's a bunch of build failures due to the placement of the `[[nodiscard]]` form of the attribute on `static` and `inline` functions; they can be refactored to have the `XSIMD_NO_DISCARD` macro placed before the other specifiers and it will still work even with the GCC attribute form: [godbolt](https://godbolt.org/z/r1Y7rhEYv)

include/xsimd/xsimd.hpp
```diff
@@ -12,7 +12,15 @@
...
 
-#if defined(__GNUC__) || defined(__clang__)
+#if defined(__has_cpp_attribute)
+// if this is true, then the compiler supports feature test macros
+#if __has_cpp_attribute(nodiscard) >= 201603L
+// if this is true, then the compiler supports [[nodiscard]] without a message
+#define XSIMD_NO_DISCARD [[nodiscard]]
+#endif
+#elif __cplusplus >= 201703L // this means we are using C++17 or higher
+#define XSIMD_NO_DISCARD [[nodiscard]]
```
serge-sans-paille, line 22: Do we need to filter on the date? Why is just `#if __has_cpp_attribute(nodiscard` not enough?

serge-sans-paille:
> Just saw that there's a bunch of build failures due to the placement of the `[[nodiscard]]` form of the attribute on `static` and `inline` functions; they can be refactored to have the `XSIMD_NO_DISCARD` macro placed before the other specifiers and it will still work even with the GCC attribute form: [godbolt](https://godbolt.org/z/19Mo9fs7f)

Sure, go ahead!

wermos:
I'll fix that keyword placement issue now.

include/xsimd/xsimd.hpp
wermos: The reason is that these [feature test macros](https://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros) were actually added in C++20, whereas `[[nodiscard]]` has been in the language since C++17.

wermos:
The build seems to be failing on GCC 6 for some reason. I am not quite sure why. Any help or pointers is appreciated 😅

marzer:
```
 error: ‘XSIMD_NO_DISCARD’ does not name a type
```

First error starts with that and just cascades after that. Doesn't make much sense to me. Maybe prior to GCC 7 the attribute being at the very beginning of the specifiers/return types is an error?

wermos:
Ohhhhh, that was some great sleuthing! Thank you so much for the help!

I like that "cascade of fallbacks" approach that you suggested. I'll incorporate that into this PR. This way, the code is sure to compile, and in the worst case, there will simply be some case where we _could_ have used `[[nodiscard]]`, but didn't (similar to the way it was before). If that happens, it can easily be fixed with another PR.

That being said however, I think the new suggested method looks good, and I don't think there are any missing cases (i.e. system configurations where we could have warned about unused results, but didn't).

wermos:
I have also added some comments to help others understand what those preprocessor checks are doing. However, I am open to suggestions on wording changes, if anyone has any.

serge-sans-paille:
I like this change! Thanks o/

PR #762, opened 2022-06-03
Fixed a typo.

wermos:
Fixed a minor spelling error (the word only has [one 'n'](https://en.wikipedia.org/wiki/Software_versioning)).


CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ option(XSIMD_REFACTORING ON)
 
 set(XSIMD_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/include)
 
-# Versionning
+# Versioning
 # ===========
```
serge-sans-paille, line 19: You need to remove one `=` below too ;-)

wermos: Good point. I'll do that

PR #765, opened 2022-06-03
GHA-based jobs for fun and forks

### LINKED ISSUE #764: Offer: GHA based CI builds
Hey @serge-sans-paille @JohanMabille,

Our xsimd support is due to be released soon in Krita 5.1. When I was testing Arm things out, I found it very useful to port Arm versions of the Azure jobs to GHA; plus, I added MSYS options to account for our use of LLVM-MinGW in Windows.

These made it possible to check for build errors without having to PR the changes (to access Azure) and wait ages (for Appveyor's single-job builds).

Would you be interested in a PR to upstream these?

amyspark:
:wave:

This PR enables most of the Azure and Appveyor-based jobs to run in GitHub Actions. I can't port those that use MSVC 2015 and 2017, since they are not supported by GitHub.

Let me know if you want the Azure Arm versions to be removed, I've left them in place for the time being.

Fixes #764


serge-sans-paille:
This looks incredible. Getting rid of azure except the few configuration we cannot reproduce on github makes sense to me.

PR #767, opened 2022-06-05
Fixed a TODO.

wermos:
I fixed the TODO, which said that using [`std::add_const`](https://en.cppreference.com/w/cpp/types/add_cv) and [`std::add_lvalue_reference`](https://en.cppreference.com/w/cpp/types/add_reference) is the preferred of building the `requires_arch` predicate.

I also removed the comment about the possible clash with the C++20 [`requires` keyword](https://en.cppreference.com/w/cpp/keyword/requires), as it is not a concern: The keyword is `requires`, so it does not clash with `requires_arch`.

### DIFF
--- a/include/xsimd/types/xsimd_register.hpp
+++ b/include/xsimd/types/xsimd_register.hpp
@@ -73,10 +73,9 @@ namespace xsimd
 
     namespace kernel
     {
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
Clean & simple, thanks!

PR #776, opened 2022-06-23
Harmonize initialization between batch(scalar) and batch{scalar}

### LINKED ISSUE #775: single value batch initialisation failing for double on avx
Screenshot (133)

It initializes the first value with all other garbage. The broadcast is failing.
https://xsimd.readthedocs.io/en/latest/api/xsimd_batch.html#_CPPv4N5xsimd5batch5batchE1T

serge-sans-paille:
Fix #775

### DIFF
--- a/include/xsimd/types/xsimd_batch.hpp
+++ b/include/xsimd/types/xsimd_batch.hpp
@@ -22,17 +22,6 @@
 namespace xsimd
 {
 
-    namespace details
-    {
-        template <class B>
-        void check_batch_init(B const&, std::initializer_list<typename B::value_type> data)
-        {
-            (void)data;
-            assert(data.size() == B::size && "consistent initialization");
-        }
-
-    }
-
     /**
      * @brief batch of integer or floating point values.
      *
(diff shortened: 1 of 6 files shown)


serge-sans-paille:
@marzer / @wermos : I'll happily take your review on that one!

include/xsimd/types/xsimd_batch.hpp
```diff
@@ -451,14 +437,15 @@ namespace xsimd
...
      */
     template <class T, class A>
-    inline batch<T, A>::batch(std::initializer_list<T> data) noexcept
-        : batch(data.begin(), detail::make_index_sequence<size>())
+    template <class... Ts>
+    inline batch<T, A>::batch(T val0, T val1, Ts... vals) noexcept
+        : batch(kernel::set<A>(batch {}, A {}, val0, val1, static_cast<T>(vals)...))
     {
-        details::check_batch_init(*this, data);
+        static_assert(sizeof...(Ts) + 2 == size, "as many arguments as batch elements");
```
wermos, line 448: Maybe the static assert message should say something like `"The constructor requires as many arguments as batch elements."`

wermos (commented):
Everything else looks alright to me 👍

include/xsimd/types/xsimd_batch.hpp
marzer: Yup, this would be only note too. Looks great to me otherwise!

serge-sans-paille:
`static_assert` updated, thanks @marzer and @wermos for the feedback and the great discussion we had on improving the batch constructor o/

Issue #771, opened 2022-06-18
Cannot create a batch of xtl::xcomplex


lamyj:
Using current HEAD or latest tags (8.1.0 for xsimd and 0.7.4 for xtl), creating a `xsimd::batch<xtl::xcomplex<double>>` yields a compile-time error.

Minimal working example:
```c++
#define XSIMD_ENABLE_XTL_COMPLEX 1
#include <xsimd/xsimd.hpp>

int main()
{
    xsimd::batch<xtl::xcomplex<double>> foo;
    return 0;
}
```

The compile error is, using `g++ -I xtl/include -I xsimd/include -std=c++14 -c foo.cpp` is:
```
In file included from xsimd/include/xsimd/memory/../config/../types/../types/./././././xsimd_sse2_register.hpp:16:0,
                 from xsimd/include/xsimd/memory/../config/../types/../types/././././xsimd_sse3_register.hpp:15,
                 from xsimd/include/xsimd/memory/../config/../types/../types/./././xsimd_ssse3_register.hpp:15,
                 from xsimd/include/xsimd/memory/../config/../types/../types/././xsimd_sse4_1_register.hpp:15,
                 from xsimd/include/xsimd/memory/../config/../types/../types/./xsimd_sse4_2_register.hpp:15,
                 from xsimd/include/xsimd/memory/../config/../types/../types/xsimd_fma3_sse_register.hpp:15,
                 from xsimd/include/xsimd/memory/../config/../types/xsimd_all_registers.hpp:12,
                 from xsimd/include/xsimd/memory/../config/xsimd_arch.hpp:19,
                 from xsimd/include/xsimd/memory/xsimd_aligned_allocator.hpp:27,
                 from xsimd/include/xsimd/xsimd.hpp:41,
                 from foo.cpp:2:
xsimd/include/xsimd/memory/../config/../types/../types/././././././xsimd_register.hpp: In instantiation of ‘struct xsimd::types::simd_register<xtl::xcomplex<double>, xsimd::sse2>’:
xsimd/include/xsimd/types/xsimd_batch.hpp:46:11:   required from ‘class xsimd::batch<xtl::xcomplex<double> >’
foo.cpp:6:41:   required from here
xsimd/include/xsimd/memory/../config/../types/../types/././././././xsimd_register.hpp:31:13: error: static assertion failed: usage of simd_register with unsupported type
             static_assert(!Arch::supported() || has_simd_register<T, Arch>::value,
[...]

PR #778, opened 2022-06-23
Bump required C++ version when using xtl

lamyj:
This fixes the [build error on arm64](https://github.com/xtensor-stack/xsimd/issues/771#issuecomment-1163152964) mentioned in #771.


test/CMakeLists.txt
```diff
@@ -44,6 +44,31 @@ OPTION(CROSS_COMPILE_ARM "cross compile for ARM targets" OFF)
...
+                    message(FATAL_ERROR "Unsupported compiler -- xsimd requires C++11 support!")
+                else()
+                    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+                endif()
+            endif()
+        endif()
+    endif()
+
+    if (NOT CROSS_COMPILE_ARM)
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -fPIC")
```
JohanMabille, line 69: This should be part of the "not WIN32" branch, otherwise the build on Windows with clang-cl fails.

JohanMabille (changes requested):

test/CMakeLists.txt
lamyj: Should be fixed with f8ddeee

JohanMabille (approved):

PR #781, opened 2022-06-25
Add NEON64 hadd for single-precision floats

drewcassidy:
I noticed this was missing. On clang at least this still produces two calls to FADDP, but without an extract instruction present in the Neon32 implementation


serge-sans-paille:
Clean & simple, thanks!

PR #785, opened 2022-07-02
Test and fill holes in xsimd scalar api

### LINKED ISSUE #784: separate scalar mode tests from full vector tests
As discussed at https://github.com/xtensor-stack/xsimd/issues/756 and at https://github.com/serge-sans-paille/pythran/issues/1999 ,  xsimd is designed to work in scalar mode on all architectures, even those without SIMD vector acceleration. pythran requires xsimd even on "unsupported" arches in order to access scalar mode.

Currently xsimd tests address full functionality and therefore fail on "unsupported" (SIMD-free) architectures.

The tests are defined in XSIMD_TESTS in test/CMakeLists.txt.  It would be useful if that list of tests could be split into scalar-mode and vector-mode tests (e.g. create two lists, XSIMD_SCALAR_TESTS and XSIMD_VECTOR_TESTS), with some kind of new cmake option to switch off XSIMD_VECTOR_TESTS when running on SIMD-free architectures.

serge-sans-paille:
Fix #784

### DIFF
--- a/benchmark/xsimd_benchmark.hpp
+++ b/benchmark/xsimd_benchmark.hpp
@@ -49,8 +49,8 @@ namespace xsimd
         for (size_t i = 0; i < size; ++i)
         {
             op0[i] = T(0.5) + std::sqrt(T(i)) * T(9.) / T(size);
-            op1[i] = T(10.2) / T(i + 2) + T(0.25);
-            op2[i] = T(20.1) / T(i + 5) + T(0.65);
+            op1[i] = T(10.2) / T(i + 3) + T(0.25);
+            op2[i] = T(20.1) / T(i + 2) + T(0.65);
         }
     }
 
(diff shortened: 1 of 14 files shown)


serge-sans-paille:
The goal here is

1. make sure every vector function has it scalar counter part available
2. make sure the scalar API is compilable on every target, even those where we don't have batch support

serge-sans-paille:
cc @JohanMabille : while doing this I already found an implementation bug in the vector version of `bitwise_andnot` and a compiler (MSVC) bug on avx512f... that goes way behond just the scalar part. i'll `rebase -i` once I'm finished

serge-sans-paille:
@JohanMabille : ready for review!

JohanMabille:
Good catch for the bitwise_andnot bug and the compiler bug!

PR #787, opened 2022-07-06
Converted the `switch` statements to `XSIMD_IF_CONSTEXPR`

### LINKED ISSUE #779: Compile-time dispatch instead of switch-case
As I pointed out in one of my messages in #775, there are many cases like [this](https://github.com/xtensor-stack/xsimd/blob/c2728632bdfecf1dfc004e7ef9ee908003169948/include/xsimd/arch/xsimd_sse2.hpp#L52), where a switch-case is employed to make the decision of which intrinsic to call, even though this decision could be done at compile-time.

As part of our program to improve debug build performance, as @marzer suggested, and for improving performance overall, this should be changed to using some `std::enable_if` type thing.

wermos:
Fixed #779, as per the discussions.

### DIFF
--- a/include/xsimd/xsimd.hpp
+++ b/include/xsimd/xsimd.hpp
@@ -37,17 +37,17 @@
 
 #ifdef __cpp_if_constexpr
 // this means that the compiler supports the `if constexpr` construct
-#define XSIMD_IF if constexpr
+#define XSIMD_IF_CONSTEXPR if constexpr
 #endif
 
-#if !defined(XSIMD_IF) && __cplusplus >= 201703L
+#if !defined(XSIMD_IF_CONSTEXPR) && __cplusplus >= 201703L
 // this means that the previous test failed, but we are using C++17 or higher
-#define XSIMD_IF if constexpr
+#define XSIMD_IF_CONSTEXPR if constexpr
 #endif
 
-#if !defined(XSIMD_IF)
+#if !defined(XSIMD_IF_CONSTEXPR)
 // this means that all the previous checks failed, so we fallback to a normal `if`
-#define XSIMD_IF if
+#define XSIMD_IF_CONSTEXPR if
 #endif
 
 #include "config/xsimd_config.hpp"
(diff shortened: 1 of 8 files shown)


wermos:
Hmm, it seems like two of the Azure builds timed out.

JohanMabille:
Indeed, the page providing the Intel SDE was temporarily unavailable. I've restarted the job.
Thanks for all the work on this PR!

include/xsimd/xsimd.hpp
```diff
@@ -35,6 +35,21 @@
 #define XSIMD_NO_DISCARD
 #endif
 
+#ifdef __cpp_if_constexpr
+// this means that the compiler supports the `if constexpr` construct
+#define XSIMD_IF if constexpr
+#endif
+
+#if !defined(XSIMD_IF) && __cplusplus >= 201703L
```
serge-sans-paille, line 43: Should be and `#elif  __cplusplus >= 201703L`

serge-sans-paille:
I'd like to suggest replacing `XSIMD_IF(cond)` by `if XSIMD_CONSTEXPR (cond)` which is more verbose but more explicit too. Thoughts?

marzer:
> I'd like to suggest replacing XSIMD_IF(cond) by if XSIMD_CONSTEXPR (cond) which is more verbose but more explicit too. Thoughts?

That leaves some ambiguity - `constexpr` has been around since C++11, but `if constexpr` specifically since C++17. You would want to call it something more specific so it doesn't look like an alias for just `constexpr`, e.g.

```cpp
#if defined(__cpp_if_constexpr) || __cplusplus >= 201703L
	#define XSIMD_CONSTEXPR_IF constexpr
#else
	#define XSIMD_CONSTEXPR_IF
#endif

// ...

if XSIMD_CONSTEXPR_IF (sizeof(Foo) == 8)
{
	// ...
}
```

Which, while eliminating the ambiguity, is redundant since now it says "if" twice.

include/xsimd/xsimd.hpp
wermos: I was simply following the way we defined `XSIMD_NO_DISCARD`, via a bunch of cascading checks. But if you still want me to change it, I will do so.

JohanMabille:
Thanks!

PR #790, opened 2022-07-19
rsqrt: account for AMD vs. Intel inconsistency in RCPPS relative error

### LINKED ISSUE #757: Adjust the rsqrt error range to account for AMD hardware
@serge-sans-paille @JohanMabille,

I reported a possible miscodegen error in `test_rsqrt` to Microsoft [here](https://developercommunity.visualstudio.com/t/MSVC-2022-returns-imprecise-results-with/10042325). Turns out, AMD has slight precision issues wrt. Intel, so the test will fail with off-by-one errors for certain inputs.

Would it be possible to set an error range like it's done for the gamma function under AppleClang?

amyspark:
:wave:

This PR fixes #757 by setting up the `rsqrt` test with the documented relative error from Intel/AMD.

(For future reference: see https://news.ycombinator.com/item?id=23891810)

Fixes #757

### DIFF
--- a/test/test_batch_float.cpp
+++ b/test/test_batch_float.cpp
@@ -59,12 +59,14 @@ protected:
     {
         // rsqrt
         {
-            array_type expected;
+            array_type res, expected;
             std::transform(lhs.cbegin(), lhs.cend(), expected.begin(),
                            [](const value_type& l)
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
Thanks for this. The goal of this test is mostly to check AP consistency, it's a bit clumsy for inaccurate functions like `rsqrt`

PR #800, opened 2022-08-10
Support fixed size SVE

### LINKED ISSUE #786: [DISCUSSION] How to support Arm SVE
I'm investigating how to support Arm SVE/SVE2 [1] in xsimd.

SVE vector is size agnotics. The register size (lanes) is determinted at run time, and the according C/C++ type is size-less (incomplete type). E.g., SVE `svint8_t` maps to NEON `int8x16_t` but without size. `sizeof(svint8_t)` does not compile, and it's illegal to declare a member with type `svint8_t` inside a struct.

For this reason, looks xsimd core data structures `xsimd_register` and `batch`(derived from xsimd_register) do not support SVE, as `xsimd_register` struct contains a data member of the SIMD register type [2]. Code snippet pasted below.
```cpp
    struct simd_register<SCALAR_TYPE, ISA>                         \
    {                                                              \
        using register_type = VECTOR_TYPE;                         \
        register_type data;                                        \
        operator register_type() const noexcept { return data; }   \
    };                                                             \
```

I studied google highway which supports size agnostic vectors (Arm SVE, RISCV RVV). AFAIK, highway only handles the SIMD register type, without holding a register value [3]. The API often requires an explicit register type argument.
Below is a simplified highway example to add two arrays vertically.
```cpp
void vsum_hwy(const int* x, const int* y, size_t size, int* z) {
  // generate explicit vector_type from element type
  const ScalableTag<int> vector_type;
  // Lanes(d) generates runtime code for SVE/RVV, constexpr otherwise
  for (size_t i = 0; i < size; i += Lanes(d)) {
    // Load/Store requires explicit vector_type
    auto vx = Load(vector_type, x + i);
    auto vy = Load(vector_type, y + i);
    auto vz = Add(vx, vy);
    Store(vz, vector_type, z + i);
  }
}
```

I would like to hear ideas on how to support SVE size-less types. Highway approach may be inspiring, but I don't think xsimd should follow it.

A simpler but not ideal way is to only support fixed size SVE, e.g., SVE-128, SVE-256, ... SVE-2048. User must recompile the code to match the vector size on target machine.

[1] https://developer.arm.com/documentation/102476/0100
[2] https://github.com/xtensor-stack/xsimd/blob/8.1.0/include/xsimd/types/xsimd_register.hpp#L40
[3] https://github.com/google/highway/blob/master/g3doc/impl_details.md#vectors-vs-tags

cyb70289:
This PR adds fixed size SVE support to xsimd. Client must specify the
vector size in compiler options to enable SVE features.
E.g., `-march=armv8-a+sve -msve-vector-bits=128`
Make sure the vector size matches the hardware runs the code. Otherwise
the program behaviour is undefined, which may lead to crashes, hangs,
wrong results, even spurious successes.

The SVE ISA in hidden in `detail::sve`, not meant to be directly used by
client code. To make it explict that we only support fixed size SVE,
trivial aliases like `sve128`, `sve256` are exported for client to use.
They are only defined if client specifies according vector size at
compile time.

CI jobs are added to verify SVE128 and SVE256, which are available on
real hardware nowadays. Please note CI runs on x86 host and uses qemu
to emulate Arm and SVE instructions.

Closes #786


cyb70289:
SVE intrinsics reference FYI.
https://developer.arm.com/documentation/100987/0000/

SVE implementation is much concise than other ISA as most SVE intrinsics are overloaded per argument type.
E.g., `svadd` for all data types, without tedious `svadd_u8, svadd_s8, svadd_u16, etc.`.

cc @JohanMabille , @serge-sans-paille , @pitrou , @guyuqi, @edponce

edponce:
@cyb70289 Thanks for working on this. PR LGTM!
I suggest to also update the README file and possibly add a note that only SVE 128/256 are supported.

Conceptually, xsimd could validate the value of `__ARM_FEATURE_SVE_BITS` at compile-time. Although as it is currently, the aliases `sve128` and `sve256` will be undefined for unsupported cases.

JohanMabille:
Many thanks for the support of SVE!

Issue #805, opened 2022-08-15
Neq != for floats/doubles doesnt confirm to IEEE 754


kgpai:
The following snippet shows that x != Nan returns false , when its supposed to return true (https://en.wikipedia.org/wiki/NaN#Comparison_with_NaN) . 

```
float nan = std::numeric_limits<float>::signaling_NaN();
float num = 1.0f;
auto numX = d_type::broadcast(num);
 auto nanX = d_type::broadcast(nan);

 auto res = numX != nanX;     //Returns all false
 auto reseq = numX == nanX; //Returns all false. 
```

If this is not intentional , I can provide a PR with fix. Let me know . Thanks ! 
This is atleast true for AVX.

mbasmanova:
CC: @serge-sans-paille @JohanMabille @Yuhta

Folks, can you help look into this issue?

JohanMabille:
I think #806 fixes it.

mbasmanova:
Nice. Thank you @JohanMabille for reviewing and merging the fix.

PR #806, opened 2022-08-15
Ensure Neq confirms to IEEE 754.

kgpai:
Fixes issue #805, where neq returns false for  x != Nan , where x is any valid floating point value.


JohanMabille:
Thanks!

PR #808, opened 2022-08-24
Fix potential ABI issue in SVE support

serge-sans-paille:
When linking two compilation units compiled with different sve length (e.g. to
support dynamic dispatch) current implementation use the same arch type for
different width, which leads to various linkage / runtime error.

Fix that by using a size-dependent type.

### DIFF
--- a/include/xsimd/config/xsimd_arch.hpp
+++ b/include/xsimd/config/xsimd_arch.hpp
@@ -184,7 +184,8 @@ namespace xsimd
     {
     };
     using all_x86_architectures = arch_list<avx512bw, avx512dq, avx512cd, avx512f, fma3<avx2>, avx2, fma3<avx>, avx, fma4, fma3<sse4_2>, sse4_2, sse4_1, /*sse4a,*/ ssse3, sse3, sse2>;
-    using all_arm_architectures = arch_list<detail::sve, neon64, neon>;
+    using all_sve_architectures = arch_list<detail::sve<512>, detail::sve<256>, detail::sve<128>>;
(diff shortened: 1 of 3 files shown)


serge-sans-paille:
This is a WIP, cc @cyb70289

serge-sans-paille:
@JohanMabille / @cyb70289 : ready for review!

include/xsimd/config/xsimd_arch.hpp
```diff
@@ -184,7 +184,8 @@ namespace xsimd
     {
     };
     using all_x86_architectures = arch_list<avx512bw, avx512dq, avx512cd, avx512f, fma3<avx2>, avx2, fma3<avx>, avx, fma4, fma3<sse4_2>, sse4_2, sse4_1, /*sse4a,*/ ssse3, sse3, sse2>;
-    using all_arm_architectures = arch_list<detail::sve, neon64, neon>;
+    using all_sve_architectures = arch_list<detail::sve<512>, detail::sve<256>, detail::sve<128>>;
+    using all_arm_architectures = typename detail::join<arch_list<neon64, neon>, all_sve_architectures>::type;
```
cyb70289, line 188: Will it cause default arch to pick NEON, even if SVE related compiler options are enabled? If compiler supports SVE, it must also support NEON.

serge-sans-paille: Correct!

cyb70289 (approved):
LGTM. Thanks @serge-sans-paille !

PR #811, opened 2022-08-25
Disable fast exp10 on OSX

### LINKED ISSUE #810: Latest release (9.0) leads to compilation error on macOS (only)
Hi, thanks for the great package!

I am using xsimd as part of a python package. Since the last release (9.0), our macOS CI fails and we have been able to pin down the problem to xsimd. I am far from an expert with C++ so please, let me know if I can provide more details. I've been able to create a "minimal" reproducible example using cython. First, here's the error:

```
In file included from /Users/runner/miniconda3/envs/tabmat/include/xsimd/xsimd.hpp:55:
Users/runner/miniconda3/envs/tabmat/include/xsimd/arch/xsimd_scalar.hpp:447:16: error: no member named 'exp10f' in the global namespace; did you mean 'exp10'?
            return ::exp10f(x);
                   ^~
    /Users/runner/miniconda3/envs/tabmat/include/xsimd/arch/xsimd_scalar.hpp:445:18: note: 'exp10' declared here
        inline float exp10(const float& x) noexcept
                     ^
    /Users/runner/miniconda3/envs/tabmat/include/xsimd/arch/xsimd_scalar.hpp:451:16: error: no member named 'exp10' in the global namespace; did you mean simply 'exp10'?
            return ::exp10(x);
                   ^~
    /Users/runner/miniconda3/envs/tabmat/include/xsimd/arch/xsimd_scalar.hpp:449:19: note: 'exp10' declared here
        inline double exp10(const double& x) noexcept
                      ^
```

To reproduce, I have two files (main.cpp and cyt.pyx)

`main.cpp`:
```cpp
#include <xsimd/xsimd.hpp>
#include <iostream>

void test_fct(int a)
{
    std::cout << a << std::endl;
}
```

`cyt.pyx`:
```python
# distutils: language = c++

cdef extern from "main.cpp":
    void test_fct(int)
```

When compiling using `cythonize -i cyt.pyx`, it leads to the error above. Some observations:

- This doesn't happen if I use a previous version of xsimd
- This doesn't seem to happen on windows or Linux (our CI passed for both)
- I am using conda/mamba to install everything (compiling using clangxx 14.0.6 - also tested 11.1.0)

serge-sans-paille:
Fix #810

### DIFF
--- a/include/xsimd/arch/xsimd_scalar.hpp
+++ b/include/xsimd/arch/xsimd_scalar.hpp
@@ -441,7 +441,7 @@ namespace xsimd
         return !(x0 == x1);
     }
 
-#ifdef _GNU_SOURCE
+#if defined(_GNU_SOURCE) && !defined(__APPLE__)
     inline float exp10(const float& x) noexcept
     {
         return ::exp10f(x);


serge-sans-paille:
Note to self: use a share infrastructure for exp10 between scalar and vector mode.

serge-sans-paille:
@MarcAntoineSchmidtQC can you check that change? Thanks!

MarcAntoineSchmidtQC:
Works perfectly on my machine. Thanks.

JohanMabille:
The MacOS 10.15 has been deprecated on Azure (and it looks like they removed it sooner than expected). I will merge this PR after #816

PR #818, opened 2022-08-30
Assert on unaligned memory when calling aligned load/store

serge-sans-paille:
This should help understanding segfaults in debug builds.


JohanMabille:
Sweet

PR #822, opened 2022-09-11
Get rid of simd_return_type

serge-sans-paille:
This type is used to turn char into either uint8 or int8 depending on its sign. Unfortunately it is not used uniformly across the code base, so nothing prevents a user from instantiating a xsimd::batch<char>, and then calling xsimd::load_aligned from a char pointer, which may resulting in to different batch types (!).

Icing on the cake: It was misused (not forwarding the Arch parameter) in many case


serge-sans-paille:
cc @JohanMabille ready for review.

JohanMabille:
This type is not only used to convert char, but also handles the instantiation of `batch_bool` instead of `batch<bool>` when trying to load a buffer of booleans, and also automatic load into to `complex<T>` from `T` when the buffer is actually complex. This allows a homogeneous call to the `load_as` API (typically what we use in xtensor).

> nothing prevents a user from instantiating a xsimd::batch, and then calling xsimd::load_aligned from a char pointer, which may resulting in to different batch types

Indeed, and this is not the goal of this type. It is meant for simplifyig the load of differents buffers in mixed arithmetic. Removing this type would mean recoding it in each downstream library already using it.

> Icing on the cake: It was misused (not forwarding the Arch parameter) in many case

This is definitely a bug that should be fixed;

serge-sans-paille:
ok, see #823 and #824 for just the individual fixes then.

PR #823, opened 2022-09-13
Always forward arch parameter

serge-sans-paille:


serge-sans-paille:
@JohanMabille please don't merge, I think I have an idea of test that would trigger this issue :-)

serge-sans-paille:
@JohanMabille ready for review now :-)

JohanMabille (approved):

PR #824, opened 2022-09-13
Do not specialize the behavior of simd_return_type for char

serge-sans-paille:
I understand the issue with char being either signed or unsigned, but

        load_as<char>

not returning a

        batch<char>

is too surprising--and currently leads to compile error anyway.


JohanMabille:
I need to check why we needed this in xtensor.

JohanMabille:
The tests of xtensor pass without this specialzation, so let's merge this. Sorry for the delay!

PR #827, opened 2022-09-19
Disable fast exp10 on MINGW

MehdiChinoune:


serge-sans-paille:
Thanks! Would you feel comfortable setting up mingw CI for xsimd?

MehdiChinoune:
> Thanks! Would you feel comfortable setting up mingw CI for xsimd?

It is already set! https://github.com/xtensor-stack/xsimd/blob/master/.github/workflows/windows.yml#L55

serge-sans-paille:
Than why did it not trigger the error you're trying to fix?

MehdiChinoune:
The error was trigged when building scipy. maybe the `xsimd_scalar` requires some options to be used!

serge-sans-paille:
ok, then I'll first add a PR that triggers the issue, then I'll merge your PR, ok ?

MehdiChinoune:
I think scipy include some files that define `_GNU_SOURCE`, most probably `pyconfig.h`. I checked It and It defines `_GNU_SOURCE` to 1
https://github.com/python/cpython/blob/v3.10.7/pyconfig.h.in#L1545

serge-sans-paille:
Can you cherry-pick https://github.com/xtensor-stack/xsimd/pull/828 and apply your patch on top of it? I've added a test case that reproduces your issue.

MehdiChinoune:
> Can you cherry-pick #828 and apply your patch on top of it? I've added a test case that reproduces your issue.

MSVC build jobs will certainly fail.

test/CMakeLists.txt
```diff
@@ -200,6 +200,10 @@ set(XSIMD_TESTS
     test_utils.hpp
 )
 
+if(NOT MSVC)
```
serge-sans-paille, line 203: Wouldn't that be more "error-proof" to do that check in `xsimd_scalar.hpp`, as you did for `__minGW32__`?

MehdiChinoune: I don't think `_GNU_SOURCE` will ever (and shouldn't) be defined with NSVC, and if so then an issue should be fixed.

serge-sans-paille: ack

serge-sans-paille:
I would love having https://github.com/xtensor-stack/xsimd/pull/827/commits/3fa0f2f29112ab9290e4388a6865ea65d5049b9b in another PR because I don't think comfortable reviewing it, and to avoid mixing concerns, otherwise looks good to me

MehdiChinoune:
> I would love having [3fa0f2f](https://github.com/xtensor-stack/xsimd/commit/3fa0f2f29112ab9290e4388a6865ea65d5049b9b) in another PR

done #829

serge-sans-paille:
Thanks for your patience!

PR #828, opened 2022-09-19
Test compatibility of xsimd/xsimd.hpp headers with _GNU_SOURCE

serge-sans-paille:
Currently only xsimd::exp10 is being specialized under that macro. This is a companion patch for #827.


serge-sans-paille:
Cherry-picked in #827

PR #833, opened 2022-09-24
Move CI to doctest instead of gtest

serge-sans-paille:


JohanMabille (changes requested):
We should add doc test as a dependency to the environment-dev file, and search for it in the CMakeLists.txt instead of vendoring it.

JohanMabille (approved):

JohanMabille:
The issue on windows seems related to the infrastructure, let's merge and try to workaround it.

PR #836, opened 2022-09-28
Disabled wrong warning in GCC12

JohanMabille:


test/CMakeLists.txt
```diff
@@ -69,6 +69,12 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER_ID MATCHES "GNU"
         if (NOT CROSS_COMPILE_ARM)
             set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -fPIC")
         endif()
+
+        if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 12)
```
serge-sans-paille, line 73: greater or just equal ?

JohanMabille: indeed maybe equal should be enough? However, this does not seem enough to fix the CI :s

serge-sans-paille:
Still fails for two config :-/ The warning suppression doesn't seem to be passed down on these config

JohanMabille:
@serge-sans-paille I suggest that we remove the MINGW64 and UCRT64 builds for now since we cannot wrokaround GCC's bug. I hope they will fix it in the next GCC and that it will be available in msys2 soon.

PR #840, opened 2022-10-05
Implement non-AVX specific zipping

### LINKED ISSUE #825: zip per vector or zip per 128-bit lane?
For vector size > 128, xsimd `zip_lo` and `zip_hi` api zips per 128-bit lane, as described in [_mm256_unpackhi_epi8](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm256_unpackhi_epi8&ig_expand=7358).
I'm wondering if this behaviour is wanted in practice. I do see real world AVX2 code has to shuffle after unpack to get a natually zipped result.
SVE suports straightforward zipping, but to match current api, it has to do some preshuffling. It looks bad as the caller may have to shuffle it back afterwards.

IMHO it would be better to change the behaviour to be `zip per vector`, not "zip per 128-bit lane". Would like to hear comments from community.

serge-sans-paille:
ignore the AVX constraint of 128 bits lane. This comes at a cost for AVX/AVX2/AVX512, but it's much faster on SVE. And closer to the intuitive definition of zipping.

Fix #825


serge-sans-paille:
Currently a WIP, only implementing the zip_lo part, and I want to make sure it works as expected on all architectures

serge-sans-paille:
@cyb70289 : can you have a look at this PR? Thanks!

cyb70289 (approved):
LGTM. Thanks @serge-sans-paille !

amyspark:
@serge-sans-paille This PR is causing [bug 463219](https://bugs.kde.org/show_bug.cgi?id=463219) in Krita on Arch and other platforms which ship xsimd 10.

serge-sans-paille:
Thanks @amyspark for reporting this API break.
I still think that zip should do what the name implies, so the documentation should be fixed.
Concerning inner-lane zipping, it looks like there's a use-case, what could be a good name?

amyspark:
>  I still think that zip should do what the name implies, so the documentation should be fixed.

It would still be an API break, and one that's not easily detectable; we only got wind of it because the glitch was visual (channels being mixed across pixels).

>  Concerning inner-lane zipping, it looks like there's a use-case, what could be a good name?

`xsimd::zip_whole` (as opposed to `lo`/`hi`) would make its intent clear, while preserving compatibility with pre-v10 codebases.

PR #841, opened 2022-10-07
Allow generic to be used in dispatch arch list

### LINKED ISSUE #638: Enable dispatching against the xsimd::generic architecture
👋 

Vc has a "Scalar" architecture that allows the library to be used without targeting any particular ISA, hence doing all computations on individual values.

xsimd does have something similar (`xsimd::generic`) that is the fallback for all architecture types. Would it be possible to complete its definition, if even with a dummy register type (a single T value), so as to make xsimd usable on potentially unknown instruction sets?

Yuhta:
Currently when `xsimd::generic` is put in dispatch arch list as a catch-all fallback, it fails the version comparison because the `version()` function is missing.  Adding it allow us to use it in dispatch, as long as the implementation does not involve any batch operation.

Fix #638


serge-sans-paille:
Thanks for the patch. This seems consistent with the other behavior, and could pave the way for a scalar implementation of generic batch...

PR #842, opened 2022-10-07
Fix typo in xsimd_avx512cd_register

Yuhta:


serge-sans-paille:
Wow, this means we don't validate that as we should... let me investigate CI

serge-sans-paille:
I've applied your patch on top of #843 . This does solve some issue, but with proper testing, some new issues are unveiled :-) I think we can close this PR and keep fixing the other bug in #843

PR #843, opened 2022-10-08
Enable CI for Intel's KNM architecture

serge-sans-paille:
This enables avx512 f, er, cd, pf only, which should improve coverage of avx512 testing.


serge-sans-paille:
@Yutha : this appeared to be much more difficult than expected, but support for kinght's mill is now finished. Some grey areas remain, but I'm mostly satisfied with the result.

PR #848, opened 2022-10-20
SVE: Implement scatter/gather

cyb70289:


serge-sans-paille:
LGTM, thanks for taking care of the SVE part!

PR #849, opened 2022-10-20
SVE: Implement nearbyint and ldexp

cyb70289:


serge-sans-paille:
Thanks for both the extra instructions *and* the doc update. Looks good to me (99% trusting the validation suite here, TBH)

Issue #315, opened 2019-10-10
Header inclusion fails on AVX512


zhihaoy:
Compile the following program
```
#include <xsimd/xsimd.hpp>
```
with command
```
/usr/bin/c++  -DXTENSOR_USE_XSIMD -isystem /where-xsimd-lives/include -g   -Wall -Wconversion -Wsign-compare -Wno-sign-conversion -fdiagnostics-color -march=skylake-avx512 -msse2 -msse3 -mssse3 -msse4.1 -msse4.2 -mavx -mfma -mbmi2 -mavx2 -mavx512f -mavx512vl -mavx512cd -mavx512dq -mavx512bw -mno-sse4a -mno-xop -mno-fma4 -mno-avx512pf -mno-avx512er -mno-avx512ifma -mno-avx512vbmi -std=c++17 -o a.o -c a.cc
```
fails with
```
In file included from /usr/lib/gcc/x86_64-linux-gnu/8/include/immintrin.h:55,
                 from /usr/lib/gcc/x86_64-linux-gnu/8/include/x86intrin.h:48,
                 from /where-xsimd-lives/g/xsimd/config/xsimd_include.hpp:17,
                 from /where-xsimd-lives/g/xsimd/types/xsimd_types_include.hpp:12,
                 from /where-xsimd-lives/g/xsimd/types/xsimd_traits.hpp:15,
                 from /where-xsimd-lives/g/xsimd/xsimd.hpp:14,
                 from a.cc:1:
/where-xsimd-lives/g/xsimd/types/xsimd_avx512_int16.hpp: In instantiation of 'static xsimd::detail::avx512_int16_batch_kernel<T>::batch_type xsimd::detail::avx512_int16_batch_kernel<T>::select(const batch_bool_type&, const batch_type&, const batch_type&) [with T = short int; xsimd::detail::avx512_int16_batch_kernel<T>::batch_type = xsimd::batch<short int, 32>; xsimd::detail::avx512_int16_batch_kernel<T>::batch_bool_type = xsimd::batch_bool<short int, 32>]':
[...]

PR #852, opened 2022-10-24
Some avx512 bugfixes

### LINKED ISSUE #850: Silent bug?
Happen to see this. In the code below, looks template parameters `A` and `T` are reversed and SFINAE may silently skip this function and fallback to non-optimized implementation when calling `scatter<A>(...)`.
https://github.com/xtensor-stack/xsimd/blob/master/include/xsimd/arch/xsimd_avx512f.hpp#L1376

Is it possible to catch similar issues as compile time?

cyb70289:
Fixes #850

### DIFF
--- a/include/xsimd/arch/generic/xsimd_generic_memory.hpp
+++ b/include/xsimd/arch/generic/xsimd_generic_memory.hpp
@@ -34,7 +34,7 @@ namespace xsimd
         inline batch<T, A> extract_pair(batch<T, A> const& self, batch<T, A> const& other, std::size_t i, requires_arch<generic>) noexcept
         {
             constexpr std::size_t size = batch<T, A>::size;
-            assert(0 <= i && i < size && "index in bounds");
+            assert(i < size && "index in bounds");
 
             alignas(A::alignment()) T self_buffer[size];
             self.store_aligned(self_buffer);
--- a/include/xsimd/arch/xsimd_avx512bw.hpp
+++ b/include/xsimd/arch/xsimd_avx512bw.hpp
@@ -362,11 +362,11 @@ namespace xsimd
         {
             XSIMD_IF_CONSTEXPR(sizeof(T) == 1)
             {
-                return _mm512_mask_blend_epi8(cond, false_br, true_br);
+                return _mm512_mask_blend_epi8(cond, false_br.data, true_br.data);
             }
             else XSIMD_IF_CONSTEXPR(sizeof(T) == 2)
             {
-                return _mm512_mask_blend_epi16(cond, false_br, true_br);
+                return _mm512_mask_blend_epi16(cond, false_br.data, true_br.data);
             }
             else
             {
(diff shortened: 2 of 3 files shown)


include/xsimd/arch/generic/xsimd_generic_memory.hpp
```diff
@@ -34,7 +34,7 @@ namespace xsimd
         inline batch<T, A> extract_pair(batch<T, A> const& self, batch<T, A> const& other, std::size_t i, requires_arch<generic>) noexcept
         {
             constexpr std::size_t size = batch<T, A>::size;
-            assert(0 <= i && i < size && "index in bounds");
```
cyb70289, line 37: `0 <= i` leads to compiler warning as `i` is unsigned.

include/xsimd/arch/xsimd_avx512bw.hpp
```diff
@@ -362,11 +362,11 @@ namespace xsimd
         {
             XSIMD_IF_CONSTEXPR(sizeof(T) == 1)
             {
-                return _mm512_mask_blend_epi8(cond, false_br, true_br);
+                return _mm512_mask_blend_epi8(cond, false_br.data, true_br.data);
```
cyb70289, line 365: Error from gcc-11 from original code:
```bash
/home/cyb/xsimd/include/xsimd/types/../arch/./xsimd_avx512bw.hpp:365:24: error: cannot convert value to a vector
  365 |                 return _mm512_mask_blend_epi8(cond.data, false_br, true_br);
```

include/xsimd/arch/xsimd_avx512bw.hpp
```diff
@@ -410,14 +410,14 @@ namespace xsimd
             batch<T, A> xx;
             if (N & 1)
             {
-                alignas(32) uint64_t buffer[8];
+                alignas(64) uint64_t buffer[8];
```
cyb70289, line 413: Occasionally segfaults from `slide` tests.
`_mm512_{load,store}_epi{32,64}` requires **64** bytes alignment, otherwise a general-protection exception may be generated.
https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm512_load_epi&ig_expand=4241

include/xsimd/arch/xsimd_avx512f.hpp
```diff
@@ -1380,18 +1380,16 @@ namespace xsimd
 
         // scatter
         template <class A, class T,
-                  typename std::enable_if<std::is_same<uint32_t, T>::value || std::is_same<int32_t, T>::value,
-                                          void>::type>
+                  class = typename std::enable_if<std::is_same<uint32_t, T>::value || std::is_same<int32_t, T>::value, void>::type>
```
cyb70289, line 1383: Missing `class =` make the template always malformed and this function is never called.

include/xsimd/arch/xsimd_avx512f.hpp
```diff
@@ -1380,18 +1380,16 @@ namespace xsimd
...
                             batch<int32_t, A> const& index,
                             kernel::requires_arch<avx512f>) noexcept
         {
             _mm512_i32scatter_epi32(dst, index, src, sizeof(T));
         }
 
-        template <class T, class A,
-                  typename std::enable_if<std::is_same<uint64_t, T>::value || std::is_same<int64_t, T>::value,
-                                          void>::type>
+        template <class A, class T,
```
cyb70289, line 1391: Reversed `A` and `T` template arguments always invalidates the third argument, together the missing `class = `, two silent bugs prohibit this function from being called.

include/xsimd/arch/xsimd_avx512bw.hpp
serge-sans-paille: how strange.

include/xsimd/arch/xsimd_avx512bw.hpp
serge-sans-paille: We should be using `A::alignment()` here instead, right?

include/xsimd/arch/xsimd_avx512f.hpp
serge-sans-paille: good catch!

include/xsimd/arch/xsimd_avx512bw.hpp
cyb70289: Yes. Will update.

cyb70289: Done

include/xsimd/arch/xsimd_avx512bw.hpp
cyb70289: Looks same issue as https://github.com/xtensor-stack/xsimd/issues/315.
I met with similar failures from SVE intrinsics, implicit cast breaks for no reason.

serge-sans-paille:
Thanks o/

PR #855, opened 2022-10-25
Fix for android builds

jamierpond:
On android builds this section is causing build errors, just wondering if this is the best solution?


serge-sans-paille:
The patch looks good to me. This basically means we should add android validation on our CI... any chance you would have experience in that topic?

jamierpond:
@serge-sans-paille unfortunately not yet, but this is something that we are also going to start doing, so I will be able to feed that back in a week or two :)

serge-sans-paille:
ok, I'll merge that one and will be looking forward to your CI patch :-)

PR #859, opened 2022-11-08
Get SSE2 value from reg[3]

joaromera:
From https://www.amd.com/system/files/TechDocs/24594.pdf page 596 ,and https://www.felixcloutier.com/x86/cpuid#fig-3-6 I see that the SSE2 flag comes from a different registry than the other flags. Does EDX correspond to reg[3] ?


serge-sans-paille:
Confirmed looking at Numpy codebase: https://github.com/numpy/numpy/blob/2cc34c46943a92c8a5e97a1af0bffa3f2712b634/numpy/core/src/common/npy_cpu_features.c#L408

Thanks for spotting this!

PR #863, opened 2022-11-18
Provide and test comparison operators overloads that accept scalars

### LINKED ISSUE #229: operator< (and others) have no scalar overload
The operator `<` doesn't have a scalar overload so that one side can be a scalar number.
Is this intentional?

serge-sans-paille:
Instead of writing the actual scalar overloads, rely on the nice property of friend operators defined inline that allow for implicit conversion.

Fix #229


serge-sans-paille:
@JohanMabille could you investigate that AVX2 issue with msvc 2022?

PR #865, opened 2022-11-19
Make xsimd compatible with -fno-exceptions

serge-sans-paille:


serge-sans-paille:
@JohanMabille are you fine with that change?

PR #788, opened 2022-07-07
Fix find_package(xsimd) for xtl enabled xsimd

kou:
If we enable xtl by -DENABLE_XTL_COMPLEX=ON, the xsimd CMake target
depends on the xtl CMake target. (See installed
xsimdTargets.cmake. You can find 'INTERFACE_LINK_LIBRARIES "xtl"' line
in the file.)

But installed xsimdConfig.cmake doesn't include
find_dependency(xtl). If we don't run find_dependency(xtl) in
xsimdConfig.cmake, the xtl CMake target isn't defined when users run
find_package(xsimd). If the xtl CMake target isn't defined, "xtl" in
'INTERFACE_LINK_LIBRARIES "xtl"' is processed as library name not
CMake target. If "xtl" is processed as library name, '-lxtl' is added
to link command line. It causes link error because 'libxtl.so' doesn't
exist.

How to reproduce:

  1. Install xsimd with -DENABLE_XTL_COMPLEX=ON:

         cmake -S . -B build -DCMAKE_INSTALL_PREFIX=/tmp/local -DENABLE_XTL_COMPLEX=ON
         cmake --build build --target install

  2. Create a test project:

     xsimd_test.cc:

         #include <xsimd/xsimd.hpp>

         int main(void) {
           return 0;
         }

     CMakeLists.txt:

          cmake_minimum_required(VERSION 3.1)
          project(xsimd_test)

          # Enable <PackageName_ROOT CMake variable for find_package().
          cmake_policy(SET CMP0074 NEW)

          find_package(xsimd REQUIRED)
          add_executable(xsimd_test xsimd_test.cc)
          target_link_libraries(xsimd_test xsimd)

  3. Build the test project:

          $ cmake -S . -B build -Dxsimd_ROOT=/tmp/local
          -- The C compiler identification is GNU 11.3.0
          -- The CXX compiler identification is GNU 11.3.0
          -- Detecting C compiler ABI info
          -- Detecting C compiler ABI info - done
          -- Check for working C compiler: /bin/cc - skipped
          -- Detecting C compile features
          -- Detecting C compile features - done
          -- Detecting CXX compiler ABI info
          -- Detecting CXX compiler ABI info - done
          -- Check for working CXX compiler: /bin/c++ - skipped
          -- Detecting CXX compile features
          -- Detecting CXX compile features - done
          -- Configuring done
          -- Generating done
          -- Build files have been written to: /tmp/abc/build
          $ cmake --build build --verbose
[...]

PR #875, opened 2022-12-27
Fix find_package(xsimd) for xtl enabled xsimd, reloaded

### LINKED ISSUE #869: xsimd install with xtensor dependency results in a broken CMake target file
👋 

I've just been told at Krita that Debian (and Ubuntu by extension)'s packaging of xsimd does not work to build Krita. I see in the Debian source that they're using `-DENABLE_XTL_COMPLEX=ON` to enable the xtl dependency, but in the xsimd-targets.cmake file, I read

```
add_library(xsimd INTERFACE IMPORTED)

set_target_properties(xsimd PROPERTIES
  INTERFACE_COMPILE_FEATURES "cxx_std_14"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "xtl"
)
```

with absolutely no reference to xtl-targets.cmake or the relevant `find_package` line.

I'm not sure if this is a xsimd or Debian packaging issue, but reporting it FYI.

amyspark:
PR #788 fixes xtl lookup by adding the corresponding find_dependency call. However, there are still two issues remaining:

- If xtl is not found, the INTERFACE_LINK_LIBRARIES property will make CMake interpret "xtl" as a .lib, pass it as a -l flag, and cause the build to fail.

- If xtl is used, the required definition XSIMD_ENABLE_XTL_COMPLEX=1 is never passsed to downstream consumers.

This commit fixes both issues, ensuring that if xtl was required at compile time, the Config module with not load unless xtl is also found.

Fixes #869

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,8 +96,8 @@ OPTION(ENABLE_XTL_COMPLEX "enables support for xcomplex defined in xtl" OFF)
 OPTION(BUILD_TESTS "xsimd test suite" OFF)
 
 if(ENABLE_XTL_COMPLEX)
-    add_definitions(-DXSIMD_ENABLE_XTL_COMPLEX=1)
     find_package(xtl 0.7.0 REQUIRED)
+    target_compile_definitions(xsimd INTERFACE XSIMD_ENABLE_XTL_COMPLEX=1)
     target_link_libraries(xsimd INTERFACE xtl)
 endif()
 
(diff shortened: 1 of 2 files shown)


amyspark:
@JohanMabille It should be noted that this commit (along with #788) **has** to be cherry-picked to 8.x for Debian to stop breaking Krita builds.

JohanMabille:
Awesome! Thanks for this fix!

PR #881, opened 2023-01-08
MSVC doesn't have a way to force sse3 / sse3 / sse4.x arch

### LINKED ISSUE #878: When compiling with MSVC the SSE level is always set to 4.2
There seems to be no way to configure xsimd to compile with e.g. SSE3 on MSVC. It is always configured to use SSE 4.2 [in this line](https://github.com/xtensor-stack/xsimd/blob/75b043b8e031f1ada8053fe80d5ba635e2a75588/include/xsimd/config/xsimd_config.hpp#L310).
This means the code will not run on some processors. It would be great if there would be some way of configuring this.

serge-sans-paille:
Fix #878

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -18,7 +18,7 @@ jobs:
           - x86
           - x64
         sys:
-         - { set: SSE, flags: "" }
+          - { set: SSE, flags: "/arch:SSE2" }
          - { set: AVX, flags: "/arch:AVX" }
          - { set: AVX2, flags: "/arch:AVX2" }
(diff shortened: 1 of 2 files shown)


serge-sans-paille:
@cyb70289 : you're the author of the patch that' I'm partially reverting here, and there was no associated CI / tests, can you please check that this patch works for you?

cyb70289:
Sure, will do.

cyb70289 (approved):

serge-sans-paille:
Thanks @cyb70289

PR #882, opened 2023-01-09
Provide dummy unsigned implementation of xsimd::abs

serge-sans-paille:
This prevents the -Wabsolute-value warning


serge-sans-paille:
That's not the right appraoch, `using std` is much better.

PR #885, opened 2023-01-21
Provide avx512f implementation of FMA and variant

### LINKED ISSUE #884: AVX-512f FMA support?
Hi folks!

I've been looking at some AVX-512f code and it seems like in that case the `xsimd_generic_arithmetic.hpp` is used with an `x * y + z` rather than issuing an `__mm512_fmadd_pd` instruction. Presumably same goes for the other fused operations. Would it be possible get support for FMA with AVX-512F added?

Thanks so much in advance!

serge-sans-paille:
Fix #884

### DIFF
--- a/include/xsimd/arch/xsimd_avx512f.hpp
+++ b/include/xsimd/arch/xsimd_avx512f.hpp
@@ -768,6 +768,45 @@ namespace xsimd
             return _mm512_roundscale_pd(self, _MM_FROUND_TO_NEG_INF);
         }
 
+        // fnma
+        template <class A>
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
cc @JohanMabille this one was clean & simple, I wonder how we could have missed that :-/

nilsdeppe:
Thank you so much!!! 🎉 🎉

JohanMabille:
> cc @JohanMabille this one was clean & simple, I wonder how we could have missed that :-/

Indeed :s

PR #890, opened 2023-01-24
Feature/document type traits

serge-sans-paille:


serge-sans-paille:
@nilsdeppe I was not happy with `fundamental_type` because batches are not fundamental types per `std::is_fundamental` so I chose `xsimd::scalar_type` but I'm not 100% convinced. I'd gladly accept review of the wording for each trait!

nilsdeppe (approved):
This looks good! I'm also not 100% on `scalar_type` but I don't have anything better. I agree, `fundamental_type` is not good.

Would it be possible to add something for getting the associated bool/mask type too:
```cpp
template <class T>
struct mask_type
{
    using type = bool;
};

template <class T, class A>
struct mask_type<batch<T, A>>
{
    using type = typename batch<T, A>::batch_bool_type;
};
```

serge-sans-paille:
@nilsdeppe added! @JohanMabille any suggestion better than `scalar_type_t<T>` ?

JohanMabille:
I think `scalar_type` is appropriated here.

JohanMabille (approved):

serge-sans-paille:
The only problem I have with `scalar_type` is for batch of complex type, would you expect that to return the real type or the complex type?

PR #896, opened 2023-01-30
Expand the value of M_LN10

serge-sans-paille:
The macro is not provided my Visual studio by default. The easiest and most straight forward way to support this is to expand its value.

### DIFF
--- a/include/xsimd/arch/xsimd_scalar.hpp
+++ b/include/xsimd/arch/xsimd_scalar.hpp
@@ -462,11 +462,11 @@ namespace xsimd
 #else
     inline float exp10(const float& x) noexcept
     {
-        return std::exp((float)(M_LN10 * (x)));
+        return std::exp(0x1.26bb1cp+1f * x);
     }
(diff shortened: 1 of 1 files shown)


amyspark:
Hi! Just catching up with things here. Why wasn't `_USE_MATH_DEFINES` defined, at least with ifndef-define, and then undef at the end of the file?

serge-sans-paille:
Because if client code starts by include <cmath> (without that define) then <xsimd/xsimd.hpp>, we can do whatever we want, we won't get the appropriate definition.

amyspark:
> client code starts by include <cmath> (without that define)

The macro could be enforced with a suitable definition in the target export.

marzer:
> The macro could be enforced with a suitable definition in the target export.

This only helps people who importing xsimd using `<insert your pet build system here>`. xsimd is a header-only library, so there's going to be plenty of use-cases for whom that alone is insufficient. Expanding it inline is a perfectly fine solution.

PR #902, opened 2023-02-13
Disable alignment test on ARM with GCC

JohanMabille:


serge-sans-paille:
some alignment issues?

JohanMabille:
Yes looks like there is a bug on arm v7 (maybe with some compilers only, not sure yet).

PR #903, opened 2023-02-21
Change bitwise_cast API

serge-sans-paille:
Take a scalar as parameter instead of a batch.


JohanMabille:
Note for releasing: this is a backward incompatible change and will required a major release. If you're ok with that, I can merge, otherwise we can wait for the shuffle implementation and a minor release before merging.

serge-sans-paille:
I'm fine with the major release. I have a few other functions to push, so yeah, let's move forward and then make a major release

PR #905, opened 2023-03-04
Provide generic signed /unsigned type conversion

### LINKED ISSUE #735: Compile error with xsimd on on ARM64/Apple Silicon
Here is the traceback, unfortunately I cannot share a bigger reproducible due to proprietary code, but feel free to ask for more information.

The xtensor-dependent program compiles fine with xsimd off.

```
error: no viable conversion from returned value of type 'xsimd::batch_bool<float, xsimd::neon64>' to function return type 'simd_return_type<float>' (aka 'xsimd::batch<float>')
```

serge-sans-paille:
A reinterpret_cast is usually the way to go. For the range of value when it's valid in C, we achieve the same behavior for batches.

Fix #735

### DIFF
--- a/include/xsimd/arch/generic/xsimd_generic_details.hpp
+++ b/include/xsimd/arch/generic/xsimd_generic_details.hpp
@@ -135,6 +135,51 @@ namespace xsimd
             }
         }
 
+        // some generic fast_cast conversion
+        namespace detail
+        {
+            template <class A>
+            inline batch<uint8_t, A> fast_cast(batch<int8_t, A> const& self, batch<uint8_t, A> const&, requires_arch<generic>) noexcept
+            {
+                return bitwise_cast<uint8_t>(self);
+            }
+            template <class A>
+            inline batch<uint16_t, A> fast_cast(batch<int16_t, A> const& self, batch<uint16_t, A> const&, requires_arch<generic>) noexcept
+            {
+                return bitwise_cast<uint16_t>(self);
+            }
+            template <class A>
+            inline batch<uint32_t, A> fast_cast(batch<int32_t, A> const& self, batch<uint32_t, A> const&, requires_arch<generic>) noexcept
+            {
+                return bitwise_cast<uint32_t>(self);
+            }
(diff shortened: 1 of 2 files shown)


include/xsimd/arch/generic/xsimd_generic_details.hpp
```diff
@@ -135,6 +135,51 @@ namespace xsimd
...
+            {
+                return bitwise_cast<uint16_t>(self);
+            }
+            template <class A>
+            inline batch<uint32_t, A> fast_cast(batch<int32_t, A> const& self, batch<uint32_t, A> const&, requires_arch<generic>) noexcept
+            {
+                return bitwise_cast<uint32_t>(self);
+            }
+            template <class A>
+            inline batch<uint64_t, A> fast_cast(batch<uint64_t, A> const& self, batch<int64_t, A> const&, requires_arch<generic>) noexcept
```
JohanMabille, line 157: Looks like you inverted the arguments here.

include/xsimd/arch/generic/xsimd_generic_details.hpp
```diff
@@ -135,6 +135,51 @@ namespace xsimd
...
+            {
+                return bitwise_cast<int64_t>(self);
```
JohanMabille, line 159: This should bitwise cast to `uint64_t`

JohanMabille (changes requested):

include/xsimd/arch/generic/xsimd_generic_details.hpp
JohanMabille: (But then I wonder why the tests are passing in the build where we force not instruction set)

JohanMabille: This one is not fixed in your latest commit

serge-sans-paille:
It now is ;-)

JohanMabille (approved):

JohanMabille:
The fialure is unrelated to the changes.

PR #906, opened 2023-03-06
Introduce and document xsimd::best_arch

### LINKED ISSUE #374: Runtime detection of instruction set
Hi,

I am wondering if it is possible to perform a runtime detection of the instruction set to use?
Basically the approach would be:
```C++
template <typename vector_type>
struct processor
{
  // ...
};

template <typename vector_type>
processor<vector_type> *create_processor()
{
   return new processor<vector_type>();
}

auto *processor = xsimd::native_dispatch<create_processor>();
```

I'm not sure if this is already possible, after reading the examples I've concluded that it isn't.

To be clear, my goal is to produce one executable that will be able to take advantage of the best instruction set available, with the lowest overhead possible.

Many thanks, and I hope that was the right way to ask.

Regards,
Alexandre

serge-sans-paille:
Fix #374


JohanMabille:
The CI failure seems totally unrelated.

PR #908, opened 2023-03-09
On some platforms __ARM_NEON is enough to decide we have Neon support

serge-sans-paille:


serge-sans-paille:
@JohanMabille I'm testing that one, please don't merge before I give you a green light.

include/xsimd/config/xsimd_config.hpp
```diff
@@ -251,7 +251,7 @@
  *
```
ricotz, line 251: I think the problem it located above, so when "#ifdef __ARM_NEON" is not enough.

serge-sans-paille: oh, I missed that indeed

serge-sans-paille: @ricotz maybe better like this?

ricotz: I guess the problem is on our side and needs to be handled downstream.
So if there is no actual problem then please hold back those changes.

serge-sans-paille: ack, please ping me whenever you find an out.

ricotz: Thank for looking into it, but this is change is not reasonable for upstream.
It is fixed downstream by respecting specifics of Ubuntu's toolchain.

serge-sans-paille:
Looks like the original issue was on mozilla side, I do think this patch is not actually needed, see https://bugzilla.mozilla.org/show_bug.cgi?id=1822901

PR #909, opened 2023-03-21
Feature/incr decr

serge-sans-paille:


serge-sans-paille:
@JohanMabille ready for review :-)

PR #910, opened 2023-03-22
Support SVE runtime dispatcher in xsimd

guyuqi:
Currently, SVE is not enabled in xsimd runtime dispatcher.
This PR adds the support for SVE runtime detecting.

### DIFF
--- a/include/xsimd/config/xsimd_cpuid.hpp
+++ b/include/xsimd/config/xsimd_cpuid.hpp
@@ -52,6 +52,7 @@ namespace xsimd
             unsigned avx512bw : 1;
             unsigned neon : 1;
             unsigned neon64 : 1;
+            unsigned sve : 1;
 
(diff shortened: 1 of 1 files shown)


include/xsimd/config/xsimd_cpuid.hpp
```diff
@@ -75,6 +76,15 @@ namespace xsimd
                 neon64 = 0;
                 best = neon::version() * neon;
 
+#elif defined(__ARM_FEATURE_SVE) && defined(__ARM_FEATURE_SVE_BITS) && __ARM_FEATURE_SVE_BITS > 0
```
serge-sans-paille, line 79: this shoudla ctually use the `XSIMD_WITH_XXX` macro definition, but that's already how the remaining part of the code are coded, so let's delay that to another patch.

PR #912, opened 2023-03-27
Fixed AVX detection

### LINKED ISSUE #911: Issue detecting AVX architecture on MacOS
I've run into some strange behaviour with a little test program that I'm writing, where `xsimd::available_architectures()` correctly detects that the CPU supports AVX2 instructions but does not detect that the CPU also supports AVX instructions.

I believe the CPU does in fact support AVX since when I run `sysctl -a | grep machdep.cpu`, I see:
Screen Shot 2023-03-25 at 6 56 16 PM

Another library that I use regularly has a [similar implementation](https://github.com/juce-framework/JUCE/blob/69795dc8e589a9eb5df251b6dd994859bf7b3fab/modules/juce_core/native/juce_intel_SharedCode.h#L50) that appears to give the correct result. I'd be happy to test out a potential fix, but I'm not sure I'd feel comfortable touching that part of the code myself, since I'd probably break more things than I would fix.

JohanMabille:
Fixes #911 

cc @serge-sans-paille

### DIFF
--- a/include/xsimd/config/xsimd_cpuid.hpp
+++ b/include/xsimd/config/xsimd_cpuid.hpp
@@ -118,31 +118,38 @@ namespace xsimd
 #endif
                 };
 
-                int regs[4];
+                int regs1[4];
 
-                get_cpuid(regs, 0x1);
+                get_cpuid(regs1, 0x1);
 
-                sse2 = regs[3] >> 26 & 1;
+                sse2 = regs1[3] >> 26 & 1;
                 best = std::max(best, sse2::version() * sse2);
 
-                sse3 = regs[2] >> 0 & 1;
+                sse3 = regs1[2] >> 0 & 1;
                 best = std::max(best, sse3::version() * sse3);
 
-                ssse3 = regs[2] >> 9 & 1;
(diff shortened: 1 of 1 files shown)


include/xsimd/config/xsimd_cpuid.hpp
```diff
@@ -141,6 +141,12 @@ namespace xsimd
                 if (sse4_2)
                     best = std::max(best, fma3<xsimd::sse4_2>::version() * fma3_sse);
 
+                avx = regs[2] >> 28 & 1;
```
serge-sans-paille, line 144: I don't understand the fix, can you explain?

JohanMabille: regs is used before being overwritten with the call o `get_cpuid(regs, 0x80000001);`

serge-sans-paille: ok. Maybe worth having separate locale variable for each call to `get_cpuid` then, right?

JohanMabille: Indeed.

PR #915, opened 2023-04-06
Finish documentation cleanup

serge-sans-paille:
harmonize basic function style with other sections'


JohanMabille:
Not sure to get it, the current TOC with the different sections is pretty nice. The issue is that each function in the table does not link to the API documentaiton anymore. In the previous version, one could click on `abs` from this [page](https://xsimd.readthedocs.io/en/latest/api/math_index.html) and get redirected to [the API](https://xsimd.readthedocs.io/en/latest/api/basic_functions.html#_CPPv4I00EN5xsimd3absE5batchI1T1AERK5batchI1T1AE)

serge-sans-paille:
If I didn"t messed up, that's also the case now, but instead of manually tagging each function with a label, we use the breathe mechanism. The only real change is that we no longer split mathematical functions to subpages. Or did I miss something?

JohanMabille:
Oh so that's the split into many subpages that was breaking the breathe mechanism?

PR #918, opened 2023-04-09
Defer to (good) compiler the task of computing ln10

serge-sans-paille:
xsimd is C++11 and hexadecimal floating point are C++17


serge-sans-paille:
See https://godbolt.org/z/5sxq7Ye5s

a16bitsysop:
this works for alpine linux as well with gcc 12 https://gitlab.alpinelinux.org/a16bitsysop/aports/-/jobs/1007269 (includes rebuild of apache-arrow)

serge-sans-paille:
thanks @a16bitsysop for the review!

PR #920, opened 2023-04-11
Setup testing of documentation example

serge-sans-paille:
That way we won't regress on documentation, which is always a plus.


serge-sans-paille:
cc @JohanMabille

JohanMabille (approved):

JohanMabille:
Awesome!

PR #923, opened 2023-04-25
Introduce XSIMD_DEFAULT_ARCH as a way to customize the default archit…

serge-sans-paille:
…ecture

One of the use-case is to detect when the default architecture is used by #defining XSIMD_DEFAULT_ARCH xsimd::unsupported.


serge-sans-paille:
@JohanMabille note the subtle bugs this patch unveils :-)

JohanMabille:
This was an easter egg just for you ;)

Issue #375, opened 2020-07-30
feature request: shuffle


abique:
Hi,

I've looked through the documentation, but I did not find a way to perform shuffle operations.
Is that missing? If so is it something that could become part of the library?

Many thanks,
Alexandre

JohanMabille:
Hi,

Indeed `xsimd` does not provide any high level API for shuffling yet. That's a feature we would like to have in the library, but we are not sure of how to implement it because of the combinatory issue.

abique:
What do you mean by combinatory issue?

serge-sans-paille:
xsimd now supports `swizzle`, that shuffles a single vector according to a constant mask. See [the doc](https://xsimd.readthedocs.io/en/latest/api/data_transfer.html#_CPPv4I000_Dp2VtE7swizzleNSt9enable_ifINSt13is_arithmeticI1TE5valueE5batchI1T1AEE4typeERK5batchI1T1AE14batch_constantI5batchI2Vt1AEDp6ValuesE)

serge-sans-paille:
@abique I know we're 3 year late on this one, but could you have a look at #925 and check that you were thinking of something along those lines? Thanks!

PR #925, opened 2023-05-14
Provide xsimd::shuffle of floating point batches

serge-sans-paille:
Related to #375


serge-sans-paille:
@JohanMabille I'm pretty happy with current state, if you have some time to spare & have a look?

JohanMabille (approved):
We should open an issue for not forgetting NEON implementation once this one is merged.

PR #927, opened 2023-05-24
Provide a generic implementation of xsimd::swizzle with dynamic indices

serge-sans-paille:
Also provide some specialization for available architectures.


serge-sans-paille:
@JohanMabille I've done the heavy lifting, some specialization may be missing but this looks good enough to go.

JohanMabille (approved):

PR #928, opened 2023-05-26
Remove 'const' from batch_bool assignment operators

mnijhuis-tos:
Assignment operators like `operator|=`, `operator&=` and `operator^=` should not be const since the purpose of those operators is updating the object that defines the operator. With `const`, the following simple function does not compile:
```
void xsimd_or(xsimd::batch_bool<bool>& b1, xsimd::batch_bool<bool>& b2) {
    b1 |= b2; 
}
```


serge-sans-paille:
How shameful. Thanks for the patch... and it also means our test suite has hole in it.

PR #929, opened 2023-05-26
Support isfinite/isinf for std::complex<X>

mnijhuis-tos:
Support xsimd::isfinite and xsimd::isinf, analogous to the existing xsimd::isnan implementation.

Extend tests so they also compile and run these functions with complex values.


serge-sans-paille:
LGTM, thanks!

PR #934, opened 2023-06-01
Update basic_usage.rst

zhmin:
fix typo error


serge-sans-paille:
Straignt to the point :-)

PR #937, opened 2023-06-20
Setup extra inliner check based on clang-query

serge-sans-paille:


serge-sans-paille:
@JohanMabille good to go?

PR #943, opened 2023-09-18
test emscripten-32wasm in the browser

DerThorsten:
* compile tests for emscripten-wasm32
* tests in the browser ( via playwright)
* wasm code is compiled with `-msse2`


serge-sans-paille:
<3. Please ping me when this is ready for review.

test/test_wasm/test_wasm.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -e
+
+# this dir
+TEST_WASM_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+SRC_DIR=$TEST_WASM_DIR/../..
+
+
+git clone https://github.com/emscripten-core/emsdk
```
serge-sans-paille, line 9: this looks like an install step. Maybe split this script into `install_wasm_deps.sh` and `test_wasm.sh`? i assume there are no distribution package on ubuntu or mamba for emsdk.

.github/workflows/emscripten.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - uses: mamba-org/setup-micromamba@v1
+        with:
+            environment-name: xsimd
+            create-args: >-
+                microsoft::playwright
+                python
```
serge-sans-paille, line 19: out of curiosity, why `python` there?

serge-sans-paille:
@DerThorsten If you're fine with current setup, could you squash it so that we can merge it?

.github/workflows/emscripten.yml
DerThorsten: python is need for playright which is a python package allowing you to start a (headless) browser and interact with it.

DerThorsten:
> @DerThorsten If you're fine with current setup, could you squash it so that we can merge it?

Actually, I want the actual tests also to be run in wasm.
I will try to port the tests, once we have that its good to go

test/test_wasm/test_wasm.sh
DerThorsten: Yeah we can split that up.
We might want to use the emsdk from conda but for all other projects I just ended up cloning emsdk as it is not really designed to be packaged (is shipping things, has a strange directory structure, and so on)

JohanMabille:
I tihnk we can fix the `exp10` implementation the way we did it for [Windows](https://github.com/xtensor-stack/xsimd/blob/61d389c87f61ca5575c77369ec5068ea450f716e/include/xsimd/arch/xsimd_scalar.hpp#L525)

DerThorsten:
> I tihnk we can fix the `exp10` implementation the way we did it for [Windows](https://github.com/xtensor-stack/xsimd/blob/61d389c87f61ca5575c77369ec5068ea450f716e/include/xsimd/arch/xsimd_scalar.hpp#L525)

Atm I used  `Approx ` in the tests instead of going for the win fix.

test/test_wasm/test_wasm.sh
DerThorsten: actually let's leave it as one script, otherwise it will be a tiny bit more complicated to pass a custom emsdk installation (as I do for local development)

DerThorsten:
> > @DerThorsten If you're fine with current setup, could you squash it so that we can merge it?
> 
> Actually, I want the actual tests also to be run in wasm. I will try to port the tests, once we have that its good to go

I am happy with the setup now, It runs the normal test-suite in the (headless) browser!
I would prefer someone would press the "Squash and merge" button in github instead of me fiddling around with git (git is not my strong suit) 
The single commit squashed  commit message can just be 'added emscripten-wasm32 tests'

JohanMabille:
The issue with the squash and merge is that you don't have a merge commit reference the PR. I will squash it manually (also the merge commit in this PR makes hte branch in conflict with itself when rebasing :s)

PR #952, opened 2023-10-09
Initial Implementation for the new WASM based instruction set

anutosh491:


anutosh491:
Now if I understand correctly as WASM support for CI has been introduced, the doctests would be run in a wasm runtime. Hence we would need all essential function ( or rather functions that don't have a fallback implementation in our `xsimd_wasm.hpp` file) so that we are able to compile the code.

How do we plan to test the new instruction set out if that's the case. We could surely implement everything in this pr itself if we would like to. Not sure if we could maybe stop it here as we already have some operations implemented here and take care of other operations ( in a bunch) in subsequent prs!

include/xsimd/arch/xsimd_wasm.hpp
```diff
@@ -0,0 +1,595 @@
...
+                {
+                    assert(false && "unsupported arch/op combination");
+                    return {};
+                }
+            }
+            else
+            {
+                XSIMD_IF_CONSTEXPR(sizeof(T) == 1)
+                {
+                    return wasm_v128_and(wasm_i8x16_splat(0xFF >> other), wasm_u32x4_shr(self, other));
```
anutosh491, line 237: I see a direct intrinsic for this in the [wasm128_simd.h](https://clang.llvm.org/doxygen/wasm__simd128_8h.html#a2f0ba4f6733ea4fa526baccadf5f64e9) header file (not listed in the [documentation ](https://emscripten.org/docs/porting/simd.html)though so I used this workaround) 
```
static __inline__ v128_t __DEFAULT_FN_ATTRS wasm_u8x16_shr(v128_t __a,
                                                           uint32_t __b) {
  return (v128_t)((__u8x16)__a >> (__b & 0x7));
}
```
Though we might be able to use the intrinsic directly ?!

include/xsimd/arch/xsimd_wasm.hpp
```diff
@@ -330,6 +330,61 @@ namespace xsimd
...
+                {
+                    return wasm_u16x8_gt(self, other);
+                }
+                else XSIMD_IF_CONSTEXPR(sizeof(T) == 4)
+                {
+                    return wasm_u32x4_gt(self, other);
+                }
+                else
+                {
+                    return gt(self, other, generic {});
```
anutosh491, line 377: I don't see a well defined intrinsic for the batch/template specialization `xsimd::batch<uint64_t, xsimd::wasm>`, hence I have used this as of now.

include/xsimd/arch/xsimd_wasm.hpp
```diff
@@ -385,6 +385,18 @@ namespace xsimd
             return wasm_f64x2_gt(self, other);
         }
 
+        // isnan
+        template <class A>
+        inline batch_bool<float, A> isnan(batch<float, A> const& self, requires_arch<wasm>) noexcept
+        {
+            return wasm_v128_or(wasm_f32x4_ne(self, self), wasm_f32x4_ne(self, self));
```
anutosh491, line 392: This is I think so the first pure emulation that I've implemented, no direct intrinsic exists for any of the data types (float or double). I think the implementation should work
1) First I negate the equality between two batches ( the NAN elements would give 1 after `ne`, others would give 0)
2) Then we can use the bitwise_or operator to generates 1 (true) and 0 (false) elements.

Let me know if this is the correct way to handle this.

include/xsimd/arch/xsimd_wasm.hpp
```diff
@@ -446,6 +446,71 @@ namespace xsimd
...
+                {
+                    return wasm_u16x8_lt(self, other);
+                }
+                else XSIMD_IF_CONSTEXPR(sizeof(T) == 4)
+                {
+                    return wasm_u32x4_lt(self, other);
+                }
+                else XSIMD_IF_CONSTEXPR(sizeof(T) == 8)
+                {
+                    return lt(self, other, generic {});
```
anutosh491, line 498: I think we might have a different approach for this case through emulation once we can implement the shuffle operation.

include/xsimd/arch/xsimd_wasm.hpp
```diff
@@ -0,0 +1,1345 @@
...
+        {
+            v128_t one = wasm_f64x2_splat(1.0);
+            return wasm_f64x2_div(one, self);
+        }
+
+        // TODO: move this in xsimd_generic
+        namespace detail
+        {
+            template <class A, class T, class = typename std::enable_if<std::is_integral<T>::value, void>::type>
+            inline T hadd_default(batch<T, A> const& self, requires_arch<wasm>) noexcept
```
JohanMabille, line 912: I think this could be the fallback implementation of hadd (therefore it should be named hadd in the `xsimd_generic` file and accept the `generic` flag).

include/xsimd/arch/xsimd_wasm.hpp
```diff
@@ -0,0 +1,1345 @@
...
+            return wasm_f64x2_div(one, wasm_f64x2_sqrt(self));
+        }
+
+        // sadd
+
+        // TODO: move this in xsimd_generic
+        namespace detail
+        {
+            template <class A, class T, class = typename std::enable_if<std::is_integral<T>::value, void>::type>
+            inline batch<T, A> sadd_default(batch<T, A> const& self, batch<T, A> const& other, requires_arch<wasm>) noexcept
```
JohanMabille, line 986: Same remark as for hadd

include/xsimd/arch/xsimd_wasm.hpp
JohanMabille: LGTM

include/xsimd/arch/xsimd_wasm.hpp
```diff
@@ -0,0 +1,1345 @@
...
+
+        // bitwise_rshift
+        template <class A, class T, class = typename std::enable_if<std::is_integral<T>::value, void>::type>
+        inline batch<T, A> bitwise_rshift(batch<T, A> const& self, int32_t other, requires_arch<wasm>) noexcept
+        {
+            if (std::is_signed<T>::value)
+            {
+                XSIMD_IF_CONSTEXPR(sizeof(T) == 1)
+                {
+                    return wasm_i8x16_shr(self, other);
```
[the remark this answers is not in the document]
anutosh491: Yes, I have been referring to the `wasm_simd128.h` header file because I see that the documentation at `https://emscripten.org/docs/porting/simd.html` is not fully upto date. I even raised a couple issues to point out issues in the documenatation (https://github.com/emscripten-core/emscripten/issues/20397, https://github.com/emscripten-core/emscripten/issues/20482)

You can find the intrinsic here (https://github.com/llvm/llvm-project/blob/3651f377f632d9152f2bd8fc2f9302ec9f9bdd5e/clang/lib/Headers/wasm_simd128.h#L967)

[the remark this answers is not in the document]
anutosh491: Yes there is a corresponding left shift intrinsic too. I'll address it in the next commit.

include/xsimd/arch/xsimd_wasm.hpp
```diff
@@ -428,6 +428,102 @@ namespace xsimd
...
+                return wasm_i32x4_make(lut32[mask & 0xF], lut32[(mask >> 4) & 0xF], lut32[(mask >> 8) & 0xF], lut32[mask >> 12]);
+            }
+            else XSIMD_IF_CONSTEXPR(sizeof(T) == 2)
+            {
+                assert(!(mask & ~0xFF) && "inbound mask");
+                return wasm_i64x2_make(lut64[mask >> 4], lut64[mask & 0xF]);
+            }
+            else XSIMD_IF_CONSTEXPR(sizeof(T) == 4)
+            {
+                return from_mask(batch_bool<float, A> {}, mask, wasm {});
```
anutosh491, line 519: I think we would like to convert the `batch_bool<float, A>` value returned here to a `batch_bool<int32_t, A>` but I am not sure we have an intrinsic or any operation already for it !

JohanMabille:
Hey @anutosh491 what about formatting this so the CI can be green and then we can merge and iterate in the next PR? This would ease the review process.

anutosh491:
Yes that sounds like a plan. I've made a commit which should hopefully fix the style format error.

anutosh491:
TODO
1) Look after casting operations from one `batch`/`batch_bool` type to another.
2) Some operations like `reduce_max`, `reduce_min`, `shuffle`, `swizzle`, `fast_cast` are left to be implemented.

include/xsimd/arch/xsimd_wasm.hpp
```diff
@@ -0,0 +1,1345 @@
...
+            }
+            else XSIMD_IF_CONSTEXPR(sizeof(T) == 8)
+            {
+                return wasm_i64x2_eq(self, other);
+            }
+            else
+            {
+                assert(false && "unsupported arch/op combination");
+                return {};
+            }
```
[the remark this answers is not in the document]
anutosh491: Sure I'll try refactoring any code duplication or repetition in the subsequent PR.

JohanMabille:
Do not hesitate to open an issue to track them, and check the boxes as you implement them.

anutosh491:
Let me know if you like me to squash all commits into a couple/few meaningful commits with good commit messages :)

JohanMabille:
Indeed, that would nice. Thank you!

anutosh491:
I'll also get rid of the test file I had added just for reference to get started (`test/test_wasm/test_wasm.cpp`). Won't really need that !

SylvainCorlay:
Congrats @anutosh491, this is a great milestone.

PR #953, opened 2023-10-10
Added missing boolean operators in xsimd_api.hpp

anutosh491:
Support for `eq` and `neq` operations was added for `bool_batch` in `xsimd_api.hpp`


JohanMabille (changes requested):
Can you also add a test for these two functions in `https://github.com/xtensor-stack/xsimd/blob/master/test/test_batch_bool.cpp` ? `test_comparison` would be a good name.

JohanMabille (approved):

PR #960, opened 2023-10-25
Implemented sadd, ssub & reduce_add through the generic implementation for xsimd_sse2

anutosh491:
The generic implementations of `sadd`, `ssub` and `hadd` were added in #952 and can be found [here](https://github.com/xtensor-stack/xsimd/blob/f27bbb85be2a60cedb0585c2306ea8405f49af90/include/xsimd/arch/generic/xsimd_generic_arithmetic.hpp#L183) and hence these TODO's can be addressed now through the generic implementation.


anutosh491:
cc @JohanMabille

PR #962, opened 2023-10-25
Implemented few operations for the wasm instruction set

anutosh491:
The following operations would be implemented through the PR

- [x] reduce_max
- [x] reduce_min
- [x] swizzle
- [x] bitwise_cast
- [x] shuffle
- [x] fast_cast

PR #963, opened 2023-10-29
Provide a generic version for float to uint32_t conversion

serge-sans-paille:
This remove duplicate code for various Intel implementation and should also help for #962.

### DIFF
--- a/include/xsimd/arch/generic/xsimd_generic_details.hpp
+++ b/include/xsimd/arch/generic/xsimd_generic_details.hpp
@@ -180,6 +180,23 @@ namespace xsimd
             {
                 return bitwise_cast<int64_t>(self);
             }
+
+            // Provide a generic uint32_t -> float cast only if we have a
+            // non-generic int32_t -> float fast_cast
+            template <class A, class _ = decltype(fast_cast(std::declval<batch<int32_t, A> const&>(), std::declval<batch<float, A> const&>(), A {}))>
+            inline batch<float, A> fast_cast(batch<uint32_t, A> const& v, batch<float, A> const&, requires_arch<generic>) noexcept
+            {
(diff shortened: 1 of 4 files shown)


include/xsimd/arch/generic/xsimd_generic_details.hpp
```diff
@@ -180,6 +180,20 @@ namespace xsimd
...
             }
+            template <class A>
+            inline batch<float, A> fast_cast(batch<uint32_t, A> const& v, batch<float, A> const&, requires_arch<generic>) noexcept
+            {
+                // see https://stackoverflow.com/questions/34066228/how-to-perform-uint32-float-conversion-with-sse
+                batch<uint32_t, A> msk_lo(0xFFFF);
+                batch<float, A> cnst65536f(65536.0f);
+
+                auto v_lo = batch_cast<int32_t>(v & msk_lo); /* extract the 16 lowest significant bits of self                             */
+                auto v_hi = batch_cast<int32_t>(v >> 16); /* 16 most significant bits of v                                                 */
```
JohanMabille, line 191: I'm a bit confused here; `fast_cast` is supposed to be called by the generic `batch_cast` when a fast cast is available. Replacing the implementation specific to instruction sets with this generic one would end in calling the "slow" batch cast, right? Or am I missing sometihng?

anutosh491: Yes I was thinking along the same lines. Can we replace specific implementations with a generic one ?

serge-sans-paille: Indeed, I've adjusted the generic conversion to only be selected if we have a fast conversion from int32 to float.

JohanMabille (approved):

PR #964, opened 2023-10-31
Provide a generic version for uint32_t to float conversion, only if t…

serge-sans-paille:
…he int32 conversion is available

This remove duplicate code for various Intel implementation and should also help for #962.


serge-sans-paille:
@JohanMabille the error seems unrelated... a defect in the testing framework installation or something?

JohanMabille:
That's weird, it looks like it does not find doctest, but in that case I would have expected a failure when including the header. I've restarted the jobs in case it is just a flaky paranormal issue.

JohanMabille:
x-files-paranormal

JohanMabille (approved):

PR #969, opened 2023-11-08
Fix sincos implementation to cope with Emscripten

serge-sans-paille:


serge-sans-paille:
@JohanMabille / @anutosh491  just disabling the #ifdef seems to work, can you explain that?

JohanMabille:
Looks like the versions of Firefox and the SDKs that we use for testing have changed since we introduced this workaround, maybe the intrinsics have been fixed? Or maybe we are not testing what we think,meaning we don't build with WASM intrinsics, but rather with SSE2 that are then used to generate WASM? This latter seems highly improbable to me, but that's a possbility.

serge-sans-paille:
Anyway I double checked the sincos implementation compared to calling sin and cos, and it's *exactly* the same.

JohanMabille:
x-files-paranormal

JohanMabille:
(I have the feeling that I'm posting this GIF way too often, not a good sign)

serge-sans-paille:
@JohanMabille I think it's ok to merge this patch, but I ultimately leave it to you.

Issue #959, opened 2023-10-18
[Feature request] Add an option to skip installation


ToKiNoBug:
I tried to introduce xsimd into my project with `FetchContent`, it works well, but all xsimd headers are installed together with my project files, which is not expected. And I failed to find any way to control xsimd's installation.

So I hope to skip the installation of xsimd when it's introduced with `FetchContent`. This can be done with comparing `CMAKE_SOURCE_DIR` and `PROJECT_SOURCE_DIR`. If they are euqal, then xsimd is built as an individual project; otherwise it's built as a subproject of others.

If this feature is not implemented yet, I'm willing to do it myself and start a pr.

serge-sans-paille:
Thanks for using xsimd :-) PR welcome!

maxmarsc:
@ToKiNoBug I actually need to install xsimd alongside my project, because it's a library with xsimd needed in the headers. So I actually use the behaviour you wanna get rid of.

Do you think it could be possible to add an option to keep it ? Or should I rather use a cmake feature I'm not aware of ?
(sorry for the question but I'm recently struggling a lot with modern cmake setup)

ToKiNoBug:
> @ToKiNoBug I actually need to install xsimd alongside my project, because it's a library with xsimd needed in the headers. So I actually use the behaviour you wanna get rid of.
> 
> Do you think it could be possible to add an option to keep it ? Or should I rather use a cmake feature I'm not aware of ? (sorry for the question but I'm recently struggling a lot with modern cmake setup)

Ofcourse, it' really easy to implement. But I'm not sure what the default behavior should be like. When used as subproject, should we skip installation by default?

ToKiNoBug:
> Do you think it could be possible to add an option to keep it ? Or should I rather use a cmake feature I'm not aware of ? (sorry for the question but I'm recently struggling a lot with modern cmake setup)

I have dig into this problem but I found indeed there is no such cmake feature. The installation behavior of subprojects can not be controlled by parent project, unless the subproject provide an option to do it.

maxmarsc:
> > @ToKiNoBug I actually need to install xsimd alongside my project, because it's a library with xsimd needed in the headers. So I actually use the behaviour you wanna get rid of.
> > 
> > Do you think it could be possible to add an option to keep it ? Or should I rather use a cmake feature I'm not aware of ? (sorry for the question but I'm recently struggling a lot with modern cmake setup)
> 
[...]

PR #972, opened 2023-11-12
Add an option to skip installtion

ToKiNoBug:
An option `XSIMD_SKIP_INSTALL` is added to decide whether to install xsimd files. This can be useful when xsimd is used as an subproject.

To keep default behavior unchanged, the default value of this option is OFF. Developers can set `XSIMD_SKIP_INSTALL` to ON in order to skip xsimd installtion.

This pr implements #959

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,6 @@ docs/build/
 # VSCode / clangd IntelliSense
 .vscode/
 .cache/
+
+# CLion / IDEA
+.idea/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -44,3 +44,6 @@ docs/build/
 # VSCode / clangd IntelliSense
 .vscode/
 .cache/
+
+# CLion / IDEA
+.idea/
```
JohanMabille, line 49: This should be in a local gitignore rather than in that of the project. However, we already have other settings for vscode, let's clean this in a dedicated PR.

JohanMabille (approved):

PR #976, opened 2023-11-19
Fix grammar in basic_usage.rst

Drew-j-Smith:
Made the warning message clearer


serge-sans-paille:
Thanks, it's much more clear like this.

PR #978, opened 2023-11-22
Simplify expand implementation

avitase:
The previous implementation relied on a raw for-loop that compiles an index sequence that was subsequently fed into a combination of select and swizzle. The new version doesn't need the latter two steps but compiles the result directly in the for-loop.

### DIFF
--- a/include/xsimd/arch/generic/xsimd_generic_memory.hpp
+++ b/include/xsimd/arch/generic/xsimd_generic_memory.hpp
@@ -40,13 +40,16 @@ namespace xsimd
         {
             constexpr std::size_t size = batch_bool<T, A>::size;
             auto bitmask = mask.mask();
-            alignas(A::alignment()) as_unsigned_integer_t<T> swizzle_buffer[size];
+
+            alignas(A::alignment()) as_unsigned_integer_t<T> x_buffer[size];
+            x.store_aligned(&x_buffer[0]);
+
+            alignas(A::alignment()) as_unsigned_integer_t<T> expand_buffer[size];
             for (size_t i = 0, j = 0; i < size; ++i)
             {
-                swizzle_buffer[i] = (bitmask & (1u << i)) ? j++ : 0;
+                expand_buffer[i] = (bitmask & (1u << i)) ? x_buffer[j++] : 0;
             }
-            auto swizzle_mask = batch<as_unsigned_integer_t<T>, A>::load_aligned(&swizzle_buffer[0]);
-            return select(mask, swizzle(x, swizzle_mask), batch<T, A>(T(0)));
+            return batch<T, A>::load_aligned(expand_buffer);
         }
 
         // extract_pair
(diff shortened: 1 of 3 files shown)


serge-sans-paille:
On my setup, your code 4 times slower than mine, can you confirm on your side?

avitase:
Oo. What exactly did you run to test it?

serge-sans-paille:
there was a problem in my test. Let me redo the benchmark (!)

serge-sans-paille:
My version is now only three time faster. The benchmark:

```c++
$ cat a.cpp b.cpp
#include <xsimd/xsimd.hpp>

__attribute__((noinline)) xsimd::batch<uint32_t, xsimd::sse4_2> doit(xsimd::batch<uint32_t, xsimd::sse4_2> const& in, xsimd::batch_bool<uint32_t, xsimd::sse4_2> const& mask);

int main(int argc, char** argv)
{
  int n = atoi(argv[1]);
  auto in = xsimd::batch<uint32_t, xsimd::sse4_2>::load_unaligned((uint32_t*)argv[0]);
  auto mask = xsimd::batch_bool<uint32_t, xsimd::sse4_2>{true, false, false, true};
  for(int i = 0; i <n ; ++i)
    auto volatile drop = doit(in, mask);
  return 0;
}

#include <xsimd/xsimd.hpp>

__attribute__((noinline)) auto doit(xsimd::batch<uint32_t, xsimd::sse4_2> const& in, xsimd::batch_bool<uint32_t, xsimd::sse4_2> const& mask) {
  return xsimd::expand(in, mask);
}

$ clang++ -I../include -O2 -msse4.2 a.cpp -DNDEBUG -O3 b.cpp
$ time ./a.out 400000000
```

avitase:
That's weird. Here are my results with g++-11.4.0 for your version:

```
build$ git log | head -n1 && g++ -I../include -O2 -march=native a.cpp -DNDEBUG -O3 b.cpp && time ./a.out 400000000
commit bbe135454c0043afdfe4115744b8b5c6ff1641e1

real	0m3.009s
user	0m3.009s
sys	0m0.000s
```

and here for my version:

```
build$ git log | head -n1 && g++ -I../include -O2 -march=native a.cpp -DNDEBUG -O3 b.cpp && time ./a.out 400000000
commit 0f6c5837abe9d55aed39fe6ff22d7a257a273ccb

real	0m2.300s
user	0m2.300s
sys	0m0.000s
```

serge-sans-paille:
Got it! clang seems to optimize the code differently from gcc. I'll try to express the intent differently so that they both generate the same (fast) code

serge-sans-paille:
@avitase I just oushed a version that make gcc and clang generate close code, can you give it a shot (with both compilers !) And then maybe combine your approach with mine and see how it goes? Thanks!

avitase:
Your solution is now consistently faster for both GCC and Clang! Great work, thanks!

Issue #975, opened 2023-11-17
How to implement an expand function / scatter for masks?


avitase:
How can I implement an efficient "expand" function that behaves like the family of `*_mask_expand_epi*` instrinsics of the AVX-512 instruction set? As gather/scatter already got their own counterparts in xsimd's API, is it worth doing the same for compress/expand?

serge-sans-paille:
Mmmh, that's an interesting scenario. It's going to be quite difficult to implement this in an efficient generic manner, but I understand the pattern. Let's discuss the expected API. Following the AVX512 intrinsics, *à la* xsimd, that could be

```
xsimd::batch<T, A> xsimd::expand(xsimd::batch<T, A> self, xsimd::batch_bool<T, A> mask, xsimd::batch<T, A> filler);
```

The main problem probably is that we're likely to require shuffles to implement this operation, and we only have shuffle with constant masks, so we actually can provide a generic

```
xsimd::batch<T, A> xsimd::expand(xsimd::batch<T, A> self, xsimd::batch_bool_constant<T, A> mask, xsimd::batch<T, A> filler);
```

Unfortunately, a constant mask probably has very little use...

avitase:
@serge-sans-paille, thanks for your reply!

> ```
> xsimd::batch<T, A> xsimd::expand(xsimd::batch<T, A> self, xsimd::batch_bool<T, A> mask, xsimd::batch<T, A> filler);
> ```

I am not sure about the purpose of `filler`. Naively, I would expect that `mask` indicates what part of `self` should be distributed into an empty array. Do you propose to do this operation in `filler` instead? 
 
> The main problem probably is that we're likely to require shuffles to implement this operation

That's what I was wondering about as well. However,  for me it doesn't feel right since `expand` is simpler/less powerful than `shuffle` or `scatter` because it does not reorder or even duplicate and shouldn't be implemented in terms of these more costly/complex operations. However, if no hardware support is available, how bad would be a generic fallback with a raw for loop? The same holds for `compress`...

serge-sans-paille:
`*mask_expand_epi*`  has a filler argument that can be used to fill the empty parts of the array. I'm fine with having the fillear always be zero if it makes the generic function easier / more efficient to write.

After some extra thoughts:
expand with a static mask is just a specialization of a single shuffle. Not super fancy but why not. The generic gather relies on a sequence of generic inserts, not super fancy either :-/

[...]

PR #981, opened 2023-11-27
Generic, simple implementation fox xsimd::compress

serge-sans-paille:
Related to #975


serge-sans-paille:
@avitase : current generic implementation probably under-performs, I'll check that :-)

serge-sans-paille:
faster generic version! I'll add an AVX512 version at the end of the day.

serge-sans-paille:
@luhenry : out of curiosity, is there a specific riscv instruction that could help here? I'm fine with keeping the generic implementation for riscv (just like for arm and wasm) but I'm just curious.

luhenry:
@serge-sans-paille probably the vcompress.vm instruction. See https://github.com/riscv/riscv-v-spec/blob/fc76ec73fc1dd4531360ee8f3138f79f02e8b1b0/v-spec.adoc#165-vector-compress-instruction for details

PR #982, opened 2023-11-29
Specialize xsimd::compress for riscv

serge-sans-paille:


serge-sans-paille:
@luhenry This is an attempt based on the link you shared with me and my (very limited) understanding of riscv. Comments welcome!

luhenry:
@serge-sans-paille I take that there are tests covering it? The code looks very straightforward, so no specific feedback there.

serge-sans-paille:
> @serge-sans-paille I take that there are tests covering it? The code looks very straightforward, so no specific feedback there.

yeah there are tests. I was just unsure of the macro soup, but it seems to work, so I'll take this as a go :-)

PR #983, opened 2023-11-30
Remove unused variable in xsimd_generic_memory.hpp

padenot:
This break our build, we use `-Werror,-Wunused-variable` downstream


serge-sans-paille:
Obsoleted by #984 that adds the check in CI

PR #984, opened 2023-11-30
Enable -Werror checks on CI

serge-sans-paille:


serge-sans-paille:
cc @JohanMabille looks good now

PR #985, opened 2023-12-04
Feature/avx512xx

serge-sans-paille:
No actual instruction, but this paves the way for them.


serge-sans-paille:
Some context: the goal is to provide detection and definitions for some architectures we want to use in https://github.com/mozilla/gemmology/

JohanMabille:
> cc @JohanMabille Any opinion on this?

I have nothing against it ;) I think it would be worth opening issues to track the missing implementation for specific instruction sets once we merge this.

JohanMabille (approved):

PR #987, opened 2023-12-05
Fix various problems with architecture version handling

serge-sans-paille:
Some architectures were not registered in the `xsimd::all_architectures` arch list.

The test for issorted was incorrect.

Some architectures add wrong version values but this went unnoticed because of the invalid check.


JohanMabille:
dawson

JohanMabille:
I think we don't have fallback implementation for all operations, therefore we cannot add new instruction sets and registers if we don't implement a minimal subset.

What I don't understand is that when Anutosh was implementing the support for WASM, the build was failing until he had implemented the minimal subset; so WASM was selected at some point. Also this explains why removing the workaround Thorsten implemented for arg / conj did not make the tests fail: at this point, we were in this configuration, where WASM was never selected.

anutosh491:
Yes I believe there was a minimalistic subset of operations. After I had implemented all of those, the tests stopped failing.

SylvainCorlay:
Thanks @serge-sans-paille! It is looking like it is going to be in working order with wasm.

serge-sans-paille:
@SylvainCorlay yep, wasm fixed, I'm handling the final touches and this should be mergeable. @JohanMabille this deserves a minor release (patch + new features)

PR #989, opened 2023-12-08
Fix/avx512vnni

serge-sans-paille:


serge-sans-paille:
@JohanMabille this one should be in the patch release too.

PR #993, opened 2023-12-20
xsimd_scalar.hpp: fix for missing exp10 on macOS < 10.9

barracuda156:
`exp10` and `exp10f` are missing on macOS before 10.9, which breaks building some dependents that use `xsimd` headers. Fix that.


serge-sans-paille:
Thanks for the patch o/

barracuda156:
@serge-sans-paille Thank you for merging!

PR #994, opened 2023-12-22
Fix dispatching mechanism

serge-sans-paille:
Traverse required arch in the order provided by the user instead of trying to guess the best one.

It is actually impossible to define the notion of a best architectures as intel instruction set have a tree structure and not a linear structure : there are multiple leaves and none of them can be considered the "best".


serge-sans-paille:
Tested under firefox through gemmology, works like a charm.

serge-sans-paille:
@JohanMabille sorry to bother you, but I think this patch + the macosx exp10 patch are enough to justify a patch release

PR #1000, opened 2024-01-27
Add support form emulated<N> arch

### LINKED ISSUE #998: scalar implementation without code duplication
I'm currently a bit confused about the purpose of `xsimd::generic`. My expectation was that its possible to use this architecture as a fallback or for development purposes.  For this I would assume that `xsimd::batch<T, xsimd::generic>` just contains a single scalar value. But this is not the case - Instead the build fails because `xsimd::has_simd_register` is `false`. 

I could specialize `load_*` and `broadcast` calls and try to separate it from the rest of the implementation which then hopefully works with scalar and simd types. But it feels like this could be abstracted through implementation of the generic architecture. 

Is there a reason why this is not done? Are there examples how to write simd and scalar implementations without duplicating code?

serge-sans-paille:
Fix #998

### DIFF
--- a/.github/workflows/emulated.yml
+++ b/.github/workflows/emulated.yml
@@ -1,0 +1,74 @@
+name: Linux emulated build
+on: [push, pull_request]
+concurrency:
+  group: ${{ github.workflow }}-${{ github.job }}-${{ github.ref }}
+  cancel-in-progress: true
+defaults:
+  run:
+    shell: bash -l {0}
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    name: '${{ matrix.sys.compiler }} ${{ matrix.sys.version }} - emulated'
+    strategy:
+      matrix:
+        sys:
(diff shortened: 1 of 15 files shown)


serge-sans-paille:
@JohanMabille getting close to an end with this one, but I still need to double check if it works with Pythran as it touches xsimd_scalar.

serge-sans-paille:
pythran check ok, @JohanMabille I let you decide

test/test_api.cpp
```diff
@@ -15,6 +15,7 @@
 #include <random>
 
 #include "test_utils.hpp"
+#if 0
```
JohanMabille, line 18: Are the tests broken? Or is this a leftover?

serge-sans-paille: a left over

serge-sans-paille:
@JohanMabille all green even with the tests un-commented \o/

PR #1014, opened 2024-04-15
Force all functions to be always inline (when possible)

### LINKED ISSUE #644: Implement an inline macro to have maximum performance on MSVC
:wave:

While debugging the still bad codegen of MSVC, I realised that on reasonably big functions, it doesn't inline any of the xsimd functions even when marked as `inline`. The real workaround is to mark all functions with `__forceinline`.

Would it be possible to implement such a macro?

serge-sans-paille:
Fix #644


serge-sans-paille:
@marzer : a bit late to the party, but that should fix the inlining issues you once had with msvc

PR #1026, opened 2024-05-26
Prefer arch-independed paths for install

matwey:
xsimd is header only library, so prefer arch-independed paths for cmake and pkg-config.

References:
 * https://github.com/xtensor-stack/xtl/commit/d877d94836aff4d0f727acf3eaab8f4880ecb625
 * https://github.com/xtensor-stack/xtensor/commit/7738389861044c9618c7d59fb7602f7dddc1df7f


serge-sans-paille:
LGTM, why `DATAROOTDIR` and not `DATADIR` ?

matwey:
Sorry, I don't understand. I don't use `DATAROOTDIR` which is synonym for `DATADIR` according to the docs.

serge-sans-paille:
Indeed :bow:

PR #1032, opened 2024-06-14
Fix xsimd::available_architectures().has() for sve and rvv archs

dimula73:
Ideally the patch CPU detection code should also check if the length of SVE and RVV is actually supported by the current CPU implementation (i.e. ZCR_Elx.LEN register for SVE and something else for RVV), but I don't have such CPUs/emulators handy, so I cannot add such checks.

Given that xsimd::available_architectures().has() is a new feature of XSIMD13 and the length check has never been present in XSIMD, this bug is not a regression at least.

The patch also adds a unittest that reproduces the error the patch fixes


serge-sans-paille:
thanks :bow:

PR #1052, opened 2024-10-05
Add support for xsimd::tranpose

### LINKED ISSUE #107: Implement transpose
We should implement a transpose operation to transpose NxN matrix blocks (where N is batch width).
The interface should probably look like the one for haddp, e.g. taking a pointer of rows.

```cpp
template <class T, N>
void transpose(batch<T, N>* rows)
{
... inplace transpose ...
}
```

serge-sans-paille:
Fix #107 (6 years old!)

### DIFF
--- a/docs/source/api/data_transfer.rst
+++ b/docs/source/api/data_transfer.rst
@@ -62,6 +62,8 @@ In place:
 Between batches:
 
 +---------------------------------------+----------------------------------------------------+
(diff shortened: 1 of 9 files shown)


serge-sans-paille:
Note: WIP PR, currently only supports 32bits scalar types for SSE

serge-sans-paille:
We now have support for float and double on all platforms except riscv \o/
I'll merge that one and open bugs to keep track of the remaining bits

PR #1064, opened 2024-11-13
Added sphinx 6 to docs environment file

JohanMabille:
This should fix the rtd build.


JohanMabille:
The failure is unrelated to the change (already present on master).

PR #1069, opened 2024-12-09
Fix install location of .pc file in CMakeLists.txt

brechtsanders:
Fix install location of
`share/pkgconfig/xsimd.pc`
to
`lib/pkgconfig/xsimd.pc`


serge-sans-paille:
This partially conflicts with a507f83ead608fffc558b615b61b20f44f2f4672 . I like the argument referenced in the mentioned commit: https://github.com/xtensor-stack/xtl/commit/d877d94836aff4d0f727acf3eaab8f4880ecb625

PR #1088, opened 2025-02-24
Fixed kernel::store for booleans

JohanMabille:
@serge-sans-paille we missed that one ;)


serge-sans-paille:
wow, how did you found this one?

JohanMabille:
I'm updating xtensor to xsimd 13.1 ;)

PR #1089, opened 2025-02-24
Added broadcast overload for bool

JohanMabille:


include/xsimd/arch/generic/xsimd_generic_memory.hpp
```diff
@@ -32,6 +32,32 @@ namespace xsimd
...
+                static XSIMD_INLINE return_type run(T v) noexcept
+                {
+                    return return_type::broadcast(v);
+                }
+            };
+
+            template <class A>
+            struct broadcaster<bool, A>
+            {
+                using return_type = batch_bool<xsimd::as_unsigned_integer_t<bool>, A>;
```
serge-sans-paille, line 52: I'm not quite sure about that. You're creating a batch_bool out of bool, which looks fine at first glance, but a batch_bool neeads to know what kind of type it's operating one, why choosing  `xsimd::as_unsigned_integer_t<bool>` ? The user has no way to specify the actual type...

serge-sans-paille (changes requested):

include/xsimd/arch/generic/xsimd_generic_memory.hpp
serge-sans-paille: ok. I *think* we need to document that somewhere, either in the doxygen comment or by moving the specializatin at the API level so that the difference in behavior is clear when passing a `bool` to `xsimd::broadcast`. Bonus point if you mention `xsimd::broadcast_as` there ;-)

JohanMabille: Moving the specialization at the API level is what I did initially (where it atucally becomes an overload, since we cannot partially specialize functions). The issue is that the overload as only one template parameter, and with the default value of the first overload, the compiler does not pick the right overload, leading to compilation issues for regular cases (i.e. not bool). I'll add more documentation.

serge-sans-paille: gotcha

serge-sans-paille:
:bow:

PR #1095, opened 2025-03-18
add batch_bool kernel::load with requires_arch<avx512f>

junparser:


include/xsimd/arch/generic/xsimd_generic_memory.hpp
```diff
@@ -260,6 +260,19 @@ namespace xsimd
...
+        template <class A, class T>
+        XSIMD_INLINE batch_bool<T, A> load(bool const* mem, batch_bool<T, A>, requires_arch<generic>) noexcept
+        {
+            using batch_type = batch<T, A>;
+            batch_type ref(0);
+            constexpr auto size = batch_bool<T, A>::size;
+            alignas(A::alignment()) T buffer[size];
+            for (std::size_t i = 0; i < size; ++i)
+                buffer[i] = mem[i] ? 1 : 0;
+            return ref != batch_type::load_aligned(&buffer[0]);
```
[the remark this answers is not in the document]
junparser: IIUC, you mean load/store batch_bool with bits rather than bytes?  except the load/sotre bandwidth, it seems there are little difference between two of them.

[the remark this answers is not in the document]
serge-sans-paille: Actually https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=movemask&ig_expand=4606 looks like a good match, right?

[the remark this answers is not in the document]
junparser: hmm, do you mean __mmask64 mask = _mm512_cmpgt_epu8_mask(vecInput, _mm512_setzero_epi32()); for avx512?  _mm512_cmpgt_epu8_mask is under avx512bw, I'll add later.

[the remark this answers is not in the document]
serge-sans-paille: If you have the possibility to check in terms of performance, I've written another approach at computing masks that looks quite efficient: https://github.com/serge-sans-paille/fast-bitset-from-bool-array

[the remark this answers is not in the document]
junparser: I've tested the case with size 8 16 32 64 & cache hit, it is 2x~8x than the original version. and it's also better than avx512bw version with size 8 16 32.  maybe better to deal with size 64 in avx512bw. 
would you like to send another patches? or should i just update with this one?

junparser:
> I've tested the case with size 8 16 32 64 & cache hit, it is 2x~8x than the original version. and it's also better than avx512bw version with size 8 16 32. maybe better to deal with size 64 in avx512bw. would you like to send another patches? or should i just update with this one?
> Just update this one \o/

done, I've also tested the batch_bool store, mask store in avx512bw has the best performance, so just keep it.

include/xsimd/arch/generic/xsimd_generic_memory.hpp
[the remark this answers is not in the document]
serge-sans-paille: please just update this one, that's fine.

include/xsimd/arch/xsimd_avx512f.hpp
```diff
@@ -1193,6 +1193,38 @@ namespace xsimd
...
         }
 
+        namespace detail
+        {
+            // Generate a bitset from an array of boolean.
+            XSIMD_INLINE unsigned char tobitset(unsigned char unpacked[8])
+            {
+                uint64_t data;
+                memcpy(&data, unpacked, sizeof(uint64_t));
+
```
serge-sans-paille, line 1203: can you quote  https://github.com/serge-sans-paille/fast-bitset-from-bool-array here?

include/xsimd/arch/xsimd_avx512f.hpp
```diff
@@ -1193,6 +1193,38 @@ namespace xsimd
...
+        }
+
+        // load
+        template <class A, class T>
+        XSIMD_INLINE batch_bool<T, A> load(bool const* mem, batch_bool<T, A>, requires_arch<avx512f>) noexcept
+        {
+            using register_type = typename batch_bool<T, A>::register_type;
+            constexpr auto size = batch_bool<T, A>::size;
+            constexpr auto iter = size / 8;
+            assert((size % 8 == 0) && "ucorrectly size of bool batch");
```
serge-sans-paille, line 1218: incorrect?

serge-sans-paille: and should be a static_assert

junparser:
Address the comments. done

include/xsimd/types/xsimd_batch.hpp
```diff
@@ -968,11 +968,7 @@ namespace xsimd
     template <class T, class A>
     XSIMD_INLINE batch_bool<T, A> batch_bool<T, A>::load_aligned(bool const* mem) noexcept
     {
-        batch_type ref(0);
-        alignas(A::alignment()) T buffer[size];
-        for (std::size_t i = 0; i < size; ++i)
-            buffer[i] = mem[i] ? 1 : 0;
-        return ref != batch_type::load_aligned(&buffer[0]);
+        return kernel::load<A>(mem, batch_bool<T, A>(), A {});
```
serge-sans-paille, line 971: Hell, I missed one point: given the implementation specialization, we now need to distinguish between the aligned and unaligned case.
Sorry for the extra back-and-forth :-/

The fact that CI passes probably means we need extra tests for this.

junparser: I dont get the point, do we need aligned and unaligned case for bool load?

junparser: and we already have  `template <class T, class A>
    XSIMD_INLINE batch_bool<T, A> batch_bool<T, A>::load_unaligned(bool const* mem) noexcept` in https://github.com/xtensor-stack/xsimd/pull/1095/files#diff-4225c7e955e5d693f65f483a7d38b0f6af3e229544cd037f4da05d6b5de2e8b2R975

serge-sans-paille: well if `mem` is unaligned and current unaligned loader calls the aligned loaded. If we end up (not the case in your avx512 implementation, but it actually could/should) calling an aligned load in that case, we're blessed with a segfault :-)

junparser: OK. you mean `A::alignment()` for the access of bool const* mem. I'm not sure how we should deal with it here, since the function name`  kernel::load/sotre ` means to me is that the alignment has been ignored here. any suggestion?

junparser: since this is Implementation defined behavior,  how about manually set `mem` unaligned when we test  `batch_bool_type::load_unaligned ` in test_batch_bool.cpp ?

serge-sans-paille: > OK. you mean `A::alignment()` for the access of bool const* mem. I'm not sure how we should deal with it here, since the function name` kernel::load/sotre` means to me is that the alignment has been ignored here. any suggestion?

We generally provide two versions for the function: load_aligned and load_unaligned

junparser: > > OK. you mean `A::alignment()` for the access of bool const* mem. I'm not sure how we should deal with it here, since the function name` kernel::load/sotre` means to me is that the alignment has been ignored here. any suggestion?
> 
> We generally provide two versions for the function: load_aligned and load_unaligned

batch_bool keeps using load/store before this patch. If we want unify batch_bool load/store into load_aligned and load_unaligned, it is better to do in another pr.

serge-sans-paille:
Thanks! I'll do the aligned / unaligned part, you've already done a lot for this PR :bow:

PR #1096, opened 2025-03-19
Add insert&incr_if&decr_if in avx512

junparser:


serge-sans-paille:
> @serge-sans-paille hi, would you please take a look at Emscripten build? thanks!

Let's see if https://github.com/xtensor-stack/xsimd/pull/1098 does the trick

serge-sans-paille:
You can rebase on master, Emscripten CI should be ok now.

include/xsimd/arch/xsimd_avx512f.hpp
```diff
@@ -1121,13 +1121,15 @@ namespace xsimd
         XSIMD_INLINE batch<float, A> insert(batch<float, A> const& self, float val, index<I>, requires_arch<avx512f>) noexcept
         {
 
-            return _mm512_castsi512_ps(_mm512_mask_set1_epi32(_mm512_castps_si512(self), __mmask16(1 << (I & 15)), *(int32_t*)&val));
+            int32_t* tmp = reinterpret_cast<int32_t*>(&val);
```
serge-sans-paille, line 1124: This is UB, the correct way to do so is through a `memcpy` and we have this wrapped up as `xsimd::bit_cast<int32_t>(val)`

junparser: OK, done, generate same instructions under O3

junparser:
@serge-sans-paille  kindly ping

PR #1102, opened 2025-03-28
Support compress&expand in avx512vbmi2

junparser:


junparser:
@serge-sans-paille `[Linux build / gcc 11 - avx512vnni (pull_request)](https://github.com/xtensor-stack/xsimd/actions/runs/14120454206/job/39559599836?pr=1102)` looks weird, I can't reproduce it with gcc&gcc12, what's the cflags used here?

junparser:
> @serge-sans-paille `[Linux build / gcc 11 - avx512vnni (pull_request)](https://github.com/xtensor-stack/xsimd/actions/runs/14120454206/job/39559599836?pr=1102)` looks weird, I can't reproduce it with gcc&gcc12, what's the cflags used here?

this caused by  _mm512_shuffle_epi8 for batch<uint8> swizzle in avx512bw which is incorrect. use _mm512_permutexvar_epi8 in avx512vbmi instead.

junparser:
@serge-sans-paille , `CMake Error at benchmark/CMakeLists.txt:12 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake. 
`, seems we hit cmake issue. would you please take a look?

serge-sans-paille:
@junparser https://github.com/xtensor-stack/xsimd/pull/1104 should do the trick

junparser:
> @junparser #1104 should do the trick

Also, after this commit, would you like to add one more ` gcc 11 - avx512vbmi2 ` in ci?

serge-sans-paille:
> gcc 11 - avx512vbmi2

https://github.com/xtensor-stack/xsimd/pull/1105

serge-sans-paille:
+ https://github.com/xtensor-stack/xsimd/pull/1106

junparser:
> * [Feature/vbmi2 ci #1106](https://github.com/xtensor-stack/xsimd/pull/1106)
Thanks, now this pr passes avx512vbmi2.  Also is it possible to add something like `[Linux build / clang18/20 - avx512vbmi2 ` in ci?

include/xsimd/config/xsimd_arch.hpp
```diff
@@ -162,7 +162,7 @@ namespace xsimd
     } // namespace detail
 
     using all_x86_architectures = arch_list<
-        avx512vnni<avx512vbmi>, avx512vbmi, avx512ifma, avx512pf, avx512vnni<avx512bw>, avx512bw, avx512er, avx512dq, avx512cd, avx512f,
+        avx512vnni<avx512vbmi2>, avx512vbmi2, avx512vbmi, avx512ifma, avx512pf, avx512vnni<avx512bw>, avx512bw, avx512er, avx512dq, avx512cd, avx512f,
```
serge-sans-paille, line 165: Can you motivate this change? Why would composition between avx512 VNNI be more legit with VMBI2 rather than VBMI?

include/xsimd/arch/xsimd_isa.hpp
```diff
@@ -92,11 +92,15 @@
...
+#include "./xsimd_avx512vbmi2.hpp"
+#endif
+
 #if XSIMD_WITH_AVX512VNNI_AVX512BW
 #include "./xsimd_avx512vnni_avx512bw.hpp"
 #endif
 
-#if XSIMD_WITH_AVX512VNNI_AVX512VBMI
+#if XSIMD_WITH_AVX512VNNI_AVX512VBMI2
 #include "./xsimd_avx512vnni_avx512vbmi.hpp"
```
serge-sans-paille, line 104: if we are to make this change, we will need some file renaming here

include/xsimd/config/xsimd_arch.hpp
junparser: I made this change based on https://en.wikichip.org/wiki/x86/avx512_vnni. The table shows that all of the arch have vnni as well as vbmi2.

PR #1103, opened 2025-03-28
Reorder make_batch_constant template arguments

serge-sans-paille:
So that we can use a default architecture, which is consistent with other API.

This is an API breaking change


serge-sans-paille:
@JohanMabille this is an API break

Issue #1107, opened 2025-04-07
reduce seems slow


DiamonDinoia:
Hi all! Thank you for developing this library. It is super useful in multiple projects!

I noticed something that might be an issue. But I am not sure.

I have a code where I multiply a complex valued (a) array with a real valued array (ker):
Basically, I need to multiply each element of 'a' twice. 
My code is as follows:
```cpp
auto func(real a1, real a2 complex ker):
     // this trick halves the number of loads for ker also the reason why I use a1 and a2 instead of a
    const auto low = xsimd::zip_lo(ker, ker);
    const auto high= xsimd::zip_hi(ker, ker);
    const auto res0 = a1 * low;
    const auto res1 = a2 * high;
```
what I noticed is that the original implementation of reduce_add on my machine can be optimized. Is it possible to have a split function that returns low and hi? By doing split + add multiple times my code is 7 times faster.

I have pushed the benchmarks here:
https://github.com/DiamonDinoia/cpp-learning/tree/master/xsimd

it results in the following performance:

|               ns/op |                op/s |    err% |          ins/op |          cyc/op |    IPC |         bra/op |   miss% |     total | benchmark
|--------------------:|--------------------:|--------:|----------------:|----------------:|-------:|---------------:|--------:|----------:|:----------
|                6.96 |      143,690,879.59 |    0.6% |           19.00 |           21.47 |  0.885 |           0.00 |    0.0% |      0.01 | `add+store`
|                2.31 |      432,949,727.65 |    0.6% |           24.00 |            7.11 |  3.374 |           0.00 |    0.0% |      0.01 | `hsum`
|                3.81 |      262,211,901.24 |    0.1% |           36.00 |           11.75 |  3.064 |           2.00 |    0.0% |      0.01 | `reduce_add`
|                2.59 |      385,491,672.62 |    0.2% |           20.00 |            7.99 |  2.503 |           0.00 |    0.0% |      0.01 | `union pun`
|                1.18 |      846,618,297.70 |    0.9% |           17.00 |            3.64 |  4.672 |           0.00 |    0.0% |      0.01 | `double union pun`

I tweaked master a bit in https://github.com/DiamonDinoia/xsimd/tree/hadd-tweaks
and I got:

|               ns/op |                op/s |    err% |          ins/op |          cyc/op |    IPC |         bra/op |   miss% |     total | benchmark
[...]

PR #1108, opened 2025-04-08
Do not use hadd_pd to implement reduce_add

serge-sans-paille:
It's generally slower than the sse2 version due to a latency of 5 (!) for hadd_pd.

Related to #1107

### DIFF
--- a/include/xsimd/arch/xsimd_avx.hpp
+++ b/include/xsimd/arch/xsimd_avx.hpp
@@ -1064,22 +1064,10 @@ namespace xsimd
             tmp = _mm256_hadd_ps(tmp, tmp);
             return _mm_cvtss_f32(_mm256_extractf128_ps(tmp, 0));
         }
-        template <class A>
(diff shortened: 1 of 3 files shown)


serge-sans-paille:
@DiamonDinoia : why did you guarded the code here: https://github.com/xtensor-stack/xsimd/commit/174c4756a3d9697e085d1bfb923e04efa5ea6879#diff-c4f5d7f47f45c737cc1723af31069217834daf7da679a0b4ff255a4a6ae73c83R1413 this "intrinsic" seems to be standard - and it passes our validation without the guard.

DiamonDinoia:
I see, I  thought it was an icc only intrinsic from the way it was used in Agner's vcl. Altough, I have been using fine on any compiler.

serge-sans-paille:
Great. I'm going to merge this patchset then. Would you mind doing the same investigation for single precision float?

DiamonDinoia:
Sure, I will have a look next week when I have a moment.

Would you mind considering the API I was suggesting for reducing interleaved complex?

PR #1112, opened 2025-04-26
Faster reduce_add on avx

serge-sans-paille:
Just forward to sse after a split and a sum.


serge-sans-paille:
the genrated code looks similar but at least we've improved the generic case and we have the same code path.

PR #1123, opened 2025-06-12
Enable xsimd with only emulated arch.

### LINKED ISSUE #1122: xsimd::emulated requires at least one supported arch
The definition of [XSIMD_NO_SUPPORTED_ARCHITECTURE](https://github.com/xtensor-stack/xsimd/blob/master/include/xsimd/config/xsimd_config.hpp#L469) does not check for XSIMD_WITH_EMULATED, so on platforms without any supported architecture, the emulated one isn't available as well.

Is this made on purpose or was the `&& !XSIMD_WITH_EMULATED` missed in https://github.com/xtensor-stack/xsimd/blob/master/include/xsimd/config/xsimd_config.hpp#L469 ?

degasus:
So do not define XSIMD_NO_SUPPORTED_ARCHITECTURE if XSIMD_WITH_EMULATED is given. This allows to use an emulated mode if no physical arch is given.

My personal use case is to generate scalar code on Windows ARM64 from a single implementation.

Fixes #1122


serge-sans-paille:
Thanks!

PR #1132, opened 2025-07-01
Reduce_mul implementation

emrys53:
While using XSIMD in our library, we realized that a dedicated reduce_mul function was missing. This PR introduces a reduce_mul implementation for all architectures and supported types, with architecture-specific optimizations for SSE, AVX, and AVX512 instruction sets. Relevant unit tests are included.

I was unable to test the AVX512-specific implementation on my machine, which only supports AVX2. For AVX512, I used _mm512_reduce_mul_{type} intrinsics. According to Intel's documentation, these are sequential rather than parallel. Since I couldn't benchmark or compare alternative implementations, I left it as-is. I'm open to suggestions for improvement.
Motivation

Although XSIMD provides a general reduce function, using it with std::complex<float> or std::complex<double> fails, as as_unsigned_integer_t<T> is not specialized for complex types. I'm not sure whether this is intended behavior, but I wanted to mention it for context. Providing a reduce_mul directly avoids this issue.

This is my first contribution to XSIMD, and I’ve only been using the library for about a week. Feedback is very welcome, and I’m happy to adjust or improve the code based on your review.

### DIFF
--- a/include/xsimd/arch/common/xsimd_common_arithmetic.hpp
+++ b/include/xsimd/arch/common/xsimd_common_arithmetic.hpp
@@ -141,6 +141,20 @@ namespace xsimd
             return res;
         }
 
+        //  hmul
+        template <class A, class T, class /*=typename std::enable_if<std::is_integral<T>::value, void>::type*/>
+        XSIMD_INLINE T hmul(batch<T, A> const& self, requires_arch<common>) noexcept
+        {
+            alignas(A::alignment()) T buffer[batch<T, A>::size];
+            self.store_aligned(buffer);
+            T res = 1;
+            for (T val : buffer)
+            {
+                res *= val;
+            }
+            return res;
+        }
+
         // incr
         template <class A, class T>
         XSIMD_INLINE batch<T, A> incr(batch<T, A> const& self, requires_arch<common>) noexcept
--- a/include/xsimd/arch/common/xsimd_common_details.hpp
+++ b/include/xsimd/arch/common/xsimd_common_details.hpp
@@ -74,6 +74,8 @@ namespace xsimd
     XSIMD_INLINE batch<as_integer_t<T>, A> nearbyint_as_int(const batch<T, A>& x) noexcept;
     template <class T, class A>
     XSIMD_INLINE T reduce_add(batch<T, A> const&) noexcept;
+    template<class T, class A>
+    XSIMD_INLINE T reduce_mul(batch<T, A> const&) noexcept;
     template <class T, class A>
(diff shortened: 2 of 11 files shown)


include/xsimd/arch/common/xsimd_common_arithmetic.hpp
```diff
@@ -141,6 +141,20 @@ namespace xsimd
             return res;
         }
 
+        //  hmul
+        template <class A, class T, class /*=typename std::enable_if<std::is_integral<T>::value, void>::type*/>
+        XSIMD_INLINE T hmul(batch<T, A> const& self, requires_arch<common>) noexcept
```
serge-sans-paille, line 146: shouldn't this just be the generic implementation of `reduce_mul`, and if so, why not using the current generic reduction implementation?

emrys53: This function is only used for integral types with sizeof(T) < 4. The reason I implemented like this was because for example when testing reduce_mul for uint8_t with AVX machine. What happens is that it has 32 elements which means general reduce function multiplies and swizzle the batch 5 times (32->16->8->4->2->1) and in the end it calls self.get(0) (which is just too inefficient anyway [issue #1133](https://github.com/xtensor-stack/xsimd/issues/1133)). So I thought it would be more efficient to implement it this way. What it does with uint8_t in my implementation is that it first splits the batch into 2 equal sse4_2 batches multiplies them together and then calls hmul.

emrys53: Also the swizzle function for uint8_t is just loading the batch to a buffer and storing the result back. So using that 5 times (in AVX)  is just too inefficient compared to split_avx function which splits it using SIMD intrinsics.

emrys53: @serge-sans-paille you haven't responded back. If you are open to the idea of reduce_mul implementation. I can fix the code so that it passes all the tests. If you have feedbacks or recommendations to the code I have written, you can write it here and I can fix it after we discuss.

DiamonDinoia:
I'm also interested in this. @emrys53 could you rebase master?
@serge-sans-paille what do you think needs to be changes to integrate it?

serge-sans-paille:
@emrys53 I've submitted #1163 as a port of this patch for other targets, as well as some support for other architectures. I was strongly motivated and inspired by your work, so please don't take it as an offense, rather as a tribute!

JohanMabille:
Closing this in favor of #1163, thanks a lot @emrys53 for initiating the work on this!

Issue #1133, opened 2025-07-01
Specialized functions for batch.get(0) case.


emrys53:
### Performance Issue: Inefficient `get(0)` Implementation

**Summary**  
The current implementation of `get()` for `batch<T, A>` always stores the entire batch into an aligned buffer and returns `buffer[I]`, even for `I == 0`. This introduces unnecessary overhead when only the first element is needed, which is common in reduction operations.

```cpp
template <class A, size_t I, class T>
XSIMD_INLINE T get(batch<T, A> const& self, ::xsimd::index<I>, requires_arch<common>) noexcept
{
    alignas(A::alignment()) T buffer[batch<T, A>::size];
    self.store_aligned(&buffer[0]);
    return buffer[I];
}
```

---

### Problem

Accessing the first element (`get(0)`) via full `store_aligned` is much more expensive than necessary. In reduce function, self.get(0) in the end which adds unneccassary cost. If we are loading the batch in a buffer, the performance benefit of using reduce function disappears as we can just load everything to buffer and then implement in a scalar fashion. The entire purpose of reduction operations are to avoid directly copying to data to a buffer.---

### Proposed Solution

Introduce a `first()` helper for efficiently accessing the first lane of a batch:

```cpp
template <class T, class A>
XSIMD_INLINE T first( batch<T, A> const& self) noexcept
{
    // Example: platform-specific optimized intrinsic
    return self.get_first(); // or use appropriate intrinsic depending on A
}
```
This could avoid the store_aligned() and instead use more efficient intrinsics like:

    _mm_cvtsd_f64() (SSE2)

    _mm256_castps256_ps128() + _mm_cvtss_f32() (AVX)

    _mm512_cvtss_f32() (AVX512)

This would dramatically improve performance for reductions and any other first-element access patterns.

This would eliminate the cost of storing the entire batch just to access the first element.

serge-sans-paille:
Thanks for the report! The `get` method is not meant to be used in production, it's just an helper for debugging etc. If we happen to use it elsewhere, that's a bug and we should fix it. Maybe by using.... `get_first` as you suggest? I'm not sold on the name, but I can see where such a method could be useful.

emrys53:
[...]

PR #1134, opened 2025-07-02
Added first() function for SSE/AVX/AVX512F

emrys53:
This pr implements new first() function for batch types. It is an improvement from the currently used self.get(0) as first() uses intrinsics to extract the first element. Addresses inefficiencies mentioned in [issue #1133](https://github.com/xtensor-stack/xsimd/issues/1133).
I have only implemented for SSE/AVX/AVX512F architectures @serge-sans-paille will implement for other architectures.
I also haven't replace any .get(0) functions with first(). That needs to be done as well

### DIFF
--- a/include/xsimd/arch/xsimd_sse2.hpp
+++ b/include/xsimd/arch/xsimd_sse2.hpp
@@ -1782,6 +1782,57 @@ namespace xsimd
         {
             return _mm_unpacklo_pd(self, other);
         }
+
+        // first
+        template <class A>
+        XSIMD_INLINE float first(batch<float, A> const& self, requires_arch<sse2>) noexcept
+        {
+            return _mm_cvtss_f32(self);
+        }
+
+        template <class A>
+        XSIMD_INLINE double first(batch<double, A> const& self, requires_arch<sse2>) noexcept
+        {
+            return _mm_cvtsd_f64(self);
+        }
+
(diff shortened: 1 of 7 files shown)


include/xsimd/arch/xsimd_sse2.hpp
```diff
@@ -1782,6 +1782,57 @@ namespace xsimd
...
+            }
+            else
+            {
+                assert(false && "unsupported arch/op combination");
+                return {};
+            }
+        }
+
+        template <class A, class T>
+        XSIMD_INLINE std::complex<T> first(batch<std::complex<T>, A> const& self, requires_arch<sse2>) noexcept
```
serge-sans-paille: shouldn't that be the common implementation? It would avoid some redunduncy

emrys53: @serge-sans-paille you are definetly right. I just fixed it for complex and batch_bool types and pushed it.

serge-sans-paille:
Looks good, but there is some formatting and windows - specific issue to handle, can you have a look?

emrys53:
These last 2 commits should be able to fix the previous issues

emrys53:
oh nice it passed all the tests. Now we need to implement it for other architectures and replace all .get(0) with first()

serge-sans-paille:
> oh nice it passed all the tests. Now we need to implement it for other architectures and replace all .get(0) with first()

I'll handle that. thanks!

PR #1136, opened 2025-07-02
Adding fmaddsub

DiamonDinoia:
Hi all,

I have added support and tests for fmaddsub since it is useful in my scientific computations.
I tested it on x86 but I do not have access to arm/neon.

I proposed fmas but I have no strong opinion on the name of the function I picked something that seems to fit the style.

Let me know what you think and if/how we can get the new API included!

Thanks,
Marco

### DIFF
--- a/include/xsimd/arch/common/xsimd_common_arithmetic.hpp
+++ b/include/xsimd/arch/common/xsimd_common_arithmetic.hpp
@@ -127,6 +127,18 @@ namespace xsimd
             return { res_r, res_i };
         }
 
+        // fmas
+        template <class A, class T>
+        XSIMD_INLINE batch<T, A> fmas(batch<T, A> const& x, batch<T, A> const& y, batch<T, A> const& z, requires_arch<common>) noexcept
+        {
+            struct even_lane
+            {
+                static constexpr bool get(unsigned const i, unsigned) noexcept { return (i & 1u) == 0; }
+            };
+            const auto mask = make_batch_bool_constant<T, even_lane, A>();
+            return fma(x, y, select(mask, neg(z), z));
+        }
+
         // hadd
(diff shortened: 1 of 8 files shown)


include/xsimd/arch/common/xsimd_common_arithmetic.hpp
```diff
@@ -127,6 +127,18 @@ namespace xsimd
...
+        // fmas
+        template <class A, class T>
+        XSIMD_INLINE batch<T, A> fmas(batch<T, A> const& x, batch<T, A> const& y, batch<T, A> const& z, requires_arch<common>) noexcept
+        {
+            struct even_lane
+            {
+                static constexpr bool get(unsigned const i, unsigned) noexcept { return (i & 1u) == 0; }
+            };
+            const auto mask = make_batch_bool_constant<T, even_lane, A>();
+            return fma(x, y, select(mask, neg(z), z));
```
serge-sans-paille, line 139: Is select a better option compared to a multiply?

DiamonDinoia: I think multiply is overly complicated. If we extend batch_constant to double (remove bitwise operators) then it might be worth it.
```cpp
struct imag_neg {
        static constexpr int get(const unsigned i, unsigned) noexcept { return (i & 1u) ? 1.0 : -1.0; }
    };

    // Generator first, then arch
    const auto mask = xsimd::batch_cast<double>(xsimd::make_batch_constant<xsimd::as_integer_t<double>, imag_neg, arch>().as_batch());
```

Also, I do not see a big difference on my machine:
image

(swizzle/multiply vs swizzle/select)

serge-sans-paille: thanks for going all the way down to the experiment!

PR #1144, opened 2025-07-13
Feature/tentative altivec support

serge-sans-paille:


serge-sans-paille:
@JohanMabille can you have a look? I can improve the coverage but it's a minimal support, quite a decent start

serge-sans-paille:
We could rename the feature VSX instead of altivec, and eventually add VMX (an alias to VMX) in the future, that may make more sense.

serge-sans-paille:
> We could rename the feature VSX instead of altivec, and eventually add VMX (an alias to VMX) in the future, that may make more sense.

I ended up doing just that

serge-sans-paille:
@JohanMabille some context to help during your review: altivec is a brand for VMX, and a later extension was added, called VSX, in the spirit of, say SSE2 and SSSE3.
I did *not* implement pure VMX support because it lacks double support, accurate floating point division etc. We may end up supporting it later but that seemed out of scope and too much effort for a summer run.

serge-sans-paille:
Merged as part of #1151

PR #1145, opened 2025-07-18
Fix build with Clang for Windows

Clemoot:
I am trying to compile and use xsimd with Clang on Windows.

I just discovered that Clang for Windows doesn't define `__GNUC__` but instead defines `_MSC_VER`. That makes the compiler choose a version of `XSIMD_INLINE` that it does not support.

 The only solution I found that do supports both platforms is by extending `#if defined(__GNUC__)` with `(defined(_MSC_VER) && defined(__clang__)`.

### DIFF
--- a/include/xsimd/config/xsimd_inline.hpp
+++ b/include/xsimd/config/xsimd_inline.hpp
@@ -12,8 +12,10 @@
 #ifndef XSIMD_INLINE_HPP
 #define XSIMD_INLINE_HPP
 
-#if defined(__GNUC__)
+#if defined __has_attribute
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
Alternative: https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fattribute.html
See https://godbolt.org/z/YYjfGzqYM

serge-sans-paille:
(and thanks for spotting this!)

Clemoot:
I updated the changes, which in fact are way cleaner with your suggestion.

PR #1146, opened 2025-07-21
Fix AVX512DQ.

degasus:
This file was completely disabled and never included, and so not tested and broken.

This patches fixes:
* Missing include in the _isa.hpp
* Typo in the include guard
* Variable name within fast_cast
* Requirements to avx512dq

I have to be honest: I have no clue if those implementations are correct. My patch "just" enables the previously dead code. Please see this PR more as a bug report than as a fully tested patch.


include/xsimd/arch/xsimd_isa.hpp
```diff
@@ -72,6 +72,10 @@
 #include "./xsimd_avx512f.hpp"
 #endif
 
+#if XSIMD_WITH_AVX512DQ
```
degasus, line 75: FYI: XSIMD_WITH_AVX512CD is missing as well. However it does not define anything, but might yield the same issue again.

serge-sans-paille:
Thanks!

DiamonDinoia:
Interesting. I did the same changes here :)
[DiamonDinoia/xsimd@improving-avx512-swizzle](https://github.com/DiamonDinoia/xsimd/tree/improving-avx512-swizzle)

PR #1147, opened 2025-07-21
avx512bw,vbmi: Optimize shift_left and shift_right.

degasus:
* Use `_maskz_permutexvar` instead of `permute + and` on avx512bw, too
* Use `xsimd::make_batch_constant` and so `_mm512_set` instead of `_mm512_load(constexpr std::array)`

The first one is just a tiny optimization already done in the AVX512VBMI implementation.

The second patch makes it easier for the compiler to move this constants into the .text section. GCC is not affected, MSVC sadly is affected *a lot*. It used to generated 8 times `mov [rbp + N], const` with a single `vmovdqu32 zmm, [rdp]` afterwards - poor MSVC.... And worse for the store_forwarding unit....

### DIFF
--- a/include/xsimd/arch/xsimd_avx512bw.hpp
+++ b/include/xsimd/arch/xsimd_avx512bw.hpp
@@ -486,17 +486,22 @@ namespace xsimd
         // slide_left
         namespace detail
         {
-            template <size_t... Is>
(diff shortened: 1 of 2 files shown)


DiamonDinoia:
Interesting. I did the same changes here :)
https://github.com/DiamonDinoia/xsimd/tree/improving-avx512-swizzle

PR #1148, opened 2025-07-22
avx512f: implement slide_left and slide_right.

degasus:
With a fast path for N = 4*i and a split version else (inspired from avx512bw).

As the vpermd actually has the lower latency on some intel cpus compared to vpermw / vpermb, also use it instead of the avx512bw and avx512vbmi implementations if trivially possible.

Also reverse the order of the slow path for lower latency. It used to be:
```
SLR -> PERM ->
               OR -> PERM
SLL         ->
```

Now the latency is reduced to:
```
SLR -> PERM ->
               OR
SLL -> PERM ->
```

And it should now generate even better code on avx512bw with N=63 with only one PERM (as already done for N=1).

For N=16,32,48, it prefers vshufi32x4 over vpermd for lower latency on Zen4 and decreased register usage.

### DIFF
--- a/include/xsimd/arch/xsimd_avx512bw.hpp
+++ b/include/xsimd/arch/xsimd_avx512bw.hpp
@@ -484,113 +484,25 @@ namespace xsimd
         }
 
         // slide_left
-        namespace detail
-        {
-            struct make_slide_perm_hi
-            {
-                static constexpr uint64_t get(size_t i, size_t)
-                {
-                    return i == 0 ? 8 : i - 1;
-                }
-            };
-
-            template <size_t N>
-            struct make_slide_left_pattern
-            {
-                static constexpr uint16_t get(size_t i, size_t)
-                {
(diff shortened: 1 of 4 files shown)


degasus:
FYI, the reason why I'm interested in those slide_left is because my code requires a `cumsum`:
```
batch cumsum(batch x) {
  for (size_t i = 0; i < std::bit_width(batch::size - 1); i++) {
    x += xsimd::slide_left<(sizeof(batch::value_type) << i)>(x); 
  }
  return x;
}
```

Is this a method you're interested in as part of XSIMD's public API?

serge-sans-paille:
cumsum... isn't that  https://xsimd.readthedocs.io/en/latest/api/reducer_index.html#_CPPv4I00E10reduce_add1TRK5batchI1T1AE ?
EDIT: it's not :-/ yeah of course add it as a generic operation!

PR #1149, opened 2025-07-23
DOC: Move 'count' to the batch-bool 'Reducers' section of the docs.

WarrenWeckesser:
<strike>`count` is in the doxygen group `batch_reducers`, and it shows up below the table of reduction operators along with `reduce`, `reduce_add`, etc., in the generated HTML, so it might as well be in the table.</strike>

Updated the PR so now `count` is moved to the 'Reducers' section of the batch-bool docs.

### DIFF
--- a/docs/source/api/reducer_index.rst
+++ b/docs/source/api/reducer_index.rst
@@ -41,6 +41,8 @@ Reduction operators
 | :cpp:func:`haddp`                     | horizontal sum across batches                      |
 +---------------------------------------+----------------------------------------------------+
 
+Also see the `batch_bool` :ref:`xsimd-batch-bool-reducers`.
+
 ----
 
 .. doxygengroup:: batch_reducers
--- a/docs/source/api/xsimd_batch_bool.rst
+++ b/docs/source/api/xsimd_batch_bool.rst
@@ -20,6 +20,8 @@ Logical operators
    :project: xsimd
    :content-only:
 
+.. _xsimd-batch-bool-reducers:
+
 Reducers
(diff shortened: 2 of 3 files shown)


serge-sans-paille:
`count` is a `batch_bool` only operator, it would be good to make that explicit, maybe by using two tables, one for `batch` and one for `batch_bool` ?

WarrenWeckesser:
I don't know.  I was originally looking for `all` and `any` reductions when I found this page and noticed that `count` was among the functions tagged as `batch_reducers` shown below the table but is not in the table.

I eventually found `any` and `all` in the [Batch of conditions](https://xsimd.readthedocs.io/en/latest/api/xsimd_batch_bool.html) page. Perhaps `count` should be listed there.

WarrenWeckesser:
> Perhaps count should be listed there.

It looks like this can be done by moving `count` from the doxygen `batch_reducers` group to the `batch_bool_reducers` group.  I'll undo my change to the table and make that group change instead.  If I can figure out how to do a cross-reference with doxygen/sphinx, I'll add a "See also" note to the [Reduction operations](https://xsimd.readthedocs.io/en/latest/api/reducer_index.html) page.

WarrenWeckesser:
I pushed a new update.  Now `count` will show up in the 'Reducers' section of the batch-bool docs.  I also added a note in the 'Reduction operators' page (where `count` used to show up) that links to the batch-bool 'Reducers' docs.

serge-sans-paille:
Looks good, thanks!

PR #1150, opened 2025-07-24
MSVC/AVX512F: Fix batch_const with 16b and 8b integers.

degasus:
There is no `_setr_` intrinsic for 16b and 8b types, so we have to flip the order ourself.

Honestly, there is a CI for catching such bugs. How can we missed to test this case?


degasus:
Oh dear: https://github.com/xtensor-stack/xsimd/blob/master/.github/workflows/windows.yml#L53 ...

FYI: With this patch, `ninja xtest` is working fine on my Windows+AVX512 machine.

serge-sans-paille:
Thanks! I'll have a look on this CI thing, it's frightening.

serge-sans-paille:
ok, we do test avx512 on Linux, (through emulation) and not on windows (see https://github.com/actions/runner/issues/1069)

PR #1151, opened 2025-07-28
Initial support for Altivec

serge-sans-paille:
Only supports altivec with vxs extension, tested on both little and big endian architectures.

Notable changes:

- document the fact that slide_left and slide_right are unreliable on big endian architecture

- gcc chokes on force_inline for altivec, so dismiss it

- some common code were making assumption of conversion between batch_bool registers and batch_registers that proved to be difficult for altivec type system, make the conversion explicit.


serge-sans-paille:
@Jenkins-J VSX support was added because I had a genuine interest in VMX/VSX and I was curious about how that would translate to xsimd. I'm glad to see it would benefit to Velox!
BTW, I submitted https://github.com/facebookincubator/velox/pull/13261 to Velox some times ago, just so that Velox would be compatible with the latest release of xsimd, but it's not merged yet, if you could have a look?

PR #1154, opened 2025-08-08
Use dedicated unpacking intrinsic instead of a shuffle in sse2 swizzl…

serge-sans-paille:
…e / short


serge-sans-paille:
Merged as part of #1155

PR #1162, opened 2025-08-21
Proposal load/store masked

DiamonDinoia:
Dear @serge-sans-paille,

I made a rough implementation of masked load/store. Before I hash it out. Can I have some early feedback? 

Thanks,
Marco

### DIFF
--- a/include/xsimd/arch/common/xsimd_common_arithmetic.hpp
+++ b/include/xsimd/arch/common/xsimd_common_arithmetic.hpp
@@ -16,6 +16,7 @@
 #include <limits>
 #include <type_traits>
 
+#include "../../types/xsimd_batch_constant.hpp"
 #include "./xsimd_common_details.hpp"
 
 namespace xsimd
--- a/include/xsimd/arch/common/xsimd_common_memory.hpp
+++ b/include/xsimd/arch/common/xsimd_common_memory.hpp
@@ -13,6 +13,7 @@
 #define XSIMD_COMMON_MEMORY_HPP
 
 #include <algorithm>
+#include <array>
 #include <complex>
 #include <stdexcept>
 
@@ -339,6 +340,102 @@ namespace xsimd
         XSIMD_INLINE batch<T_out, A> load_unaligned(T_in const* mem, convert<T_out> cvt, requires_arch<common>) noexcept
         {
             return detail::load_unaligned<A>(mem, cvt, common {}, detail::conversion_type<A, T_in, T_out> {});
+        }
+
+        template <class A, class T>
+        XSIMD_INLINE batch<T, A> load(T const* mem, aligned_mode, requires_arch<A>) noexcept
+        {
+            return load_aligned<A>(mem, convert<T> {}, A {});
+        }
+
+        template <class A, class T>
+        XSIMD_INLINE batch<T, A> load(T const* mem, unaligned_mode, requires_arch<A>) noexcept
+        {
+            return load_unaligned<A>(mem, convert<T> {}, A {});
+        }
+
+        template <class A, class T_in, class T_out, bool... Values, class alignment>
+        XSIMD_INLINE batch<T_out, A>
+        load_masked(T_in const* mem, batch_bool_constant<T_out, A, Values...>, convert<T_out>, alignment, requires_arch<common>) noexcept
+        {
+            constexpr std::size_t size = batch<T_out, A>::size;
+            alignas(A::alignment()) std::array<T_out, size> buffer {};
+            constexpr bool mask[size] = { Values... };
+
+            for (std::size_t i = 0; i < size; ++i)
+                buffer[i] = mask[i] ? static_cast<T_out>(mem[i]) : T_out(0);
+
+            return batch<T_out, A>::load(buffer.data(), aligned_mode {});
+        }
+
+        template <class A, class T_in, class T_out, bool... Values, class alignment>
+        XSIMD_INLINE void
+        store_masked(T_out* mem, batch<T_in, A> const& src, batch_bool_constant<T_in, A, Values...>, alignment, requires_arch<common>) noexcept
+        {
+            constexpr std::size_t size = batch<T_in, A>::size;
+            constexpr bool mask[size] = { Values... };
+
+            for (std::size_t i = 0; i < size; ++i)
+                if (mask[i])
+                {
+                    mem[i] = static_cast<T_out>(src.get(i));
+                }
+        }
+
+        template <class A, bool... Values, class Mode>
+        XSIMD_INLINE batch<int32_t, A> load_masked(int32_t const* mem, batch_bool_constant<int32_t, A, Values...>, convert<int32_t>, Mode, requires_arch<A>) noexcept
+        {
+            const auto f = load_masked<A>(reinterpret_cast<const float*>(mem), batch_bool_constant<float, A, Values...> {}, convert<float> {}, Mode {}, A {});
+            return bitwise_cast<int32_t>(f);
+        }
+
+        template <class A, bool... Values, class Mode>
+        XSIMD_INLINE batch<uint32_t, A> load_masked(uint32_t const* mem, batch_bool_constant<uint32_t, A, Values...>, convert<uint32_t>, Mode, requires_arch<A>) noexcept
+        {
+            const auto f = load_masked<A>(reinterpret_cast<const float*>(mem), batch_bool_constant<float, A, Values...> {}, convert<float> {}, Mode {}, A {});
+            return bitwise_cast<uint32_t>(f);
+        }
+
+        template <class A, bool... Values, class Mode>
+        XSIMD_INLINE typename std::enable_if<types::has_simd_register<double, A>::value, batch<int64_t, A>>::type
+        load_masked(int64_t const* mem, batch_bool_constant<int64_t, A, Values...>, convert<int64_t>, Mode, requires_arch<A>) noexcept
+        {
+            const auto d = load_masked<A>(reinterpret_cast<const double*>(mem), batch_bool_constant<double, A, Values...> {}, convert<double> {}, Mode {}, A {});
+            return bitwise_cast<int64_t>(d);
+        }
+
+        template <class A, bool... Values, class Mode>
+        XSIMD_INLINE typename std::enable_if<types::has_simd_register<double, A>::value, batch<uint64_t, A>>::type
+        load_masked(uint64_t const* mem, batch_bool_constant<uint64_t, A, Values...>, convert<uint64_t>, Mode, requires_arch<A>) noexcept
+        {
+            const auto d = load_masked<A>(reinterpret_cast<const double*>(mem), batch_bool_constant<double, A, Values...> {}, convert<double> {}, Mode {}, A {});
+            return bitwise_cast<uint64_t>(d);
+        }
+
+        template <class A, bool... Values, class Mode>
+        XSIMD_INLINE void store_masked(int32_t* mem, batch<int32_t, A> const& src, batch_bool_constant<int32_t, A, Values...>, Mode, requires_arch<A>) noexcept
+        {
+            store_masked<A>(reinterpret_cast<float*>(mem), bitwise_cast<float>(src), batch_bool_constant<float, A, Values...> {}, Mode {}, A {});
+        }
+
+        template <class A, bool... Values, class Mode>
+        XSIMD_INLINE void store_masked(uint32_t* mem, batch<uint32_t, A> const& src, batch_bool_constant<uint32_t, A, Values...>, Mode, requires_arch<A>) noexcept
+        {
+            store_masked<A>(reinterpret_cast<float*>(mem), bitwise_cast<float>(src), batch_bool_constant<float, A, Values...> {}, Mode {}, A {});
+        }
+
+        template <class A, bool... Values, class Mode>
+        XSIMD_INLINE typename std::enable_if<types::has_simd_register<double, A>::value, void>::type
+        store_masked(int64_t* mem, batch<int64_t, A> const& src, batch_bool_constant<int64_t, A, Values...>, Mode, requires_arch<A>) noexcept
+        {
+            store_masked<A>(reinterpret_cast<double*>(mem), bitwise_cast<double>(src), batch_bool_constant<double, A, Values...> {}, Mode {}, A {});
+        }
+
+        template <class A, bool... Values, class Mode>
+        XSIMD_INLINE typename std::enable_if<types::has_simd_register<double, A>::value, void>::type
+        store_masked(uint64_t* mem, batch<uint64_t, A> const& src, batch_bool_constant<uint64_t, A, Values...>, Mode, requires_arch<A>) noexcept
+        {
+            store_masked<A>(reinterpret_cast<double*>(mem), bitwise_cast<double>(src), batch_bool_constant<double, A, Values...> {}, Mode {}, A {});
         }
 
         // rotate_right
--- a/include/xsimd/arch/xsimd_avx.hpp
+++ b/include/xsimd/arch/xsimd_avx.hpp
@@ -18,6 +18,7 @@
 #include <type_traits>
 
 #include "../types/xsimd_avx_register.hpp"
+#include "../types/xsimd_batch_constant.hpp"
 
 namespace xsimd
 {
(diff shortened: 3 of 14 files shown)


serge-sans-paille:
interesting. Before going into the details, what are the memory effects of a masked load in terms of read memory and value stored? Are the masked elements set to 0 ? to an undefined value ? AVX etc seems to set the value to zero?

I'm not sold on the common implementation which looks quite heavy in scalar operation. I can see that we can't do a plain load followed by an and because it could lead to access to unallocated memory. If the mask were constant, we could optimize statically some common patterns, but with a dynamic mask as you propose...

DiamonDinoia:
Some thoughts:
1. ~~Undefined for masked values. Since depending on the operations 0 or 1 might be the correct values. In that case they could use the mask itself to initialize the values. Also, because imagine I want to polulate the even elements from one memory location and the odd from another. Masked loads (I think) are faster than a gather.~~ set to 0 following the x86 convention
2. We could remove the dynamic mask entirely. I added for completeness.
3. We could do a la vcl and have a load partial, store partial where we just optimize for head and tail. I preferred this solution as I'm assuming xsimd users know the performance implications of the API.
4. ~~For now this is fast only on avx, av2 but for sse even if it heavy on scalar it is slow only when reading bytes or short. We could optimize these cases.~~
5. I'm not sure about sve/neon
6. With static masks if the first and the last element are read it is possible to do load+and

In general, I use these operations when I want to vectorize and inner loop that is not a multiple of the simd width. This is a small inner loop nested in a loop executed a lot of times. Depending on the operations, padding sometimes is slower than masking.

DiamonDinoia:
Hi @serge-sans-paille,

This is starting to take shape now. I still need to clean up. But this is a first working implementation.

I extended the mask operation in batch_bool to mimic std::bit so that i could reuse the code for all masked operations. I feel having those living outside the class in detail:: is not the best. They are also useful to the user I think.

DiamonDinoia:
@serge-sans-paille do you have any clue on why mingw32 fails? 

I need to do a final pass make the code coherent (I am mixing reinterpret_casts and c-style casts). Affter that I think will ready for review.

include/xsimd/arch/common/xsimd_common_memory.hpp
```diff
@@ -348,6 +348,153 @@ namespace xsimd
...
+                    return batch<T_out, A>::load(buffer.data(), aligned_mode {});
+                }
+                else
+                {
+                    return load_masked<A>(mem, mask.as_batch_bool(), convert<T_out> {}, common {});
+                }
+            }
+            else
+            {
+                // Fallback to runtime path for non prefix/suffix masks
```
serge-sans-paille: Unless the compiler optimizes those out, the dynamic load is going to perform serveral checks that are going to be always false (none / all).
I'm still not convinced by the relevancy of dynamic mask support. Maybe we could start small with "just" the static mask support and eventually add a dynamic version later if it appears to be needed? But maybe you have specific case in mind ?

DiamonDinoia: It is just that in my mind I can offer a public runtime mask API
then the compile time mask optimizes as much as it can and if there is no alternative it calls the runtime mask.

Same as the swizzle. So I was not planning of optimizing the runtime at all we just get a maskl_load/store where possible or a common since sse does not support masking. 

PS:
This can be optimized in the future with using avx on sse or avx512vl but I will open a discussion once this is merged

DiamonDinoia: I am cleaning this to only offer compile time for now and the runtime can be a separate PR

DiamonDinoia:
I don't think the error is due to anything related to this PR. But, I like the looks of it now. Ready to review!

include/xsimd/arch/common/xsimd_common_memory.hpp
```diff
@@ -348,6 +349,102 @@ namespace xsimd
...
+        template <class A, class T_in, class T_out, bool... Values, class alignment>
+        XSIMD_INLINE batch<T_out, A>
+        load_masked(T_in const* mem, batch_bool_constant<T_out, A, Values...>, convert<T_out>, alignment, requires_arch<common>) noexcept
+        {
+            constexpr std::size_t size = batch<T_out, A>::size;
+            alignas(A::alignment()) std::array<T_out, size> buffer {};
+            constexpr std::array<bool, size> mask { Values... };
+
+            for (std::size_t i = 0; i < size; ++i)
+                buffer[i] = mask[i] ? static_cast<T_out>(mem[i]) : T_out(0);
```
serge-sans-paille, line 366: question: de we want a defined value in non-masked slots, or an undefined value? I guess it depends on how the various arch behave...

include/xsimd/arch/common/xsimd_common_memory.hpp
```diff
@@ -348,6 +349,102 @@ namespace xsimd
...
+            return load_unaligned<A>(mem, convert<T> {}, A {});
+        }
+
+        template <class A, class T_in, class T_out, bool... Values, class alignment>
+        XSIMD_INLINE batch<T_out, A>
+        load_masked(T_in const* mem, batch_bool_constant<T_out, A, Values...>, convert<T_out>, alignment, requires_arch<common>) noexcept
+        {
+            constexpr std::size_t size = batch<T_out, A>::size;
+            alignas(A::alignment()) std::array<T_out, size> buffer {};
+            constexpr std::array<bool, size> mask { Values... };
```
serge-sans-paille: It seems to me that `operator[]` is not constexpr for std::array in C++11. let's use a plain C array then?

include/xsimd/arch/common/xsimd_common_memory.hpp
serge-sans-paille: neon does not support masked load natively anyway, I guess zeroing is ok.

include/xsimd/arch/xsimd_avx.hpp
```diff
@@ -63,6 +78,17 @@ namespace xsimd
             {
                 return _mm256_insertf128_pd(_mm256_castpd128_pd256(low), high, 1);
             }
+            template <class T>
+            XSIMD_INLINE batch<T, sse4_2> lower_half(batch<T, avx> const& self) noexcept
```
serge-sans-paille: this seems recursive. don't you want to use `self.data` ?

include/xsimd/types/xsimd_batch_constant.hpp
```diff
@@ -117,8 +165,84 @@ namespace xsimd
...
+        {
+            static_assert(Begin <= End, "splice: Begin must be <= End");
+            static_assert(End <= size, "splice: End must be <= size");
+            static_assert(N == batch_bool<T, A2>::size, "splice: target arch size must match submask length");
+            return splice_impl<A2, Begin>(detail::make_index_sequence<N>());
+        }
+
+        // Convenience helpers for half splits (require even size and appropriate target arch)
+        template <class A2>
+        static constexpr auto lower_half() noexcept
```
serge-sans-paille: I'm not a big fan of this, batch_constant size are tied to an architecture...

serge-sans-paille: Maybe we could put some of the function below as free-functions somewhere?

include/xsimd/arch/common/xsimd_common_memory.hpp
DiamonDinoia: AFAIK x86 defaults to 0 so I kept it consistent.

include/xsimd/arch/common/xsimd_common_memory.hpp
DiamonDinoia: Yes, I'll replace it. It is `constexpr` from 14/17

include/xsimd/types/xsimd_batch_constant.hpp
DiamonDinoia: I am not against moving splice outside the class. I implemented it here because it avoid passing the current architecture. It can go in xsimd::common somewhere maybe not common memory as swizzle or other APIs might use this in the future.

serge-sans-paille: It could stay in the same file, just as free functions?

serge-sans-paille:
@DiamonDinoia we plan to do a release in the forthcoming weeks. Do you want this PR to be part of it?

DiamonDinoia:
> @DiamonDinoia we plan to do a release in the forthcoming weeks. Do you want this PR to be part of it?

Only if it does not delay the release by much. I would like to have a release by mid-November. This gives me two weeks to update my downstream projects. For an end of the year release. 


PS: I am planning on addressing the swizzle issue after this PR.

JohanMabille:
> I would like to have a release by mid-November

That is more or less what we planned ;)

include/xsimd/arch/common/xsimd_common_arithmetic.hpp
```diff
@@ -16,6 +16,7 @@
 #include <limits>
 #include <type_traits>
 
+#include "../../types/xsimd_batch_constant.hpp"
```
serge-sans-paille, line 19: Why do you need that change?

include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -116,6 +117,160 @@ namespace xsimd
...
+            XSIMD_IF_CONSTEXPR(mask.none())
+            {
+                return _mm256_setzero_si256();
+            }
+            else XSIMD_IF_CONSTEXPR(mask.all())
+            {
+                return load<A>(mem, Mode {});
+            }
+            // confined to lower 128-bit half (4 lanes) → forward to SSE
+            else XSIMD_IF_CONSTEXPR(mask.countl_zero() >= 4)
```
serge-sans-paille: Thinking of it, we could remove the need for `countl_zero` by just masking, something like

```
if((mask.mask() & 0xF0) == 0)
```

which would be a decent implementation for
```
if(mask.lower_half() == 0) // or lower_mask ?
```
which has the nice property of not encoding any "magical value"

include/xsimd/arch/xsimd_avx.hpp
```diff
@@ -865,6 +881,134 @@ namespace xsimd
...
+            {
+                constexpr auto mlo = ::xsimd::detail::lower_half<sse2>(mask);
+                const batch<double, sse2> lo(_mm256_castpd256_pd128(src));
+                store_masked<sse2>(mem, lo, mlo, Mode {}, sse4_2 {});
+            }
+            // confined to upper 128-bit half (2 lanes) → forward to SSE2
+            else XSIMD_IF_CONSTEXPR(mask.countr_zero() >= 2)
+            {
+                constexpr auto mhi = ::xsimd::detail::upper_half<sse2>(mask);
+                const batch<double, sse2> hi(_mm256_extractf128_pd(src, 1));
```
serge-sans-paille: that's a upper_half, right?

include/xsimd/arch/xsimd_avx.hpp
```diff
@@ -865,6 +881,134 @@ namespace xsimd
...
+            }
+            else XSIMD_IF_CONSTEXPR(mask.all())
+            {
+                src.store(mem, Mode {});
+            }
+            // confined to lower 128-bit half (2 lanes) → forward to SSE2
+            else XSIMD_IF_CONSTEXPR(mask.countl_zero() >= 2)
+            {
+                constexpr auto mlo = ::xsimd::detail::lower_half<sse2>(mask);
+                const batch<double, sse2> lo(_mm256_castpd256_pd128(src));
```
serge-sans-paille: that's a lower_half, right?

include/xsimd/arch/xsimd_avx.hpp
```diff
@@ -865,6 +881,134 @@ namespace xsimd
...
+            {
+                constexpr auto mhi = ::xsimd::detail::upper_half<sse2>(mask);
+                const batch<double, sse2> hi(_mm256_extractf128_pd(src, 1));
+                store_masked<sse2>(mem + 2, hi, mhi, Mode {}, sse4_2 {});
+            }
+            else
+            {
+                _mm256_maskstore_pd(mem, mask.as_batch(), src);
+            }
+        }
```
serge-sans-paille, line 999: epiphany : if you express everything in terms of lower / upper half, you're pretty close to a type-generic implementation that would avoid a lot of redunduncy. just need to wrap the actual `_mm256_maskstore*` and we're good, right?

include/xsimd/arch/xsimd_avx512f.hpp
```diff
@@ -1391,6 +1493,238 @@ namespace xsimd
             return _mm512_loadu_pd(mem);
         }
 
+        // load_masked
```
serge-sans-paille: why did you split the load_masked in two parts of the header? The generic epiphany for avx seems to apply here too :-)

serge-sans-paille (commented):
A new wave of comments. I have the feeling that we could simplify the implementation a lot by making it more generic.

include/xsimd/arch/xsimd_avx.hpp
DiamonDinoia: I guess you are right. It might be possible to express things in terms of size and halves. This might be recursive though and we might need to arch to half arch, similarly to what happens in widen.

DiamonDinoia: I'll also have to think if the size of the element matters. I'll have a look next week

include/xsimd/arch/xsimd_avx.hpp
```diff
@@ -865,6 +881,134 @@ namespace xsimd
...
+            else XSIMD_IF_CONSTEXPR(mask.all())
+            {
+                return load<A>(mem, Mode {});
+            }
+            // confined to lower 128-bit half (4 lanes) → forward to SSE2
+            else XSIMD_IF_CONSTEXPR(mask.countl_zero() >= 4)
+            {
+                constexpr auto mlo = ::xsimd::detail::lower_half<sse4_2>(mask);
+                const auto lo = load_masked(mem, mlo, convert<float> {}, Mode {}, sse4_2 {});
+                return batch<float, A>(detail::merge_sse(lo, batch<float, sse4_2>(0.f)));
```
serge-sans-paille: could be a call to _mm256_castps128_ps256  instead of merge + batch of zero

include/xsimd/arch/xsimd_avx.hpp
```diff
@@ -865,6 +881,134 @@ namespace xsimd
...
+            }
+            // confined to upper 128-bit half (4 lanes) → forward to SSE2
+            else XSIMD_IF_CONSTEXPR(mask.countr_zero() >= 4)
+            {
+                constexpr auto mhi = ::xsimd::detail::upper_half<sse4_2>(mask);
+                const auto hi = load_masked(mem + 4, mhi, convert<float> {}, Mode {}, sse4_2 {});
+                return batch<float, A>(detail::merge_sse(batch<float, sse4_2>(0.f), hi));
```
serge-sans-paille: I don't know if we could use a similar trick to `_mm256_castps128_ps256 ` here

serge-sans-paille:
@AntoinePrv once you update that branch we can do a release \o/
EDIT: I meant @DiamonDinoia of course :-)

DiamonDinoia:
> @AntoinePrv once you update that branch we can do a release \o/

I will work on this asap. I just have something to deliver on another project this week so it might take a bit for me to get back to this.

serge-sans-paille:
Note that once you're done with Intel support, i still have to implement ARM et cie when possible

include/xsimd/arch/common/xsimd_common_arithmetic.hpp
DiamonDinoia: So I removed some includes form xsimd_common_fwd and replaced them with forward declarations. So know the concrete types needs to be included in some of the other header.

This speed up compilation a bit and I thought that fwd should really contain only forward declarations.

include/xsimd/arch/xsimd_avx2.hpp
DiamonDinoia: The reason why I went for countl_zero is to maintain parity with std::bit. I feel like this is public API so it is worth for users not having to learn a new naming scheme

include/xsimd/arch/xsimd_avx.hpp
DiamonDinoia: The reason why I kept the implementation separate is as follows:

The general implementation in terms of upper/lower ais possible. This haves the code in avx, avx2 still has to provide a new implementation so that it can take advantage of the new intrinsics so we for from 2 versions in avx down to 1 and 2 in avx2 down to 1. Same for avx512.


But then for merging we can not use more optimized intrinsics than _mm256_castps128_ps256. Is it worth it?

The solution here is to pass the MaskLoad/Store and the merge function as templates to this general implementation. Then we could have 1 general implementation maybe for all the kernels. But, I think it might be too much templating. 

In general also wrapping _mm256_maskstore* also makes a clean way to expose this to the used if we wish to do so. 


What do you suggest?

DiamonDinoia:
I removed most of the duplication at least I stopped when I thought things were getting out of hand. I think in the future we might offer direct access to masked load with dynamic mask. Which is basically exposing the internal helper functions but adapting them to the xsimd convention.

include/xsimd/arch/xsimd_avx512f.hpp
```diff
@@ -872,11 +1010,11 @@ namespace xsimd
 
             XSIMD_IF_CONSTEXPR(sizeof(T) == 4)
             {
-                return _mm512_mask_sub_epi32(self, mask.data, self, _mm512_set1_epi32(1));
+                return _mm512_mask_sub_epi32(self, mask, self, _mm512_set1_epi32(1));
```
DiamonDinoia: this sneaked in. I'll revert. My OCD clicked in...

serge-sans-paille:
Merged as 239fbbd96dda5bb48a6c9a205e93836a724de351
Thanks a lot @DiamonDinoia for the hard work. It's now on me to implement part of it for other architectures.

include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -116,6 +117,137 @@ namespace xsimd
...
+                // Use the raw register-level maskload helpers for the remaining cases.
+                return detail::maskload(reinterpret_cast<const int_t*>(mem), mask.as_batch());
+            }
+        }
+
+        template <class A, bool... Values, class Mode>
+        XSIMD_INLINE batch<int32_t, A> load_masked(int32_t const* mem, batch_bool_constant<int32_t, A, Values...> mask, convert<int32_t>, Mode, requires_arch<avx2>) noexcept
+        {
+            return load_masked<A>(mem, mask, convert<int32_t> {}, Mode {}, avx2 {});
+        }
```
AntoinePrv, line 166: @DiamonDinoia I am having trouble with this function above. Is this an infinite loop?

DiamonDinoia: I might have meant calling uint -> int32 there.

DiamonDinoia: I am currently travelling. I have a look once I am back.

PR #1163, opened 2025-08-23
generalization of reduce_mul implementation

serge-sans-paille:


serge-sans-paille:
Note to self: we should provide dedicated reduction for `or`,  `and` and  `xor` operator, and do it in a generic way if possible. RISCV supports those reductions.

include/xsimd/arch/common/xsimd_common_arithmetic.hpp
```diff
@@ -139,20 +139,6 @@ namespace xsimd
...
-        {
-            alignas(A::alignment()) T buffer[batch<T, A>::size];
-            self.store_aligned(buffer);
-            T res = 0;
-            for (T val : buffer)
-            {
-                res += val;
-            }
-            return res;
-        }
```
JohanMabille, line 154: Technically, this means the next release should be a major one. On the other hand, since `reduce_add`  was already available, I wonder if this function was really used.

JohanMabille (approved):

serge-sans-paille:
On Fri, Aug 29, 2025 at 08:45:48AM -0700, Johan Mabille wrote:
> @JohanMabille approved this pull request.
> 
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> 
> In [1]include/xsimd/arch/common/xsimd_common_arithmetic.hpp:
> 
> > @@ -139,20 +139,6 @@ namespace xsimd
>              return fma(x, y, select(mask, neg(z), z));
>          }
> 
> -        // hadd
> -        template <class A, class T, class /*=typename std::enable_if<std::is_integral<T>::value, void>::type*/>
> -        XSIMD_INLINE T hadd(batch<T, A> const& self, requires_arch<common>) noexcept
> -        {
> -            alignas(A::alignment()) T buffer[batch<T, A>::size];
> -            self.store_aligned(buffer);
> -            T res = 0;
> -            for (T val : buffer)
> -            {
> -                res += val;
> -            }
> -            return res;
> -        }
> 

well, it's not part of the standard API, is it?

JohanMabille:
Ah indeed, so we're good

PR #1164, opened 2025-08-29
NEW API: Fixed shift, Rotation

DiamonDinoia:
1. adds the API bitwise_[l|r]shift<N>(...) and rot[l|r]<N>(...)
2. updates the test to use the API
3. fixes documentation

I implemente xoshiro (vecotrized) the current API looks like this:
```cpp
    const auto result = xsimd::rotl(m_state[0] + m_state[3], 23) + m_state[0];
    const auto t = m_state[1] << 17;

    m_state[2] ^= m_state[0];
    m_state[3] ^= m_state[1];
    m_state[2] ^= t;

    m_state[3] = xsimd::rotl(m_state[3], 45);

    return result;
```
and with AVX2 achieves `393.06 M samples/s`
The new API is:
```cpp
    const auto result = xsimd::rotl<23>(m_state[0] + m_state[3]) + m_state[0];
    const auto t = xsimd::bitwise_lshift<17>(m_state[1]);

    m_state[2] ^= m_state[0];
    m_state[3] ^= m_state[1];
    m_state[1] ^= m_state[2];
    m_state[0] ^= m_state[3];

    m_state[2] ^= t;

    m_state[3] = xsimd::rotl<45>(m_state[3]);

    return result;
```
and with AVX2 achieves `423.76 M samples/s`

With this change causing a 7% speed increase, xoshiro is the fastest RNG among the ones I tested. Othwewise, PCG64 beats it.

### DIFF
--- a/docs/source/api/bitwise_operators_index.rst
+++ b/docs/source/api/bitwise_operators_index.rst
@@ -40,9 +40,9 @@ Bitwise operators
 +---------------------------------------+----------------------------------------------------+
 | :cpp:func:`bitwise_andnot`            | per slot bitwise and not                           |
 +---------------------------------------+----------------------------------------------------+
-| :cpp:func:`bitwise_lshift`            | per slot bitwise and                               |
+| :cpp:func:`bitwise_lshift`            | per slot bitwise left shift                        |
 +---------------------------------------+----------------------------------------------------+
-| :cpp:func:`bitwise_rshift`            | per slot bitwise and not                           |
+| :cpp:func:`bitwise_rshift`            | per slot bitwise right shift                       |
 +---------------------------------------+----------------------------------------------------+
 | :cpp:func:`rotr`                      | per slot rotate right                              |
 +---------------------------------------+----------------------------------------------------+
--- a/include/xsimd/arch/common/xsimd_common_arithmetic.hpp
+++ b/include/xsimd/arch/common/xsimd_common_arithmetic.hpp
@@ -34,6 +34,13 @@ namespace xsimd
                                  { return x << y; },
                                  self, other);
         }
+        template <size_t shift, class A, class T, class /*=typename std::enable_if<std::is_integral<T>::value, void>::type*/>
+        XSIMD_INLINE batch<T, A> bitwise_lshift(batch<T, A> const& self, requires_arch<common>) noexcept
+        {
+            constexpr auto bits = std::numeric_limits<T>::digits + std::numeric_limits<T>::is_signed;
+            static_assert(shift < bits, "Shift must be less than the number of bits in T");
+            return bitwise_lshift(self, shift, A {});
+        }
 
         // bitwise_rshift
         template <class A, class T, class /*=typename std::enable_if<std::is_integral<T>::value, void>::type*/>
@@ -42,6 +49,13 @@ namespace xsimd
             return detail::apply([](T x, T y) noexcept
                                  { return x >> y; },
                                  self, other);
+        }
+        template <size_t shift, class A, class T, class /*=typename std::enable_if<std::is_integral<T>::value, void>::type*/>
+        XSIMD_INLINE batch<T, A> bitwise_rshift(batch<T, A> const& self, requires_arch<common>) noexcept
+        {
+            constexpr auto bits = std::numeric_limits<T>::digits + std::numeric_limits<T>::is_signed;
+            static_assert(shift < bits, "Shift must be less than the number of bits in T");
+            return bitwise_rshift(self, shift, A {});
         }
 
         // decr
@@ -166,16 +180,30 @@ namespace xsimd
         template <class A, class T, class STy>
         XSIMD_INLINE batch<T, A> rotl(batch<T, A> const& self, STy other, requires_arch<common>) noexcept
         {
-            constexpr auto N = std::numeric_limits<T>::digits;
-            return (self << other) | (self >> (N - other));
+            constexpr auto bits = std::numeric_limits<T>::digits + std::numeric_limits<T>::is_signed;
+            return (self << other) | (self >> (bits - other));
+        }
+        template <size_t count, class A, class T>
+        XSIMD_INLINE batch<T, A> rotl(batch<T, A> const& self, requires_arch<common>) noexcept
+        {
+            constexpr auto bits = std::numeric_limits<T>::digits + std::numeric_limits<T>::is_signed;
+            static_assert(count < bits, "Count amount must be less than the number of bits in T");
+            return bitwise_lshift<count>(self) | bitwise_rshift<bits - count>(self);
         }
 
         // rotr
         template <class A, class T, class STy>
         XSIMD_INLINE batch<T, A> rotr(batch<T, A> const& self, STy other, requires_arch<common>) noexcept
         {
-            constexpr auto N = std::numeric_limits<T>::digits;
-            return (self >> other) | (self << (N - other));
+            constexpr auto bits = std::numeric_limits<T>::digits + std::numeric_limits<T>::is_signed;
+            return (self >> other) | (self << (bits - other));
+        }
+        template <size_t count, class A, class T>
+        XSIMD_INLINE batch<T, A> rotr(batch<T, A> const& self, requires_arch<common>) noexcept
+        {
+            constexpr auto bits = std::numeric_limits<T>::digits + std::numeric_limits<T>::is_signed;
+            static_assert(count < bits, "Count must be less than the number of bits in T");
+            return bitwise_rshift<count>(self) | bitwise_lshift<bits - count>(self);
         }
 
         // sadd
--- a/include/xsimd/arch/xsimd_avx2.hpp
+++ b/include/xsimd/arch/xsimd_avx2.hpp
@@ -17,6 +17,8 @@
 
 #include "../types/xsimd_avx2_register.hpp"
 
+#include <limits>
+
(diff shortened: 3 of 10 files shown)


serge-sans-paille:
We already have a similar approach for `rotate_left` and `rotate_right`, `slide_left` and `slide_right`, so that looks good.
We also have `insert` that has a different API:

```
    template <class T, class A, size_t I>
    XSIMD_INLINE batch<T, A> insert(batch<T, A> const& x, T val, index<I> pos) noexcept
```

which makes me wonder if it's a good idea.

Anyway I like your idea, further comments inline.

include/xsimd/arch/common/xsimd_common_arithmetic.hpp
```diff
@@ -34,6 +34,11 @@ namespace xsimd
                                  { return x << y; },
                                  self, other);
         }
+        template <int shift, class A, class T, class /*=typename std::enable_if<std::is_integral<T>::value, void>::type*/>
```
serge-sans-paille: can the shift be negative? If not we should use an unsigned type for `I`.

include/xsimd/arch/common/xsimd_common_arithmetic.hpp
```diff
@@ -183,6 +193,12 @@ namespace xsimd
             constexpr auto N = std::numeric_limits<T>::digits;
             return (self << other) | (self >> (N - other));
         }
+        template <int count, class A, class T>
```
serge-sans-paille: same here, we tend to use `size_t` for this kind of parameter.

include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -172,17 +172,40 @@ namespace xsimd
...
+        }
+
         template <class A, class T, class = typename std::enable_if<std::is_integral<T>::value, void>::type>
         XSIMD_INLINE batch<T, A> bitwise_lshift(batch<T, A> const& self, batch<T, A> const& other, requires_arch<avx2>) noexcept
         {
             XSIMD_IF_CONSTEXPR(sizeof(T) == 4)
             {
                 return _mm256_sllv_epi32(self, other);
             }
+#if XSIMD_WITH_AVX512VL
```
serge-sans-paille: I'm not totally fine with this kind of macro guard. the instruction used should depend on the batch architecture (`A`) and not macros. We could have someone wanting an avx2 kernel within an avx512 context, and I have not yet thought about how to handle instruction sets that extend previous generation instructions. Could you postpone that particular change?

Not saying it's incorrect, It's just that I need to think a bit about the pattern.

include/xsimd/arch/xsimd_avx512bw.hpp
```diff
@@ -223,7 +223,6 @@ namespace xsimd
                 }
             }
         }
-
```
serge-sans-paille: ?

include/xsimd/config/xsimd_config.hpp
```diff
@@ -244,6 +244,17 @@
 #define XSIMD_WITH_AVX512DQ 0
 #endif
 
+/**
```
serge-sans-paille: should be in an independent commit, preferably with a test case that catches the scenario we missed.

serge-sans-paille (changes requested):
Could you also add sse2's `_mm_slli_epi32` and `_mm_srai_epi16`?
You provided a generic implementation so ,other architectures shouldn't be impacted, but I'll still have a look and check if some specific instructions are available.

include/xsimd/arch/xsimd_avx512bw.hpp
DiamonDinoia: My mistake, spurious change. I added an API there and then deleted.

include/xsimd/arch/xsimd_avx2.hpp
DiamonDinoia: I was thinking of using SSEVL, AVXVL as new archs & register types. Then from AVX512 we forward to those. We also change make_sized_batch_t to return those instead of plain sse,avx. This is useful to me as I use custom sized batches even with AVX512 to avoid padding. It should make things faster expecially in sse as it remove vzeroupper if I am not mistaken.

include/xsimd/arch/common/xsimd_common_arithmetic.hpp
```diff
@@ -43,6 +49,12 @@ namespace xsimd
                                  { return x >> y; },
                                  self, other);
         }
+        template <size_t shift, class A, class T, class /*=typename std::enable_if<std::is_integral<T>::value, void>::type*/>
+        XSIMD_INLINE batch<T, A> bitwise_rshift(batch<T, A> const& self, requires_arch<common>) noexcept
+        {
+            static_assert(shift < std::numeric_limits<T>::digits, "shift must be less than the number of bits in T");
```
serge-sans-paille: `digits` does not count the sign bit, so you should probably add `std::is_signed<T>` to the condition.

docs/source/api/bitwise_operators_index.rst
```diff
@@ -40,9 +40,9 @@ Bitwise operators
 +---------------------------------------+----------------------------------------------------+
 | :cpp:func:`bitwise_andnot`            | per slot bitwise and not                           |
 +---------------------------------------+----------------------------------------------------+
-| :cpp:func:`bitwise_lshift`            | per slot bitwise and                               |
+| :cpp:func:`bitwise_lshift`            | per slot bitwise left shift                        |
```
serge-sans-paille, line 43: ^^!

include/xsimd/arch/common/xsimd_common_arithmetic.hpp
```diff
@@ -169,6 +181,13 @@ namespace xsimd
             constexpr auto N = std::numeric_limits<T>::digits;
             return (self << other) | (self >> (N - other));
         }
+        template <size_t count, class A, class T>
+        XSIMD_INLINE batch<T, A> rotl(batch<T, A> const& self, requires_arch<common>) noexcept
+        {
+            static_assert(count < std::numeric_limits<T>::digits, "count must be less than the number of bits in T");
```
serge-sans-paille: same here

include/xsimd/arch/common/xsimd_common_arithmetic.hpp
```diff
@@ -177,6 +196,13 @@ namespace xsimd
             constexpr auto N = std::numeric_limits<T>::digits;
             return (self >> other) | (self << (N - other));
         }
+        template <size_t count, class A, class T>
+        XSIMD_INLINE batch<T, A> rotr(batch<T, A> const& self, requires_arch<common>) noexcept
+        {
+            static_assert(count < std::numeric_limits<T>::digits, "count must be less than the number of bits in T");
+            constexpr auto N = std::numeric_limits<T>::digits;
```
serge-sans-paille: same here, and you could reuse `N` in the `static_assert`

include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -172,17 +174,35 @@ namespace xsimd
...
         template <class A, class T, class = typename std::enable_if<std::is_integral<T>::value, void>::type>
         XSIMD_INLINE batch<T, A> bitwise_lshift(batch<T, A> const& self, batch<T, A> const& other, requires_arch<avx2>) noexcept
         {
             XSIMD_IF_CONSTEXPR(sizeof(T) == 4)
             {
                 return _mm256_sllv_epi32(self, other);
             }
-            else XSIMD_IF_CONSTEXPR(sizeof(T) == 8)
-            {
-                return _mm256_sllv_epi64(self, other);
```
serge-sans-paille: why did you remove this one?

include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -272,10 +343,6 @@ namespace xsimd
                 {
                     return _mm256_srlv_epi32(self, other);
                 }
-                else XSIMD_IF_CONSTEXPR(sizeof(T) == 8)
```
serge-sans-paille: same question here.

include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -252,6 +272,57 @@ namespace xsimd
...
+                else XSIMD_IF_CONSTEXPR(sizeof(T) == 4)
+                {
+                    return _mm256_srai_epi32(self, shift);
+                }
+                else
+                {
+                    return bitwise_rshift<shift>(self, avx {});
+                }
+            }
+            else
```
serge-sans-paille, line 311: why not adapting the solution from the `is_signed` branch when `sizeof(T) == 1`?

include/xsimd/arch/xsimd_avx2.hpp
DiamonDinoia: that was my bad. I was convinced these where AVX512VL extentions added by me. Reverted.

DiamonDinoia:
I touched the rotl rotr implementation and tests in general because accounting for the sign bit `std::numeric_limits<T>::digits + std::numeric_limits<T>::is_signed` broke some tests. 

I then renamed N to bits for clarity.

It seems to me that this was the right way to go about it but I am not 100% sure on it. In my mind rotation on signed types makes no sense.

serge-sans-paille:
That's a great addition! I'll have a look at other instruction sets ;-)

PR #1165, opened 2025-09-08
Swap arguments to _mm*_{min,max}_ps

kalenedrael:
`std::min` and `std::max` return their second argument if either argument is NAN, while the intrinsics return the first argument. This brings `xsimd::{min,max}` in line with `std::{min,max}`, at least on x86.

### DIFF
--- a/test/test_xsimd_api.cpp
+++ b/test/test_xsimd_api.cpp
@@ -658,6 +658,29 @@ struct xsimd_api_float_types_functions
         value_type val(0);
         CHECK_EQ(extract(xsimd::log1p(T(val))), std::log1p(val));
     }
+
+    void test_max_nan()
+    {
+        value_type val0(2.7818);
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
Thanks! Could you add a test case for this?

kalenedrael:
Added a test case. I'm not entirely clear on how the architecture-specific tests work though, so let me know if something else needs to be added.

serge-sans-paille:
> Added a test case. I'm not entirely clear on how the architecture-specific tests work though, so let me know if something else needs to be added.

That's fine, thanks. I'll do the portage to other architectures if needs be!

kalenedrael:
Correction for potential future readers: The first comment has it backwards - `std::min` and `std::max` return their _first_ argument if either argument is NAN, while the intrinsics return their _second_ argument.

PR #1166, opened 2025-09-10
Activate min/max testing wrt. NaN on all architectures

serge-sans-paille:


serge-sans-paille:
Giving up with this one.

PR #1172, opened 2025-09-28
Add `{load,store}_unaligned` for `batch_bool` on x86

kalenedrael:
These work around GCC not being able to optimize the baseline implementations to SIMD operations, and compilers in general not being able to know that the representation of `batch_bool` is a bitmask of all 0 or all 1, thus integer 0 or -1.

Also added some more robust tests for bool load/store that test for bitwise correctness (either 0 or 1).


kalenedrael:
Currently wresting with the 12905790283 flavors and non-orthogonality of x86 SIMD instructions...

serge-sans-paille:
I like this change, thanks!

Issue #1143, opened 2025-07-10
xsimd no supported architecture


DiamonDinoia:
Hi team,

[this issue](https://github.com/flatironinstitute/finufft/issues/711#issuecomment-3056139068) was brought to my attention.

Can I propose to keep the function included and return an emulated register with the size of an sse register? Maybe xsimd does not support the architecture but the compiler can still vectorize using MMX, AltiVect which are not directly supported by xsimd.

Otherwise, I need to wrap the code in ifdefs to fallback on scalar in this case.

Thanks,
Marco

serge-sans-paille:
Would `-DXSIMD_WITH_EMULATED=1 -DXSIMD_DEFAULT_ARCH=emulated\<128\>` do the trick?

DiamonDinoia:
I came up with this:
```cpp
#ifndef FINUFFT_XSIMD_HPP
#define FINUFFT_XSIMD_HPP

#include <xsimd/config/xsimd_config.hpp>

#ifdef XSIMD_NO_SUPPORTED_ARCHITECTURE
#undef XSIMD_NO_SUPPORTED_ARCHITECTURE
#define XSIMD_WITH_EMULATED 1
#define XSIMD_DEFAULT_ARCH  emulated<128>
#endif

#include <xsimd/xsimd.hpp>

#endif // FINUFFT_XSIMD_HPP
```

PR #1178, opened 2025-10-12
Make xsimd_scalar.hpp independent from xsimd::batch

serge-sans-paille:
This paves the way for cleaner error when trying to instantiate an xsimd::batch when to architecture is available.

Related to #1143

### DIFF
--- a/include/xsimd/xsimd.hpp
+++ b/include/xsimd/xsimd.hpp
@@ -63,7 +63,16 @@
 #include "memory/xsimd_aligned_allocator.hpp"
 
 #if defined(XSIMD_NO_SUPPORTED_ARCHITECTURE)
-// to type definition or anything appart from scalar definition and aligned allocator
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
@DiamonDinoia / @JohanMabille this patch changes the behavior when no arch is supported. Previous behavior was "lots of meaningless error". After the patch of this patchset, the error becomes a type error "xsimd::batch does not exist" and after the second patch of this patcheset, we have a proper static assert.
I'm satisified with that behavior but I'd be happy to read your opinion on this before merging.

DiamonDinoia:
Hi @serge-sans-paille I think this is a good idea. 

Do you think my workaround https://github.com/xtensor-stack/xsimd/issues/1143#issuecomment-3206432187 is going to break due to the static_assert?

serge-sans-paille:
@DiamonDinoia given the include order and the change I've made, I don't think it will break: you'lle never end up with the fallback batch type

PR #1180, opened 2025-10-27
Fixes XSIMD_INLINE for compilers that don't have always_inline

DiamonDinoia:
Small fix for compilers (and linters) that do not have the attribute always_inline. This avoids a cascade of false positives.

Clion nova seems to have a problem with this so many users might be affected: https://youtrack.jetbrains.com/issue/CPP-47004/hasattributealwaysinline-is-not-handled-properly


JohanMabille (approved):

JohanMabille:
Thanks!

PR #1184, opened 2025-10-29
Extend support of batch_cast<...> to upcasting to a type twice as big

### LINKED ISSUE #1179: `batch_cast` does not use SIMD intrinsic when available on Intel
Plenty of conversion are [available on Intel](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#itext=cvt&techs=SSE_ALL).
For instance
- Upcast integers: `_mm_cvtepi32_epi64` on SSE `i32` -> `i64` 
- Convert ints to floats: `_mm_cvtepi32_ps` on SEE `i32` -> `f32`
- Downcasting has [more limited](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_pack&techs=SSE_ALL,AVX_ALL) support and is saturating (contrary to `static_cast`).

I'm not sure about other architectures, I believe arm does not have it. Still this could be a nice boost.

serge-sans-paille:
Fix #1179


serge-sans-paille:
@AntoinePrv : this is just some mockup to check if we're good with the API

serge-sans-paille:
Also cc @JohanMabille & @DiamonDinoia  for the API. I didn't want to support the generic conversion, say from uint8 to uint32 with larger arrays, let's start humble

DiamonDinoia:
I would call the API `extension`, `extend` or `widen` because it does not just cast, it zero extend and 1 extend. I think in intrinsics terminology is called widening and convert with widening. 

On the implementation:
1. using `std::conditional` and `std::make_signed` halves the number of struct you need.
~2. #1162 introduces lower/upper :)~

```cpp
namespace {
template<class T> struct widen;

template<> struct widen<uint8_t>  { using type = uint16_t; };
template<> struct widen<uint16_t> { using type = uint32_t; };
template<> struct widen<uint32_t> { using type = uint64_t; };

template<class T>
using widen_t_unsigned = typename widen<T>::type;

} // namespace 

template<class T>
struct widen {
    static_assert(std::is_integral<T>::value && !std::is_same<T,bool>::value,
                  "integral non-bool type required");

    using U     = typename std::make_unsigned<T>::type;
    using UWide = detail::widen_t_unsigned<U>;

    using type = typename std::conditional<
        std::is_signed<T>::value,
        typename std::make_signed<UWide>::type,
        UWide
    >::type;
};

template<class T>
using widen_t = typename widen<T>::type;
```

serge-sans-paille:
@DiamonDinoia yep, widen looks like a good name, thanks!

include/xsimd/types/xsimd_traits.hpp
```diff
@@ -332,6 +332,38 @@ namespace xsimd
...
+        template <>
+        struct widen<uint8_t>
+        {
+            using type = uint16_t;
+        };
+        template <>
+        struct widen<float>
+        {
+            using type = double;
+        };
```
AntoinePrv, line 362: I think a trait that convert a byte size to a `uint`/`float` type could be more generic:

```cpp
template <>
struct sized_uint<4>
{
    using type = uint32_t;
};

...
```

And then use with `batch<sized_uint_t<2*sizeof(T)>>` (or an alias for that).

DiamonDinoia: I am not in favor of this. Then we need `sized_int` and `sized_uint` and the API will need an `std::conditional<std::is_signed<T>, sized_int_t<2*sizeof(T)>, sized_uint_t<2*sizeof(T)>`

DiamonDinoia:
My only suggestion would be to merge masked_memory ops. Or merge lower/upper from it and use it here.

serge-sans-paille:
> My only suggestion would be to merge masked_memory ops. Or merge lower/upper from it and use it here.

Yup, I'll wait for it to be merged!

include/xsimd/arch/common/xsimd_common_cast.hpp
```diff
@@ -0,0 +1,40 @@
...
+            alignas(A::alignment()) T buffer[batch<T, A>::size];
+            x.store_aligned(&buffer[0]);
+
+            using T_out = widen_t<T>;
+            alignas(A::alignment()) T_out out_buffer[batch<T, A>::size];
+            for (size_t i = 0; i < batch<T, A>::size; ++i)
+                out_buffer[i] = static_cast<T_out>(buffer[i]);
+
+            return { batch<T_out, A>::load_aligned(&out_buffer[0]),
+                     batch<T_out, A>::load_aligned(&out_buffer[batch<T_out, A>::size]) };
```
DiamonDinoia, line 33: `out_buffer` 
and `out_buffer + batch<T_out, A>::size` seems clearer to me

serge-sans-paille:
> My only suggestion would be to merge masked_memory ops. Or merge lower/upper from it and use it here.

@DiamonDinoia : Are you fine if I split (haha) that part of your commit in a seperate one so that we don't block each other?

DiamonDinoia:
> > My only suggestion would be to merge masked_memory ops. Or merge lower/upper from it and use it here.
> 
> @DiamonDinoia : Are you fine if I split (haha) that part of your commit in a seperate one so that we don't block each other?

Sure, go ahead. I'll rebase once I have time

include/xsimd/arch/common/xsimd_common_cast.hpp
DiamonDinoia: (I mean doing pointer arithmetic directly instead of using the [] and & operators)

serge-sans-paille:
> My only suggestion would be to merge masked_memory ops. Or merge lower/upper from it and use it here.

done (as in: I created a separate PR to extract that part of your patch, merged it and rebased my PR on top of it)

PR #1185, opened 2025-10-30
Fix/use enable if default

serge-sans-paille:


serge-sans-paille:
@DiamonDinoia the failing test is the one you wanted to have a look at, correct?

DiamonDinoia:
I remember seeing this error. But I did not investigate it. I cannot really remember the context now.

PR #1187, opened 2025-10-31
Fix static_assert on i8mm<neon64>

AntoinePrv:
This failed locally on ``i8mm<neon64>``.
Not sure this is the best way to express it.


serge-sans-paille:
That's totally fine. Could you add validation for in on our github ci, or do you want me to handle it an other commit?

AntoinePrv:
I'm on an Apple M3 here using Clang 20 from conda-forge (so NOT AppleClang).
I'm surprised this was not cached in the CI so not sure where it is coming from.
I have a `-march=native` that shows up in my compilation flags, that's probably where it comes from.

PR #1189, opened 2025-10-31
AVX runtime float/double swizzle small improvement

AntoinePrv:
I have a number of of swizzle improvements to suggest, but I am starting small to get better accustomed to xsimd.

What do you make of the following change? My motivation was that `_mm256_permute2f128_ps` is the most expensive operation (though not sure if that's a problem in a CPU pipeline) so this PR suggest using it only once.

It also replaces modulo with a select mask to make sure this is properly optimized.

### DIFF
--- a/include/xsimd/arch/xsimd_avx.hpp
+++ b/include/xsimd/arch/xsimd_avx.hpp
@@ -1447,14 +1447,15 @@ namespace xsimd
         XSIMD_INLINE batch<double, A> swizzle(batch<double, A> const& self, batch<uint64_t, A> mask, requires_arch<avx>) noexcept
         {
             // swap lanes
-            __m256 swapped = _mm256_permute2f128_pd(self, self, 0x01); // [high | low]
+            __m256d swapped = _mm256_permute2f128_pd(self, self, 0x01); // [high | low]
 
-            // normalize mask taking modulo 2
-            batch<uint64_t, A> half_mask = mask & 0b1u;
+            // The half mask value is found in mask modulo 2, but the intrinsic expect it in the
+            // second least significant bit. We use negative as a cheap alternative to lshift.
+            batch<uint64_t, A> half_mask = -(mask & 0b1u);
 
             // permute within each lane
-            __m256 r0 = _mm256_permutevar_pd(self, half_mask);
-            __m256 r1 = _mm256_permutevar_pd(swapped, half_mask);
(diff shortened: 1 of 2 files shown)


serge-sans-paille:
Thanks! nice fine tuning \o/

AntoinePrv:
Looks like AVX tests are good now, but I'm unsure what the remaining failure is.
👀 @serge-sans-paille @JohanMabille

serge-sans-paille:
LGTM, I'll fix the emulated part, not something you should worry on. Would you mind squashing the last two commits?

serge-sans-paille:
@AntoinePrv when you squash, you can also rebase on `master` which now contains a fix for the emulated part

include/xsimd/arch/xsimd_avx.hpp
```diff
@@ -1427,41 +1427,39 @@ namespace xsimd
...
-            // duplicate low and high part of input
-            // Duplicate lanes separately
-            // 1) duplicate low and high lanes
-            __m256 lo = _mm256_permute2f128_ps(self, self, 0x00); // [low | low]
-            __m256 hi = _mm256_permute2f128_ps(self, self, 0x11); // [high| high]
+            // swap lanes
+            __m256 swapped = _mm256_permute2f128_ps(self, self, 0x01); // [high | low]
 
-            // normalize mask
-            batch<uint32_t, A> half_mask = mask % 4;
```
DiamonDinoia, line 1437: does this generate different asm actually?

PS I am fine either way. Might be worth having a `normalize<value>` API so that we can use everywhere that converts value to mask if it is a pow2

DiamonDinoia:
I like this PR! Nice that you found an ulterior way to optimize this!

I would suggest that generating the blend mask and the normalization to be pure method or class method that we can use elsewhere. I think we might need them here and there. 

This also allow to unit test them individually making debugging and development easier. 

Again, this is just a suggestion. Feel free to ignore it.

serge-sans-paille:
Merged as 9d41ad9be0d4f4b2e58552344f7753929e98afbe  (once squashed)

include/xsimd/arch/xsimd_avx.hpp
AntoinePrv: I'm unsure. For a regular integer operation I would say so but with intrinsics I had a doubt.

PR #1197, opened 2025-11-10
Add 8/16 bits AVX2 swizzle

AntoinePrv:
Add 8 and 16 bits AVX2 swizzle implementation.
These two sizes are not available on AVX.
Algorithm very similar to GH-1189.

### DIFF
--- a/include/xsimd/arch/xsimd_avx2.hpp
+++ b/include/xsimd/arch/xsimd_avx2.hpp
@@ -1142,36 +1142,17 @@ namespace xsimd
         }
 
         // swizzle (constant mask)
-        template <
-            class A, typename T,
-            uint8_t V0, uint8_t V1, uint8_t V2, uint8_t V3, uint8_t V4, uint8_t V5, uint8_t V6, uint8_t V7,
-            uint8_t V8, uint8_t V9, uint8_t V10, uint8_t V11, uint8_t V12, uint8_t V13, uint8_t V14, uint8_t V15,
-            uint8_t V16, uint8_t V17, uint8_t V18, uint8_t V19, uint8_t V20, uint8_t V21, uint8_t V22, uint8_t V23,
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
looking at the build failure on x86_64, you need to add some `std::enable_if` to disambiguate.

AntoinePrv:
Or could it be that I should add the 8/16 bits overload as batch constants as well (which we'd also want)?

AntoinePrv:
Seems like the failure is already on main.
The `batch_constant` is simply a naive fallback, I will address them in a separate PR.
CC @JohanMabille

serge-sans-paille:
Looks good, squash everything with a nice commit message and we're good to go!

include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -1095,42 +1095,86 @@ namespace xsimd
...
+        template <class A, typename T, detail::enable_sized_t<T, 2> = 0>
+        XSIMD_INLINE batch<T, A> swizzle(batch<T, A> const& self, batch<uint16_t, A> const& mask, requires_arch<avx2> req) noexcept
         {
-            return bitwise_cast<int32_t>(swizzle(bitwise_cast<uint32_t>(self), mask, avx2 {}));
+            return bitwise_cast<T>(swizzle(bitwise_cast<uint16_t>(self), mask, req));
         }
 
         // swizzle (constant mask)
+        template <
+            class A, typename T,
```
serge-sans-paille: Nit: could you use a variadic argument count here and just static_assert on the count? (here and the function below).

AntoinePrv:
@serge-sans-paille you should be able to squash-merge directly from the Github UI once ready.
Let me know if you prefer I squash instead.

serge-sans-paille:
@AntoinePrv : I've done it but I generally prefer when the author creates a consistent commit message during the squash: the author knows better!

PR #1200, opened 2025-11-11
Improve xsimd::expand common implementation

serge-sans-paille:
Previous implementation was making the assumption of cheap xsimd::insert, and it always generated batch::size inserts.

This implementation can take advantage of smaller popcount on the bitmask.

Note to self: it would be great to have a good implementation for constant mask.

### DIFF
--- a/include/xsimd/arch/common/xsimd_common_memory.hpp
+++ b/include/xsimd/arch/common/xsimd_common_memory.hpp
@@ -88,28 +88,21 @@ namespace xsimd
         }
 
         // expand
-        namespace detail
-        {
-            template <class IT, class A, class I, size_t... Is>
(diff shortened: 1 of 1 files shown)


DiamonDinoia:
Can you cherry pick my masks from the masked_ops? Then you have popcount for free. I will try to fix the mased memory ops in the following couple of days.

serge-sans-paille:
@DiamonDinoia that would only be useful for constant mask, right? We currently have expand only for dynamic mask (I dislike that, I'll have to provide a generic implementation for constant mask in the future, but post-release)

DiamonDinoia:
Oh yes, I deleted the dynamic mask when I removed the dynamic masked ops. With rebasing the changes might be lost.

PR #1141, opened 2025-07-08
Improving avx/avx2 swizzles

DiamonDinoia:
Hi,

This PR optimzes swizzles where possible
Fixed a bug in the float avx2 swizzle which did not allow duplicates
Adds more tests

PR #1201, opened 2025-11-11
Add Avx2 constant mask swizzle 8/16 and improve 32/64

AntoinePrv:
- Add AVX2 implementation of constant mask swizzle for `uint8_t` `uint16_t`
- Remove `all_different` limitation to `uint32_t` version, add `is_identity` case
- No significative changes for `uint64_t`. Making it closer the `uint32_t` implementation.
- Remove unused utility functions (`is_all_different`, `is_no_duplicates`).

Similar to the dynamic version, all functions are implemented for a sized `uint`, and other types forward to the proper size.

### DIFF
--- a/include/xsimd/arch/xsimd_avx2.hpp
+++ b/include/xsimd/arch/xsimd_avx2.hpp
@@ -1141,67 +1141,198 @@ namespace xsimd
             return bitwise_cast<T>(swizzle(bitwise_cast<uint16_t>(self), mask, req));
         }
 
+        namespace detail
+        {
+            template <typename T>
+            constexpr T swizzle_val_none()
+            {
+                // Most significant bit of the byte must be 1
+                return 0x80;
+            }
+
+            template <typename T>
+            constexpr bool swizzle_val_is_cross_lane(T val, T idx, T size)
+            {
+                return (idx < (size / 2)) != (val < (size / 2));
+            }
+
+            template <typename T>
+            constexpr bool swizzle_val_is_defined(T val, T size)
+            {
+                return (0 <= val) && (val < size);
+            }
+
+            template <typename T>
+            constexpr T swizzle_self_val(T val, T idx, T size)
+            {
+                return (swizzle_val_is_defined(val, size) && !swizzle_val_is_cross_lane(val, idx, size))
+                    ? val % (size / 2)
+                    : swizzle_val_none<T>();
+            }
+
+            template <typename T, typename A, T... Vals, std::size_t... Ids>
+            constexpr auto swizzle_make_self_batch_impl(::xsimd::detail::index_sequence<Ids...>)
+                -> batch_constant<T, A, swizzle_self_val(Vals, T(Ids), static_cast<T>(sizeof...(Vals)))...>
+            {
+                return {};
+            }
+
+            template <typename T, typename A, T... Vals>
+            constexpr auto swizzle_make_self_batch()
+                -> decltype(swizzle_make_self_batch_impl<T, A, Vals...>(::xsimd::detail::make_index_sequence<sizeof...(Vals)>()))
+            {
+                return {};
+            }
+
+            template <typename T>
+            constexpr T swizzle_cross_val(T val, T idx, T size)
+            {
+                return (swizzle_val_is_defined(val, size) && swizzle_val_is_cross_lane(val, idx, size))
+                    ? val % (size / 2)
+                    : swizzle_val_none<T>();
+            }
+
+            template <typename T, typename A, T... Vals, std::size_t... Ids>
+            constexpr auto swizzle_make_cross_batch_impl(::xsimd::detail::index_sequence<Ids...>)
+                -> batch_constant<T, A, swizzle_cross_val(Vals, T(Ids), static_cast<T>(sizeof...(Vals)))...>
+            {
+                return {};
+            }
+
+            template <typename T, typename A, T... Vals>
+            constexpr auto swizzle_make_cross_batch()
+                -> decltype(swizzle_make_cross_batch_impl<T, A, Vals...>(::xsimd::detail::make_index_sequence<sizeof...(Vals)>()))
+            {
+                return {};
+            }
+        }
+
         // swizzle (constant mask)
-        template <class A, typename T, uint8_t... Vals, detail::enable_sized_t<T, 1> = 0>
-        XSIMD_INLINE batch<T, A> swizzle(batch<T, A> const& self, batch_constant<uint8_t, A, Vals...> mask, requires_arch<avx2> req) noexcept
+        template <class A, uint8_t... Vals>
+        XSIMD_INLINE batch<uint8_t, A> swizzle(batch<uint8_t, A> const& self, batch_constant<uint8_t, A, Vals...> mask, requires_arch<avx2>) noexcept
         {
             static_assert(sizeof...(Vals) == 32, "Must contain as many uint8_t as can fit in avx register");
-            return swizzle(self, mask.as_batch(), req);
+
+            XSIMD_IF_CONSTEXPR(detail::is_identity(mask))
+            {
+                return self;
+            }
+
+            constexpr auto lane_mask = mask % make_batch_constant<uint8_t, (mask.size / 2), A>();
+
+            XSIMD_IF_CONSTEXPR(!detail::is_cross_lane(mask))
+            {
+                return _mm256_shuffle_epi8(self, lane_mask.as_batch());
+            }
+            XSIMD_IF_CONSTEXPR(detail::is_only_from_lo(mask))
+            {
+                __m256i broadcast = _mm256_permute2x128_si256(self, self, 0x00); // [low | low]
+                return _mm256_shuffle_epi8(broadcast, lane_mask.as_batch());
+            }
+            XSIMD_IF_CONSTEXPR(detail::is_only_from_hi(mask))
+            {
+                __m256i broadcast = _mm256_permute2x128_si256(self, self, 0x11); // [high | high]
+                return _mm256_shuffle_epi8(broadcast, lane_mask.as_batch());
+            }
+
+            // swap lanes
+            __m256i swapped = _mm256_permute2x128_si256(self, self, 0x01); // [high | low]
+
+            // We can outsmart the dynamic version by creating a compile-time mask that leaves zeros
+            // where it does not need to select data, resulting in a simple OR merge of the two batches.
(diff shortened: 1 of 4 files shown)


AntoinePrv:
Ping @serge-sans-paille @JohanMabille and @DiamonDinoia @pitrou following conversation in GH-1141

AntoinePrv:
I'm surprised at that last failure, since it was green before and seems unrelated to these last changes.

I thought the "lane" logic was AVX specific but I see it is also there in SSE2 and AVX512 (where there are in fact four lanes). We should write a meta-algorithm that can be specialized with each intrinsics but this is a longer shot. At least we'll have these in for the release.
For `uint8_t` can also improve the `is_dup_lo` `is_dup_hi` logic with smaller sizes that can be permuted.

include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -1141,67 +1141,204 @@ namespace xsimd
...
+            uint16_t V8, uint16_t V9, uint16_t V10, uint16_t V11,
+            uint16_t V12, uint16_t V13, uint16_t V14, uint16_t V15>
+        XSIMD_INLINE batch<uint16_t, A> swizzle(
+            batch<uint16_t, A> const& self,
+            batch_constant<uint16_t, A, V0, V1, V2, V3, V4, V5, V6, V7, V8, V9, V10, V11, V12, V13, V14, V15>,
+            requires_arch<avx2> req) noexcept
+        {
+            const auto self_bytes = bitwise_cast<uint8_t>(self);
+            // If a mask entry is k, we want 2k in low byte and 2k+1 in high byte
+            auto constexpr mask_2k_2kp1 = batch_constant<
```
DiamonDinoia, line 1273: I like using auto too. But the codebase seems to do `constexpr batch_constant<....> mask_2k_2kp1 {};

include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -1141,67 +1141,204 @@ namespace xsimd
...
-        XSIMD_INLINE batch<double, A> swizzle(batch<double, A> const& self, batch_constant<uint64_t, A, V0, V1, V2, V3> mask, requires_arch<avx2>) noexcept
-        {
-            XSIMD_IF_CONSTEXPR(detail::is_identity(mask)) { return self; }
             XSIMD_IF_CONSTEXPR(!detail::is_cross_lane(mask))
             {
-                constexpr auto imm = ((V0 & 1) << 0) | ((V1 & 1) << 1) | ((V2 & 1) << 2) | ((V3 & 1) << 3);
-                return _mm256_permute_pd(self, imm);
+                constexpr auto lane_mask = mask % make_batch_constant<uint32_t, (mask.size / 2), A>();
+                // Cheaper intrinsics when not crossing lanes
+                // We could also use _mm256_permute_ps which uses a imm8 constant, though it has the
```
DiamonDinoia: then why not using _mm256_permute_ps ? I think there is a mask() method that returns an immediate? Or did I add it in masked_memory_ops?

include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -1141,67 +1141,204 @@ namespace xsimd
...
+            }
+
+            template <typename T>
+            constexpr bool swizzle_val_is_cross_lane(T val, T idx, T size)
+            {
+                return (idx < (size / 2)) != (val < (size / 2));
+            }
+
+            template <typename T>
+            constexpr bool swizzle_val_is_defined(T val, T size)
```
serge-sans-paille, line 1160: Note for my future self: I wonder if we should assert that val is always defined. I'm not quite sure we have a portable behavior across architectures if that's not the case.

include/xsimd/arch/xsimd_avx2.hpp
serge-sans-paille: That's fine, we can improve the code base!

include/xsimd/arch/xsimd_avx2.hpp
serge-sans-paille: And this one uses one more register, and one extra load to build the mask, so yeah, why?

AntoinePrv: > why not using _mm256_permute_ps

They had the same latency so I was not sure the immediate was better. I can make the change!

AntoinePrv:
> I assume you compared the generated assembly with GCC and clang.

@serge-sans-paille in fact no (I'm bad at it), my changes are based on the number of intrinsics called and their reported latency (the dispatch being done at compile time).
Do you recommend a particular way of doing it?

DiamonDinoia:
> > I assume you compared the generated assembly with GCC and clang.
> 
> @serge-sans-paille in fact no (I'm bad at it), my changes are based on the number of intrinsics called and their reported latency (the dispatch being done at compile time). Do you recommend a particular way of doing it?

I run compiler explorer locally on my machine. 

Otherwise breakpoint + gdb also works.

serge-sans-paille:
> > I assume you compared the generated assembly with GCC and clang.
> 
> @serge-sans-paille in fact no (I'm bad at it), my changes are based on the number of intrinsics called and their reported latency (the dispatch being done at compile time). Do you recommend a particular way of doing it?

If you're a decent assembly reader, just dump the assembly from your source code. You can take inspiration from this:

```
printf '#include <xsimd/xsimd.hpp>\nauto foo(xsimd::batch<float> a, xsimd::batch<float> b) { return xsimd::avg(a, b); }' | clang++ -mavx2 -O2 -S -xc++ -o- - -Iinclude -DNDEBUG | pygmentize -lasm
```

AntoinePrv:
For many of them, I unsurprisingly I see the instruction associated with the intrinsic.

For the most complex one, the `uint8_t` general case, clang seems to not respect the intrinsict:
c++
```cpp
__m256i swapped = _mm256_permute2x128_si256(self, self, 0x01); // [high | low]
constexpr auto self_mask = detail::swizzle_make_self_batch<uint8_t, A, Vals...>();
constexpr auto cross_mask = detail::swizzle_make_cross_batch<uint8_t, A, Vals...>();
__m256i r0 = _mm256_shuffle_epi8(self, self_mask.as_batch());
__m256i r1 = _mm256_shuffle_epi8(swapped, cross_mask.as_batch());
return _mm256_or_si256(r0, r1);
```
gcc-15 (good instructions)
```asm
.cfi_startproc
vperm2i128      $1, %ymm0, %ymm0, %ymm1
vpshufb .LC0(%rip), %ymm0, %ymm0
vpshufb .LC1(%rip), %ymm1, %ymm1
vpor    %ymm1, %ymm0, %ymm0
ret
```

<details>
<summary>gcc-15 dynamic version</summary>

```asm
.cfi_startproc
movl    $252645135, %eax
vperm2i128      $1, %ymm0, %ymm0, %ymm3
vmovd   %eax, %xmm2
movl    $269488144, %eax
vpbroadcastd    %xmm2, %ymm2
vpand   %ymm2, %ymm1, %ymm2
vpshufb %ymm2, %ymm3, %ymm3
vpshufb %ymm2, %ymm0, %ymm0
vmovd   %eax, %xmm2
vpbroadcastd    %xmm2, %ymm2
vpand   %ymm2, %ymm1, %ymm1
vpcmpeqb        .LC2(%rip), %ymm1, %ymm1
vpblendvb       %ymm1, %ymm0, %ymm3, %ymm0
ret
.cfi_endproc
```

</details>

clang-21 (not the instructions asked for)
```asm
.cfi_startproc
# %bb.0:
pushq   %rbp
.cfi_def_cfa_offset 16
.cfi_offset %rbp, -16
movq    %rsp, %rbp
.cfi_def_cfa_register %rbp
andq    $-32, %rsp
subq    $32, %rsp
movq    %rdi, %rax
vinserti128     $1, 16(%rbp), %ymm0, %ymm0
vmovdqa 16(%rbp), %ymm1
vmovdqa .LCPI0_0(%rip), %ymm2           # ymm2 = [0,0,0,0,u,u,u,u,u,u,u,u,u,u,u,u,255,255,255,255,u,u,255,255,u,u,u,u,255,255,0,0]
vpblendvb       %ymm2, %ymm0, %ymm1, %ymm0
vpshufb .LCPI0_1(%rip), %ymm0, %ymm0    # ymm0 = ymm0[0,0,0,0,1,1,1,1,2,2,2,2,3,3,3,3,30,30,30,30,30,30,30,30,30,30,30,29,23,19,19,16]
vmovdqa %ymm0, (%rdi)
movq    %rbp, %rsp
popq    %rbp
.cfi_def_cfa %rsp, 8
vzeroupper
retq
```

<details>
<summary>clang-21 dynamic version</summary>

```asm
.cfi_startproc
# %bb.0:
pushq   %rbp
.cfi_def_cfa_offset 16
.cfi_offset %rbp, -16
movq    %rsp, %rbp
.cfi_def_cfa_register %rbp
andq    $-32, %rsp
subq    $32, %rsp
movq    %rdi, %rax
vmovdqa 48(%rbp), %ymm0
vmovdqa 16(%rbp), %ymm1
vpermq  $78, %ymm1, %ymm2               # ymm2 = ymm1[2,3,0,1]
vpand   .LCPI0_0(%rip), %ymm0, %ymm3
vpshufb %ymm3, %ymm1, %ymm1
vpshufb %ymm3, %ymm2, %ymm2
vpand   .LCPI0_1(%rip), %ymm0, %ymm0
vpcmpeqb        .LCPI0_2(%rip), %ymm0, %ymm0
vpblendvb       %ymm0, %ymm1, %ymm2, %ymm0
vmovdqa %ymm0, (%rdi)
movq    %rbp, %rsp
popq    %rbp
.cfi_def_cfa %rsp, 8
vzeroupper
retq
```

</details>

DiamonDinoia:
In this case, I usually write a gtest stub to measure the performance
between compilers if clang is much slower I look for a workaround or raise
the issue to llvm.

serge-sans-paille:
It's generally fine for a compiler to further optimize a kernel we described in "the best way possible". And as suggested by @DiamonDinoia if that's an issue, we can open a bug on the upstream compiler.

clang represents vector intrinsics in a generic way in LLVM IR, optimizes them, and then perform instruction selection to generate assembly, it may happens that the end result is worst than what we expressed in the beginning. Or better :-)

Tracking the quality of generated code for each compiler and each compiler version is a tremendous goal, we have a best effort goal on that topic.

serge-sans-paille:
BTW, @AntoinePrv if you're happy with the result (I am) can you squash you commit stack and write a proper aggregated commit message so I can merge? thanks!

AntoinePrv:
@serge-sans-paille All good for me, thanks!

PR #1203, opened 2025-11-12
Add constant batch_constant generator

AntoinePrv:
A useful utility to make operations on batch_constants.


include/xsimd/types/xsimd_batch_constant.hpp
```diff
@@ -287,13 +302,41 @@ namespace xsimd
         return {};
     }
 
+    /**
```
serge-sans-paille, line 305: This is a user-facing API, but kit lacks the proper doxygen group and doc integration. I'll fix this.

PR #1206, opened 2025-11-14
Frail fix for batch_constant documentation

serge-sans-paille:


serge-sans-paille:
@AntoinePrv I'm not super happy with that change, but that's the best I found.

PR #1207, opened 2025-11-15
Improve and update documentation

### LINKED ISSUE #1202: [Documentation] ELI5 the interaction between Compiler settings and how Xsimd leverages runtime architecture detection
Perhaps I've missed it, but in the documentation, for a new user of this library, it is unclear what the ideal basic usage pattern is in terms of the compiler's target architecture and whatever is happening under the hood in Xsimd.

Do you target compilation at the highest CPU architecture possible and let Xsimd figure it out from there? The lowest? The former case would seem to prevent the binary from running on lower CPUs (because the compiler might optimize some code to something like AVX512x but a user's CPU is limited to AVX1). The latter case seems safer (run everywhere), but then does that constrain Xsimd to the maximum instruction set defined by the compiler?

For example, in the readme, the section "[Auto detection of the instruction set extension to be used](https://github.com/xtensor-stack/xsimd?tab=readme-ov-file#auto-detection-of-the-instruction-set-extension-to-be-used)" does not currently specify what to do with the compilation settings. However, the block above "[Explicit use of an instruction set extension](https://github.com/xtensor-stack/xsimd?tab=readme-ov-file#explicit-use-of-an-instruction-set-extension)" specifies to set the compiler to target AVX. The same clarity for the automatic case would be extremely useful.

Dispatch would seem to handle this at runtime... but it's still unclear what architecture to target.

In summary, just a little bit of clarification in the docs would go a long way, I think.

serge-sans-paille:
Update supported architecture and make various usage scenario more explicit.

Fix #1202

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ This example outputs:
 
 ### Auto detection of the instruction set extension to be used
 
-The same computation operating on vectors and using the most performant instruction set available:
+The same computation operating on vectors and using the most performant instruction set available at compile time, based on the provided compiler flags (e.g. ``-mavx2`` for GCC and Clang to target AVX2):
 
 ```cpp
 #include <cstddef>
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -12,15 +12,27 @@ C++ wrappers for SIMD intrinsics.
 Introduction
 ------------
 
-SIMD (Single Instruction, Multiple Data) is a feature of microprocessors that has been available for many years. SIMD instructions perform a single operation
+`SIMD`_ (Single Instruction, Multiple Data) is a feature of microprocessors that has been available for many years. SIMD instructions perform a single operation
 on a batch of values at once, and thus provide a way to significantly accelerate code execution. However, these instructions differ between microprocessor
 vendors and compilers.
 
 `xsimd` provides a unified means for using these features for library authors. Namely, it enables manipulation of batches of scalar and complex numbers with the same arithmetic
 operators and common mathematical functions as for single values.
 
-`xsimd` makes it easy to write a single algorithm, generate one version of the algorithm per micro-architecture and pick the best one at runtime, based on the
-running processor capability.
+There are several ways to use `xsimd`:
+
+- one can write a generic, vectorized, algorithm and compile it as part of their
+  application build, with the right architecture flag;
+
+- one can write a generic, vectorized, algorithm and compile several version of
+  it by just changing the architecture flags, then pick the best version at
+  runtime;
+
+- one can write a vectorized algorithm specialized for a given architecture and
+  still benefit from the high-level abstraction proposed by `xsimd`.
+
+Of course, nothing prevents the combination of several of those approach, but
+more about this in section :ref:`Writing vectorized code`.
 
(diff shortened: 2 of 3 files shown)


serge-sans-paille:
cc @Jaegermeiste

README.md
```diff
@@ -142,7 +142,7 @@ This example outputs:
 
 ### Auto detection of the instruction set extension to be used
 
-The same computation operating on vectors and using the most performant instruction set available:
+The same computation operating on vectors and using the most performant instruction set available at compile time, absed on the provided compiler flags (e.g. ``-mavx2`` for GCC and Clang to target AVX2):
```
DiamonDinoia: we could add a reference to [this](https://gitlab.com/x86-psABIs/x86-64-ABI) and maybe suggest the arch levels to the user?

DiamonDinoia: I provide an example here: https://github.com/DiamonDinoia/random

serge-sans-paille:
> we could add a reference to [2]this and maybe suggest the arch levels to the
> user?

I'm not sure about that: we're ine README, and I assume people should know stuff
about SIMD when using xsimd. One thing we could do is point at an academic
lesson on SIMD (in addition to the wikipedia page I've already added).

DiamonDinoia:
> we could add a reference to [2]this and maybe suggest the arch levels to the user?
> I'm not sure about that: we're ine README, and I assume people should know stuff about SIMD when using xsimd. One thing we could do is point at an academic lesson on SIMD (in addition to the wikipedia page I've already added).

I agree, maybe in the dynamic dispatch page we can mention this.

Jaegermeiste:
All these changes look positive; however, they still don't make clear the interaction between Xsimd's capabilities and the compiler target for the application.

We have 3 primary variables:
1. Processor Capabilities
2. Compiler Architecture Target for Xsimd library
3. Compiler Architecture Target for the client application

What is the interplay between these variables?

Can I do something like compile Xsimd for AVX512 no matter what, and automagically, with one client binary (compiled targeting say SSE2, since that was the former Windows 10 baseline hardware requirement), and if the processor supports say AVX2, Xsimd will leverage that? Or are we capped at SSE2 no matter what in this scenario?

If I'm using dispatch to handle multiple architectures, say SSE2 and AVX2, do I target SSE2 or AVX2 in my compiler settings?

Answers to these questions are still not made clear.

This has little to do with understanding how SIMD works per se, nor processor architecture settings, and everything to do with how the library works under the hood, which is still not crystal clear.

DiamonDinoia:
[simd leverage.pdf](https://github.com/user-attachments/files/23750742/simd.leverage.pdf)

Hi @Jaegermeiste, 

Maybe this presentation helps? 

- Slide 10/11 show the micro-architecture levels and compile flags for gcc/clang
- Slide 12 shows how things are compiled and linked together.

Jaegermeiste:
Interesting - so that makes sense: compile the base application and dispatcher at the lowest common denominator (e.g. SSE2), and each "advanced" version of a translation unit needs to be separately compiled with the appropriate flags, prior to the linking stage.

That clears the fog for me.

I think that's the clarity that needs to explictly be in the documentation.

A nice bonus would be how to reasonably accomplish setting those compiler flags at the per-file level in MSVC and GCC, but I have a feeling you guys would see that as out of scope.

PR #1208, opened 2025-11-17
Add batch_constant comparison and shift

AntoinePrv:
- Add batch_constant comparison operators**
- Fix `size_t` not being C
- Add shift operators

### DIFF
--- a/include/xsimd/types/xsimd_batch_constant.hpp
+++ b/include/xsimd/types/xsimd_batch_constant.hpp
@@ -12,6 +12,8 @@
 #ifndef XSIMD_BATCH_CONSTANT_HPP
 #define XSIMD_BATCH_CONSTANT_HPP
 
+#include <cstddef>
+
 #include "./xsimd_batch.hpp"
 #include "./xsimd_utils.hpp"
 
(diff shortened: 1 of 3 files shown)


include/xsimd/types/xsimd_batch_constant.hpp
```diff
@@ -227,6 +216,56 @@ namespace xsimd
...
+        struct boolean_eq
+        {
+            constexpr bool operator()(T x, T y) const { return x == y; }
+        };
+        struct boolean_ne
+        {
+            constexpr bool operator()(T x, T y) const { return x != y; }
+        };
+        struct boolean_gt
+        {
```
serge-sans-paille, line 252: Better use https://en.cppreference.com/w/cpp/utility/functional/greater.html et cie?

AntoinePrv: I wanted to but they are only constexpr since C++14

AntoinePrv:
@serge-sans-paille this is ready for me if it is for you

DiamonDinoia:
This is complementary to #1162 that adds other binary ops. Might cause some merge conflicts.

include/xsimd/types/xsimd_batch_constant.hpp
serge-sans-paille: Indeed!

PR #1212, opened 2025-11-18
Proposal for Pixi workflow

### LINKED ISSUE #1209: Pixi dev workflow
Currently, I locally use [Pixi](https://pixi.sh) which is a modern project-oriented Conda environment manager, which I would like to contribute here.

It can easily manage multiple environment and tasks (along with task dependencies).
I have it setup so that running `pixi run -e clang-17 test debug-avx` will run an `avx` build with `clang-17` compiler from scratch.
That is, this command is all it takes to get stated, from there dependencies download, env creation, building are all done automatically (with caching).

It's extremely convenient to quickly test any configuration, and very friendly to beginners.
Of course, having such manifest file does not mean all developers *have* to use it, the project keep working without.

As discussed with @JohanMabille

AntoinePrv:
This is the proposed Pixi workflow

Tasks:
- [x] Add pixi.toml
- [x] Add CMakePresets
- [x] Add usage to documentation
- [x] Add sample usage in CI

Some notes and possible improvements:
- The `CMakePresets.json` is a bit verbose but I prefered to keep it modular so that we can derive all useful presets. It would be nice to make it converge with the options in CI so that it can be use there too. This is low risk since it is simply a different way to configure the same options.
- On the CI workflow sample: along with the official `setup-pixi` GH action that caches dependencies, this is a very powerful way to run a lot of configurations. If the first job works well, it could also be used to run some simple gcc/clang configurations.

Close #1209

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,121 @@
+{
+  "version": 5,
+  "cmakeMinimumRequired": {
+    "major": 3,
+    "minor": 23,
+    "patch": 0
+  },
+  "configurePresets": [
+    {
+      "name": "xsimd-all",
+      "cacheVariables": {
+        "ENABLE_XTL_COMPLEX" :"ON",
+        "BUILD_TESTS": "ON",
+        "BUILD_BENCHMARK": "ON"
+      }
+    },
+
+    {
+      "name": "debug-base",
+      "hidden": true,
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE" :"Debug"
+      }
+    },
+    {
+      "name": "release-base",
+      "hidden": true,
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE" :"Release"
+      }
+    },
+
+    {
+      "name": "native-base",
+      "hidden": true,
+      "cacheVariables": { "TARGET_ARCH": "native" }
+    },
+    {
+      "name": "sse2-base",
+      "hidden": true,
+      "cacheVariables": { "TARGET_ARCH": "x86-64 -mno-sse3" }
+    },
+    {
+      "name": "sse3-base",
+      "hidden": true,
+      "cacheVariables": { "TARGET_ARCH": "core2" }
+    },
+    {
+      "name": "ssse3-base",
+      "hidden": true,
+      "cacheVariables": { "TARGET_ARCH": "core2 -mssse3" }
+    },
+    {
+      "name": "sse4.1-base",
+      "hidden": true,
+      "cacheVariables": { "TARGET_ARCH": "nehalem" }
+    },
+    {
+      "name": "sse4.2-base",
+      "hidden": true,
+      "cacheVariables": { "TARGET_ARCH": "nehalem" }
+    },
+    {
+      "name": "avx-base",
+      "hidden": true,
+      "cacheVariables": { "TARGET_ARCH": "nehalem -msse4.2" }
+    },
+    {
+      "name": "avx2-base",
+      "hidden": true,
+      "cacheVariables": { "TARGET_ARCH": "haswell" }
+    },
+    {
+      "name": "neon-base",
+      "hidden": true,
+      "cacheVariables": { "TARGET_ARCH": "armv8-a" }
+    },
+
+    {
+      "name": "dev-base",
+      "hidden": true,
+      "inherits": ["debug-base", "xsimd-all"]
+    },
+    {
+      "name": "debug-native",
+      "inherits": ["dev-base", "native-base"]
+    },
+    {
+      "name": "debug-sse2",
+      "inherits": ["dev-base", "sse2-base"]
+    },
+    {
+      "name": "debug-sse3",
+      "inherits": ["dev-base", "sse3-base"]
+    },
+    {
+      "name": "debug-ssse3",
+      "inherits": ["dev-base", "ssse3-base"]
+    },
+    {
+      "name": "debug-sse4.1",
+      "inherits": ["dev-base", "sse4.1-base"]
+    },
+    {
+      "name": "debug-sse4.2",
+      "inherits": ["dev-base", "sse4.2-base"]
+    },
+    {
+      "name": "debug-avx",
+      "inherits": ["dev-base", "avx-base"]
+    },
+    {
+      "name": "debug-avx2",
+      "inherits": ["dev-base", "avx2-base"]
+    },
+    {
+      "name": "debug-neon",
+      "inherits": ["dev-base", "neon-base"]
+    }
+  ]
+}
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -33,47 +33,52 @@
     {
       "name": "native-base",
       "hidden": true,
-      "cacheVariables": { "TARGET_ARCH": "native" }
+      "cacheVariables": { "CMAKE_CXX_FLAGS": "$CXXFLAGS -march=native" }
     },
     {
       "name": "sse2-base",
       "hidden": true,
-      "cacheVariables": { "TARGET_ARCH": "x86-64 -mno-sse3" }
+      "cacheVariables": { "CMAKE_CXX_FLAGS": "$CXXFLAGS -march=x86-64 -msse2 -mno-sse3" }
     },
     {
       "name": "sse3-base",
       "hidden": true,
-      "cacheVariables": { "TARGET_ARCH": "core2" }
+      "cacheVariables": { "CMAKE_CXX_FLAGS": "$CXXFLAGS -march=x86-64 -msse3 -mno-ssse3" }
     },
     {
       "name": "ssse3-base",
       "hidden": true,
-      "cacheVariables": { "TARGET_ARCH": "core2 -mssse3" }
+      "cacheVariables": { "CMAKE_CXX_FLAGS": "$CXXFLAGS -march=x86-64 -mssse3 -mno-sse4.1" }
     },
     {
       "name": "sse4.1-base",
       "hidden": true,
-      "cacheVariables": { "TARGET_ARCH": "nehalem" }
+      "cacheVariables": { "CMAKE_CXX_FLAGS": "$CXXFLAGS -march=x86-64 -msse4.1 -mno-sse4.2" }
     },
     {
       "name": "sse4.2-base",
       "hidden": true,
-      "cacheVariables": { "TARGET_ARCH": "nehalem" }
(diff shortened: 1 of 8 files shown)


AntoinePrv:
@serge-sans-paille I am curious what you make of this error (with `gcc-15` and `-march=x86-64 -mavx2 -mno-avx512f`)

```
sccache /home/debian/xsimd/.pixi/envs/gcc-15/bin/g++ -DXSIMD_ENABLE_XTL_COMPLEX=1 -I/home/debian/xsimd/include -isystem /home/debian/xsimd/.pixi/envs/gcc-15/include -march=x86-64 -mavx2 -mno-avx512f -Wunused-parameter -Wextra -Wreorder -std=c++17 -g -fPIC -g -fdiagnostics-color=always -MD -MT test/CMakeFiles/test_xsimd.dir/test_load_store.cpp.o -MF test/CMakeFiles/test_xsimd.dir/test_load_store.cpp.o.d -o test/CMakeFiles/test_xsimd.dir/test_load_store.cpp.o -c /home/debian/xsimd/test/test_load_store.cpp
In file included from /home/debian/xsimd/include/xsimd/types/../arch/xsimd_isa.hpp:64,
                 from /home/debian/xsimd/include/xsimd/types/xsimd_batch.hpp:528,
                 from /home/debian/xsimd/include/xsimd/xsimd.hpp:77,
                 from /home/debian/xsimd/test/test_load_store.cpp:12:
In function 'xsimd::batch<int, A> xsimd::kernel::load_masked(const int32_t*, xsimd::batch_bool_constant<int, A, Values ...>, convert<int>, Mode, requires_arch<xsimd::avx2>) [with A = xsimd::avx2; bool ...Values = {false, false, false, false, false, false, false, false}; Mode = xsimd::aligned_mode]',
    inlined from 'xsimd::batch<unsigned int, A> xsimd::kernel::load_masked(const uint32_t*, xsimd::batch_bool_constant<unsigned int, A, Values ...>, convert<unsigned int>, Mode, requires_arch<xsimd::avx2>) [with A = xsimd::avx2; bool ...Values = {false, false, false, false, false, false, false, false}; Mode = xsimd::aligned_mode]' at /home/debian/xsimd/include/xsimd/types/../arch/./xsimd_avx2.hpp:171:42:
/home/debian/xsimd/include/xsimd/types/../arch/./xsimd_avx2.hpp:163:40: error: inlining failed in call to 'always_inline' 'xsimd::batch<int, A> xsimd::kernel::load_masked(const int32_t*, xsimd::batch_bool_constant<int, A, Values ...>, convert<int>, Mode, requires_arch<xsimd::avx2>) noexcept [with A = xsimd::avx2; bool ...Values = {false, false, false, false, false, false, false, false}; Mode = xsimd::aligned_mode]': function not considered for inlining
  163 |         XSIMD_INLINE batch<int32_t, A> load_masked(int32_t const* mem, batch_bool_constant<int32_t, A, Values...> mask, convert<int32_t>, Mode, requires_arch<avx2>) noexcept
      |                                        ^~~~~~~~~~~
/home/debian/xsimd/include/xsimd/types/../arch/./xsimd_avx2.hpp:165:34: note: called from here
  165 |             return load_masked<A>(mem, mask, convert<int32_t> {}, Mode {}, avx2 {});
      |                    ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  ```
  
  And the function in question. Could it be an infinite recursion? Or do my compilation option not make sense?
  
  ```cpp
  template <class A, bool... Values, class Mode>
XSIMD_INLINE batch<int32_t, A> load_masked(int32_t const* mem, batch_bool_constant<int32_t, A, Values...> mask, convert<int32_t>, Mode, requires_arch<avx2>) noexcept
{
    return load_masked<A>(mem, mask, convert<int32_t> {}, Mode {}, avx2 {});
}
```

This seems to start with #1162, it should work well without it.

DiamonDinoia:
can you try https://github.com/DiamonDinoia/xsimd/tree/pixi-workflow-fix

it is not infinite recursion because of SFINAE. All of this can be greatly simplified with c++17 once available :)

AntoinePrv:
@DiamonDinoia yes runs smoothly for me: I tried clang-18, clang-21 and gcc-15 🙏

AntoinePrv:
@serge-sans-paille what do you think of this?

pixi.lock
```diff
@@ -0,0 +1,6391 @@
+version: 6
```
serge-sans-paille, line 1: I assume (and hope) this file should not be checked in?

AntoinePrv: Well it depends. Nowadays it is not uncommon to check in lockfiles in the repository to use them in CI.
This one is rather large because we create all these different compilers environments.
To use `pixi` in CI, I would say we should keep it because it avoids solving end make caching easier. However we can still do without I believe (need to try out other options). Your call.

AntoinePrv: Note that this can also be hidden by default in github PRs (and stats I believe).
https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

AntoinePrv:
Requesting your feedback again @serge-sans-paille.

I personally don't think the lock file is an issue, it's pretty common nowadays. However if we take it out, I would also disable the CI.

AntoinePrv:
I see the Github action is failing due to missing docker image, perhaps this could be the opportunity to use pixi for running clang-format and other static analyzers ;)

AntoinePrv:
@serge-sans-paille if you don't have strong objections, I would like to get this one in, along with the lockfile which should not change unless we modify things and is hidden from diff.

serge-sans-paille:
I'm honestly not a fan of tool-specific configuration in the main repo (now pixi, then what? Dockerfiles ?).
And even less of a fan of keeping the lock/cache file in-tree.
For CI, this will bring more reproducibility, but maybe we want diversity in CI too, or at least something closer to dev environment instead of normalizing it?

Sorry for not being thrilled by this PR, we already have a lot of commit noise due to CI, I'd like to not too increase that :-/

AntoinePrv:
> I'm honestly not a fan of tool-specific configuration in the main repo (now pixi, then what? Dockerfiles ?).

I think we owe to contributors to have **one** way of getting started quickly. It could have been Dockerfiles, or perhaps Nix, or anything. I find this one to be low friction (it's declarative, has the same usage on all platforms, and does not take `update-alternatives` modifying global user system for the use of a single project). It is also a stack that we understand pretty well.
Documentation is not a good solution IMHO as it is not tested automatically and get outdated fast.

>  And even less of a fan of keeping the lock/cache file in-tree.

It's not pretty, but it speed things up, and time is important too.

> but maybe we want diversity in CI too

I agree with this, and I did not mean to replace all CI at all. For diversity, we should in fact include *some* conda-forge builds. But I was mostly thinking of streamlining things like
- `sudo apt install clang-tools` which is not pinned to any version
- `python3 -m pip install ninja` which is kind of weird
- unknown version `cmake` from Github's own runners *etc*
- other tools we may wish to add: cmake-format, typos, maybe conventional commit tooling etc.

I'd add a grain of salt though. Diversity is good, and we want to check many settings, but the goal is also **not** to check that *cmake* or *gcc-15* are properly working across all distributions. We're testing xsimd, not the tools.

> or at least something closer to dev environment instead of normalizing it

I'm not sure what you mean. I did not mean to impose this workflow. In fact it is very much optional in this PR.

> Sorry for not being thrilled by this PR, we already have a lot of commit noise due to CI, I'd like to not too increase that :-/

I understand, I also don't like working on CI. That's why I like approaches where I can reproduce as much as possible locally without many steps, or without making system-level changes (additionally without CI provider lock-ins).
I will never get 100% the same, especially some things are not worth the effort to setup on our own (MSVC, AppleClang...). But at least using the same tools, I know there are no CDN issue, no misconfiguration, no script that silently fails somewhere *etc.* and keep CI failure for things worth noticing.

serge-sans-paille:
I wanted to get deeper knowledge of pixi, and started with

```
$ pixi tree
  × failed to parse project manifest
    ╭─[pixi.toml:22:2]
 21 │ # https://pixi.sh
 22 │ [workspace]
    ·  ────┬────
    ·      ╰── unknown field `workspace`, expected one of `project`, `system-requirements`, `target`, `dependencies`, `host-dependencies`, `build-dependencies`, `pypi-dependencies`, `activation`, `tasks`, `feature`, `environments`, `pypi-options`, `tool`, `$schema`
 23 │ channels = ["conda-forge"]
    ╰────
```

not a good start /o\

AntoinePrv:
First thank you for giving it a try ! As for the error, you must be running an outdated version of pixi.

serge-sans-paille:
After some playing around, I'll answer to that one statement

> I think we owe to contributors to have one way of getting started quickly.

I do generally agree, but I don't think pixi is the best approach there. First xsimd is a rather low-dependency project: you only need a compiler and cmake to start hacking. the test dependency can be downloaded from cmake, the formatting is pretty standard (clang-format / git-clang-format). 

Pixi is not a widely deployed tool, you have a perfect `curl | sh` as the default install, so it's not the most inviting tool.
Digging into the pixi configuration, it suggests that we should provide a standard way to run various actions, and have either ci action, cmake, pixi or whatever use them. But the central place should be mostly env agnostic.

I'm not opposed to us having a pixi configuration file hanging around in a `tooling/directory`, no caching, and anyone watning to use pixi could just symlink it (or set the appropriate env variable etc). That pixi config should use whatever standard stuff we've setup (I like your cmake presets for instance).

That's my take on it!

CMakePresets.json
```diff
@@ -0,0 +1,211 @@
+{
```
serge-sans-paille, line 1: love this! should we also have some for ci?

pixi.toml
```diff
@@ -0,0 +1,230 @@
...
+typos-lsp = "*"
+neocmakelsp = "*"
+[feature.dev.target.linux.dependencies]
+gdb = "*"
+valgrind = "*"
+[feature.dev.target.osx.dependencies]
+lldb = "*"
+
+[feature.dev.tasks.fmt-clang]
+cmd = "find . -name '*.[ch]pp' | xargs clang-format -i"
```
serge-sans-paille, line 206: we should probably have this as a script somewhere, althoug I personnaly just rely on `git-clang-format`

serge-sans-paille (changes requested):
I'm definitively not a fan of keeping the lock file checked-in.
From a fresh clone, running `pixi tree` on my setup takes less that one second, lock file creation included

AntoinePrv:
Thanks for taking the time to try it. Ok first let me make a PR for the CMakePresets and perhaps some utilities in `dev` folder.

PR #1215, opened 2025-11-19
Improve Avx2 constant shift on uint8_t and fix sse2

AntoinePrv:
- **Fix shift tests**
- **Add 8bit lshift avx2 constant**
- **Improve 8bit lshift avx2 constant**

Not sure if that's worth it for the dynamic version though


AntoinePrv:
Looks like the tests were not set properly and this caught a bug.

AntoinePrv:
@serge-sans-paille this is good to go.
For me you can just squash-merge with the PR title if you're happy with the changes.

PR #1216, opened 2025-11-20
Improve masked load/store for sse2 when only the first element is sel…

serge-sans-paille:
…ected

### DIFF
--- a/include/xsimd/arch/xsimd_sse2.hpp
+++ b/include/xsimd/arch/xsimd_sse2.hpp
@@ -1068,10 +1068,45 @@ namespace xsimd
         }
 
         // load_masked
+        template <class A, class T, bool... Values, class Mode, class = typename std::enable_if<std::is_integral<T>::value>::type>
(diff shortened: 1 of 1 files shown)


serge-sans-paille:
cc @DiamonDinoia . It mostly helps GCC, clang is generally capable of finding the right pattern on its own.

DiamonDinoia:
Nice! I will review more carefully next week but I like the changes.

serge-sans-paille:
@DiamonDinoia I've rebased on top of latest changes, we're getting close to a release, I only have to implement wasm support and we're good to go (and maybe ppc too)...

serge-sans-paille:
@DiamonDinoia gentle ping :-)

DiamonDinoia (approved):
LGTM.

I was thinking of doing something like: `mask.countr_one() == 1 && mask.popcount() == 1` instead of `mask.mask() == 0x1` bit I think the latter is better in this case.

PR #1220, opened 2025-11-21
Add shift by multiple constant

### LINKED ISSUE #1218: Proposal for `batch_constant` biwise shift.
There is a trick borrowed from Lemire that I am using to implement left shift when unavailable: multiply by 2^n.
(The rest of the trick is that for right shifting, if you known that you data will "fit", you can left shift everything by some amount followed by a constant right shift on the whole batch).

With this, we can add missing variable shift on `x86_64` for: 
- SSE2:
  - left `uint16_t`
  - left `uint32_t`
- AVX2:
  - left `uint16_t`

On top of this, we can implement variable shift by calling twice a shift twice + mask on an integer size twice as big, so that further enables:

- SSE2:
  - left `uint8_t`
- AVX2:
  - left `uint8_t`
  - right `uint16_t`
  - right `uint8_t`

The main drawback is adding a new `batch_constant` API for left shift (more code to re-dispatch in the general case). This is necessary because we need the left shift to create the power of two.

I think this is still useful because it enables some shifts for the whole SSE inheritance.

I have a working partial implementation that I could turn into a PR in the next couple days, and hopefully into the 14.0 release. What do you think @serge-sans-paille?

AntoinePrv:
@serge-sans-paille @JohanMabille this ideas works, but I cannot figure out how to refactor `bitwise_lshift_as_twice_larger` into a separate header.

The issue:
- `xsimd_sse2.hpp` needs `utils/shits.hpp` for `bitwise_lshift_as_twice_larger`
- but `bitwise_lshift_as_twice_larger` needs definitions from `xsimd_sse2.hpp`.
- and likewise for `avx2`

Forward declaring all the needed functions overloads from `sse2` and `avx2` and the arch types in `utils/shits.hpp`.
Perhaps it should use the functions from `xsimd_api` instead? (which would also be better if there is a better implementation further down in the inheritance tree).

Close #1218

### DIFF
--- a/include/xsimd/arch/utils/shifts.hpp
+++ b/include/xsimd/arch/utils/shifts.hpp
@@ -1,0 +1,82 @@
+/***************************************************************************
+ * Copyright (c) Johan Mabille, Sylvain Corlay, Wolf Vollprecht and         *
+ * Martin Renou                                                             *
+ * Copyright (c) QuantStack                                                 *
+ * Copyright (c) Serge Guelton                                              *
+ * Copyright (c) Marco Barbone                                              *
+ *                                                                          *
+ * Distributed under the terms of the BSD 3-Clause License.                 *
+ *                                                                          *
+ * The full license is in the file LICENSE, distributed with this software. *
+ ****************************************************************************/
+
+#ifndef XSIMD_UTILS_SHIFTS_HPP
+#define XSIMD_UTILS_SHIFTS_HPP
+
+#include "../../config/xsimd_inline.hpp"
+#include "../../types/xsimd_batch.hpp"
+#include "../../types/xsimd_batch_constant.hpp"
+#include "../../types/xsimd_traits.hpp"
+
+namespace xsimd
+{
+    namespace kernel
+    {
+        namespace utils
+        {
+            template <typename I, I offset, I length, I... Vs>
+            struct select_stride
+            {
+                template <typename K>
+                static constexpr K get(K i, K)
+                {
+                    constexpr I values_array[] = { Vs... };
+                    return static_cast<K>(values_array[length * i + offset]);
+                }
+            };
+
+            template <typename I>
+            constexpr I lsb_mask(I bit_index)
+            {
+                if (bit_index == 8 * sizeof(I))
+                {
+                    return ~I { 0 };
+                }
+                return static_cast<I>((I { 1 } << bit_index) - I { 1 });
+            }
+
+            template <class T, class A, T V0, T... Vs>
+            constexpr bool all_equals(batch_constant<T, A, V0, Vs...> c)
+            {
+                return (c == std::integral_constant<T, V0> {}).all();
+            }
+
+            template <class T, class A, T... Vs>
+            XSIMD_INLINE batch<T, A> bitwise_lshift_as_twice_larger(
+                batch<T, A> const& self, batch_constant<T, A, Vs...>) noexcept
+            {
+                using T2 = widen_t<T>;
+
+                const auto self2 = bitwise_cast<T2>(self);
+
+                // Lower byte: shift as twice the size and mask bits flowing to higher byte.
+                constexpr auto shifts_lo = make_batch_constant<T2, select_stride<T, 0, 2, Vs...>, A>();
+                constexpr auto mask_lo = lsb_mask<T2>(8 * sizeof(T));
+                const auto shifted_lo = bitwise_lshift(self2, shifts_lo);
+                constexpr auto batch_mask_lo = make_batch_constant<T2, mask_lo, A>();
+                const auto masked_lo = bitwise_and(shifted_lo, batch_mask_lo.as_batch());
+
+                // Higher byte: mask bits that would flow from lower byte and shift as twice the size.
+                constexpr auto shifts_hi = make_batch_constant<T2, select_stride<T, 1, 2, Vs...>, A>();
+                constexpr auto mask_hi = mask_lo << (8 * sizeof(T));
+                constexpr auto batch_mask_hi = make_batch_constant<T2, mask_hi, A>();
+                const auto masked_hi = bitwise_and(self2, batch_mask_hi.as_batch());
+                const auto shifted_hi = bitwise_lshift(masked_hi, shifts_hi);
+
+                return bitwise_cast<T>(bitwise_or(masked_lo, shifted_hi));
+            }
+        }
+    }
+}
+
+#endif
--- a/include/xsimd/arch/xsimd_avx2.hpp
+++ b/include/xsimd/arch/xsimd_avx2.hpp
@@ -17,6 +17,7 @@
 
 #include "../types/xsimd_avx2_register.hpp"
 #include "../types/xsimd_batch_constant.hpp"
+#include "./utils/shifts.hpp"
 
 #include <limits>
 
@@ -330,6 +331,29 @@ namespace xsimd
             {
                 return bitwise_lshift(self, other, avx {});
             }
+        }
+
+        // bitwise_lshift multiple (constant) specific implementations.
+        // Missing implementations are dispatched to the `batch` overload in xsimd_api.
+        // The 1 byte constant implementation calls the 2 bytes constant version, the 2 bytes
+        // constant version calls into the 4 bytes version which resolves to the dynamic one above.
+        template <class A, class T, T... Vs,
+                  std::enable_if_t<std::is_integral<T>::value && (sizeof(T) <= 2), int> = 0>
+        XSIMD_INLINE batch<T, A> bitwise_lshift(
+            batch<T, A> const& self, batch_constant<T, A, Vs...> shifts, requires_arch<avx2> req) noexcept
+        {
+            using uint_t = typename std::make_unsigned<T>::type;
+
+            // AVX2 only supports 16-bit shifts with a uniform bitshift value,
+            // otherwise emulate using 32-bit shifts.
+            XSIMD_IF_CONSTEXPR(utils::all_equals(shifts))
+            {
+                return bitwise_lshift<shifts.get(0), A>(self, req);
+            }
+            return bitwise_cast<T>(
+                utils::bitwise_lshift_as_twice_larger<uint_t>(
+                    bitwise_cast<uint_t>(self),
+                    batch_constant<uint_t, A, static_cast<uint_t>(Vs)...> {}));
         }
 
         // bitwise_or
--- a/include/xsimd/types/xsimd_api.hpp
+++ b/include/xsimd/types/xsimd_api.hpp
@@ -353,6 +353,43 @@ namespace xsimd
         return kernel::bitwise_cast<A>(x, batch<T_out, A> {}, A {});
     }
 
+    namespace detail
+    {
+        // Detection for kernel overloads accepting ``batch_constant`` in ``bitwise_lshift``
+        // directly (or in a parent register function).
+        // The ``batch_constant`` overload is a rare but useful optimization.
+        // Running the detection here is less error prone than to add a fallback to all
+        // architectures.
+
+        template <class Arch, class Batch, class BatchConstant, class = void>
+        struct has_bitwise_lshift_batch_const : std::false_type
+        {
+        };
+
+        template <class Arch, class Batch, class BatchConstant>
+        struct has_bitwise_lshift_batch_const<
+            Arch, Batch, BatchConstant,
+            void_t<decltype(kernel::bitwise_lshift<Arch>(
+                std::declval<Batch>(), std::declval<BatchConstant>(), Arch {}))>>
+            : std::true_type
+        {
+        };
+
+        template <class Arch, class T, T... Values>
+        XSIMD_INLINE batch<T, Arch> bitwise_lshift_batch_const(batch<T, Arch> const& x, batch_constant<T, Arch, Values...> shift, std::true_type) noexcept
+        {
+            // Optimized ``batch_constant`` implementation
+            return kernel::bitwise_lshift<Arch>(x, shift, Arch {});
+        }
+
+        template <class Arch, class T, T... Values>
+        XSIMD_INLINE batch<T, Arch> bitwise_lshift_batch_const(batch<T, Arch> const& x, batch_constant<T, Arch, Values...> shift, std::false_type) noexcept
+        {
+            // Fallback to regular run-time implementation
+            return kernel::bitwise_lshift<Arch>(x, shift.as_batch(), Arch {});
+        }
+    }
+
     /**
      * @ingroup batch_bitwise
      *
@@ -367,17 +404,24 @@ namespace xsimd
         detail::static_check_supported_config<T, A>();
         return kernel::bitwise_lshift<A>(x, shift, A {});
     }
+    template <size_t shift, class T, class A>
+    XSIMD_INLINE batch<T, A> bitwise_lshift(batch<T, A> const& x) noexcept
+    {
+        detail::static_check_supported_config<T, A>();
+        return kernel::bitwise_lshift<shift, A>(x, A {});
+    }
     template <class T, class A>
     XSIMD_INLINE batch<T, A> bitwise_lshift(batch<T, A> const& x, batch<T, A> const& shift) noexcept
     {
         detail::static_check_supported_config<T, A>();
         return kernel::bitwise_lshift<A>(x, shift, A {});
     }
-    template <size_t shift, class T, class A>
-    XSIMD_INLINE batch<T, A> bitwise_lshift(batch<T, A> const& x) noexcept
+    template <class T, class A, T... Values>
+    XSIMD_INLINE batch<T, A> bitwise_lshift(batch<T, A> const& x, batch_constant<T, A, Values...> shift) noexcept
     {
         detail::static_check_supported_config<T, A>();
-        return kernel::bitwise_lshift<shift, A>(x, A {});
+        using has_batch_const_impl = detail::has_bitwise_lshift_batch_const<A, decltype(x), decltype(shift)>;
+        return detail::bitwise_lshift_batch_const<A>(x, shift, has_batch_const_impl {});
     }
 
     /**
--- a/include/xsimd/arch/xsimd_sse2.hpp
+++ b/include/xsimd/arch/xsimd_sse2.hpp
@@ -18,6 +18,7 @@
 
 #include "../types/xsimd_batch_constant.hpp"
 #include "../types/xsimd_sse2_register.hpp"
+#include "./utils/shifts.hpp"
 
 namespace xsimd
 {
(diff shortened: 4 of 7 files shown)


include/xsimd/arch/utils/shifts.hpp
AntoinePrv, line 1: I named this `util` as `common` seems to be more used for implementing the `common` architecture.

include/xsimd/arch/utils/shifts.hpp
```diff
@@ -0,0 +1,63 @@
...
+                template <typename K>
+                static constexpr K get(K i, K)
+                {
+                    return static_cast<K>(values_array[length * i + offset]);
+                }
+            };
+
+            template <class T, class T2, class A, class R, T... Vs>
+            XSIMD_INLINE batch<T, A> bitwise_lshift_as_twice_larger(
+                batch<T, A> const& self, batch_constant<T, A, Vs...>, R req) noexcept
```
AntoinePrv: Probably should remove `R req` here

AntoinePrv (commented):
Also we need to add all the proper forward to dynamic for all other architectures to avoid ending up with a `common` implementation (unless that's all there is).

serge-sans-paille:
@AntoinePrv unless there's is a huge need on your side, I won't schedule that one for the release. Let's just add masked load / store support for other architecture so that we can release, then begin a new commit wave :-)

AntoinePrv:
@serge-sans-paille I had hoped I could do it quickly before the weekend but it seems it will take longer.
Let's release indeed, I'll add the fix in Arrow with a TODO for cleaning it up in a future release.

AntoinePrv:
I have a few questions on the current implementation

- For the public API, wouldn't this benefit from a `std::enable_if<std::is_integral<T>::value>::type`?
  https://github.com/xtensor-stack/xsimd/blob/23faec42b777a847a2c61f2077564186e8242903/include/xsimd/types/xsimd_api.hpp#L370
- For neon, the unsigned API takes shifts as signed, yet there is not overload for unsigned shifts, which means that in this case we fallback on common?
  https://github.com/xtensor-stack/xsimd/blob/23faec42b777a847a2c61f2077564186e8242903/include/xsimd/arch/xsimd_neon.hpp#L2421

include/xsimd/types/xsimd_api.hpp
```diff
@@ -353,6 +353,41 @@ namespace xsimd
...
+        struct has_bitwise_lshift_batch_const : std::false_type
+        {
+        };
+
+        template <class A, class B, class C>
+        struct has_bitwise_lshift_batch_const<A, B, C,
+                                              void_t<decltype(kernel::bitwise_lshift<A>(std::declval<B>(), std::declval<C>(), A {}))>>
+            : std::true_type
+        {
+        };
```
AntoinePrv, line 376: We run the detection of an optimization in the `xsimd_api` to avoid adding many error-prone overloads.

AntoinePrv:
I've made a fresh take, and added the fallback to runtime API in `xsimd_api.hpp` since this is an edge-case API.

This seems much improved over the runtime version. E.g. with GCC 15, SSE2, and uin8_t

<details>
<summary>Compile time (new)</summary>

```asm
        .cfi_startproc
        movdqa  (%rdi), %xmm1
        movdqa  .LC0(%rip), %xmm0
        pcmpeqd %xmm2, %xmm2
        movdqa  %xmm2, %xmm3
        psllw   $8, %xmm2
        pmullw  %xmm1, %xmm0
        psrlw   $8, %xmm3
        pand    %xmm2, %xmm1
        pmullw  .LC3(%rip), %xmm1
        pand    %xmm3, %xmm0
        por     %xmm1, %xmm0
        ret
        .cfi_endproc
.LFE6862:
        .size   _Z1fRKN5xsimd5batchIhNS_4sse2EEE, .-_Z1fRKN5xsimd5batchIhNS_4sse2EEE
        .section        .rodata.cst16,"aM",@progbits,16
        .align 16
.LC0:
        .value  2
        .value  8
        .value  16
        .value  4
        .value  2
        .value  2
        .value  2
        .value  2
        .align 16
.LC3:
        .value  4
        .value  16
        .value  8
        .value  2
        .value  2
        .value  2
        .value  2
        .value  2
```

</details>

<details>
<summary>Run time (before)</summary>

```asm
        .cfi_startproc
        pushq   %r15
        .cfi_def_cfa_offset 16
        .cfi_offset 15, -16
        pushq   %r14
        .cfi_def_cfa_offset 24
        .cfi_offset 14, -24
        pushq   %r13
        .cfi_def_cfa_offset 32
        .cfi_offset 13, -32
        pushq   %r12
        .cfi_def_cfa_offset 40
        .cfi_offset 12, -40
        pushq   %rbp
        .cfi_def_cfa_offset 48
        .cfi_offset 6, -48
        pushq   %rbx
        .cfi_def_cfa_offset 56
        .cfi_offset 3, -56
        movzbl  15(%rdi), %eax
        movzbl  13(%rdi), %edx
        movzbl  12(%rdi), %ecx
        movzbl  8(%rdi), %r15d
        movzbl  7(%rdi), %r12d
        movzbl  11(%rdi), %esi
        movzbl  10(%rdi), %r8d
        movb    %al, -12(%rsp)
        movzbl  9(%rdi), %r13d
        movzbl  14(%rdi), %eax
        movb    %dl, -10(%rsp)
        xorl    %edx, %edx
        movzbl  6(%rdi), %r9d
        movzbl  3(%rdi), %ebx
        movb    %cl, -9(%rsp)
        xorl    %ecx, %ecx
        movzbl  5(%rdi), %r10d
        movzbl  4(%rdi), %r11d
        addl    %r13d, %r13d
        addl    %r8d, %r8d
        movzbl  2(%rdi), %ebp
        movzbl  1(%rdi), %r14d
        movb    %al, -11(%rsp)
        sall    $4, %ebx
        movzbl  (%rdi), %edi
        movzbl  %bl, %ebx
        sall    $4, %r11d
        movzbl  %r8b, %r8d
        sall    $2, %r14d
        sall    $3, %ebp
        movzbl  %r11b, %r11d
        salq    $24, %rbx
        movb    %dil, %dl
        movzbl  %bpl, %ebp
        salq    $32, %r11
        salq    $16, %rbp
        addb    %dl, %dl
        sall    $3, %r10d
        movq    %rdx, -40(%rsp)
        movq    %r14, %rdx
        movzbl  %r10b, %r10d
        sall    $2, %r9d
        movq    %rcx, -32(%rsp)
        salq    $40, %r10
        movzbl  %r9b, %r9d
        salq    $16, %r8
        movq    -40(%rsp), %rcx
        salq    $48, %r9
        movb    %dl, %ch
        movq    -32(%rsp), %rdx
        movq    %rcx, -40(%rsp)
        movq    -40(%rsp), %rdi
        movq    %rdx, -32(%rsp)
        movq    -32(%rsp), %rdx
        andq    $-16711681, %rdi
        orq     %rbp, %rdi
        movq    %rdx, -32(%rsp)
        movq    -32(%rsp), %rdx
        movabsq $-4278190081, %rbp
        movq    %rdi, -40(%rsp)
        movq    -40(%rsp), %rdi
        movq    %rdx, -32(%rsp)
        movq    -32(%rsp), %rdx
        andq    %rbp, %rdi
        orq     %rbx, %rdi
        movq    %rdx, -32(%rsp)
        movq    -32(%rsp), %rdx
        movabsq $-1095216660481, %rbx
        movq    %rdi, -40(%rsp)
        movq    -40(%rsp), %rdi
        movq    %rdx, -32(%rsp)
        movq    -32(%rsp), %rdx
        andq    %rbx, %rdi
        orq     %r11, %rdi
        movq    %rdx, -32(%rsp)
        movq    -32(%rsp), %rdx
        movabsq $-280375465082881, %r11
        movq    %rdi, -40(%rsp)
        movq    -40(%rsp), %rdi
        movq    %rdx, -32(%rsp)
        movq    -32(%rsp), %rcx
        andq    %r11, %rdi
        orq     %r10, %rdi
        movb    %r15b, %cl
        movabsq $-71776119061217281, %r10
        movq    %rdi, -40(%rsp)
        movq    -40(%rsp), %rdi
        addb    %cl, %cl
        andq    %r10, %rdi
        orq     %r9, %rdi
        leal    (%r12,%r12), %r9d
        movq    %rdi, -40(%rsp)
        movq    -40(%rsp), %r12
        salq    $56, %r9
        movabsq $72057594037927935, %rdi
        andq    %rdi, %r12
        orq     %r9, %r12
        movq    %r12, -40(%rsp)
        movq    -40(%rsp), %rdx
        movq    %rdx, -40(%rsp)
        movq    %r13, %rdx
        movq    %rcx, -32(%rsp)
        movq    -32(%rsp), %rcx
        movzbl  -11(%rsp), %eax
        movb    %dl, %ch
        movzbl  -10(%rsp), %edx
        movq    %rcx, -32(%rsp)
        movq    -32(%rsp), %r9
        movzbl  -9(%rsp), %ecx
        andq    $-16711681, %r9
        orq     %r9, %r8
        addl    %esi, %esi
        addl    %ecx, %ecx
        addl    %edx, %edx
        movq    %r8, -32(%rsp)
        movq    -32(%rsp), %r8
        movzbl  %sil, %esi
        addl    %eax, %eax
        salq    $24, %rsi
        movzbl  %cl, %ecx
        movzbl  %dl, %edx
        movzbl  %al, %eax
        salq    $32, %rcx
        andq    %rbp, %r8
        salq    $40, %rdx
        orq     %r8, %rsi
        salq    $48, %rax
        movq    %rsi, -32(%rsp)
        movq    -32(%rsp), %rsi
        andq    %rbx, %rsi
        orq     %rsi, %rcx
        movq    %rcx, -32(%rsp)
        movq    -32(%rsp), %rcx
        andq    %r11, %rcx
        orq     %rcx, %rdx
        movq    %rdx, -32(%rsp)
        movq    -32(%rsp), %rdx
        andq    %r10, %rdx
        orq     %rdx, %rax
        movq    %rax, -32(%rsp)
        movzbl  -12(%rsp), %eax
        leal    (%rax,%rax), %edx
        movq    -32(%rsp), %rax
        salq    $56, %rdx
        andq    %rdi, %rax
        orq     %rdx, %rax
        movq    %rax, -32(%rsp)
        movdqa  -40(%rsp), %xmm0
        popq    %rbx
        .cfi_def_cfa_offset 48
        popq    %rbp
        .cfi_def_cfa_offset 40
        popq    %r12
        .cfi_def_cfa_offset 32
        popq    %r13
        .cfi_def_cfa_offset 24
        popq    %r14
        .cfi_def_cfa_offset 16
        popq    %r15
        .cfi_def_cfa_offset 8
        ret
        .cfi_endproc
```

</summary>

AntoinePrv:
@serge-sans-paille this one is ready too

serge-sans-paille:
I'm a bit puzzled by the gain here: you're actually not adding anything except the possibility to pass a batch of constant, which is then dispatched to either the original code that takes a constant, or the original code that takes a batch, right?

AntoinePrv:
@serge-sans-paille There are three things in this PR:
- A new API for lshift taking a `batch_constant` as input
- Two cases (SSE2 with 2 bytes data and SSE4.1 with 4 byte data) with a new implementation that uses multiply as a fallback for missing left shit
- A utility function that can implement left shift using left shift on data twice as large.

They all play out nicely together, for instance with SSE2 and uint8_t data in https://github.com/xtensor-stack/xsimd/pull/1220#issuecomment-3909841924
- Call `bitwise_lshift<uint8_t>` 
  - Use `bitwise_lshift_as_twice_larger<uint16_t>`
    - There, use `the batch_constant` to make some computation at compile time
      - Call into to `bitwise_lshift<uint16_t>`
        - There, use the `batch_constant`  to generate the "mulitply" quantities at compile time
      - Call into to `bitwise_lshift<uint16_t>` (same)
        - ...

Result after the compiler does it jobs: with 2 multiply, 2 AND, 1 OR (and a few other things) you get a lshift for `uint8_t`.

`bitwise_lshift_as_twice_larger` actually does not care what the underlying shift implementation is.
In the he AVX2 case of this PR, there is `_mm256_sllv_epi32` available for 4 byte data (`bitwise_lshift<uint32_t>`). We use it as a target to generate
- `bitwise_lshift<uint16_t>`, using 2 `_mm256_sllv_epi32`, 2 AND, and 1 OR
- `bitwise_lshift<uint8_t>`, recursively using 4 `_mm256_sllv_epi32`, 6 AND, and 3 OR

From the code perspective of xsimd, these looks like very niche case. But in practice, this unlocks `bitwise_lshift` for many sizes in the SSE family. A common operation on a common architecture!

Another nice cheat happening here is that to generate the "multiply" quantities as a fallback for lshift, we use... lshift, but at compile time 😅

serge-sans-paille:
I'm all in for the second and third effect, but still puzzled by the first case :-)

include/xsimd/types/xsimd_api.hpp
```diff
@@ -367,17 +402,24 @@ namespace xsimd
...
         return kernel::bitwise_lshift<A>(x, shift, A {});
     }
-    template <size_t shift, class T, class A>
-    XSIMD_INLINE batch<T, A> bitwise_lshift(batch<T, A> const& x) noexcept
+    template <class T, class A, T... Values>
+    XSIMD_INLINE batch<T, A> bitwise_lshift(batch<T, A> const& x, batch_constant<T, A, Values...> shift) noexcept
     {
         detail::static_check_supported_config<T, A>();
-        return kernel::bitwise_lshift<shift, A>(x, A {});
+        using has_batch_const_impl = detail::has_bitwise_lshift_batch_const<A, decltype(x), decltype(shift)>;
```
serge-sans-paille, line 423: it seems to me that the dispatch here should be 'if all `Values...` are the same, dispatch to the overload that takes a single parameter, otherwise dispatch to the generic overload. Wouldn't that simplify the whole implementation?

AntoinePrv: That is a nice improvement to add as well, but this is not what `has_bitwise_lshift_batch_const` is checking.
It is a compile time check to detect if there is a `batch_constant` overload in the `kernel` namespace.

I added this pattern because in most cases, the extra overload would forward to the dynamic version.
This is tedious and possibly error-prone, *à la* #1267.

include/xsimd/arch/utils/shifts.hpp
```diff
@@ -0,0 +1,89 @@
...
+#include "../../types/xsimd_batch_constant.hpp"
+
+namespace xsimd
+{
+    namespace kernel
+    {
+        namespace utils
+        {
+            template <typename I, I offset, I length, I... Vs>
+            struct select_stride
```
pitrou, line 28: Perhaps a more hardcore C++ developer wouldn't mind, but I would add comments to these helpers so that people don't have to decipher the code.

include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -332,6 +333,40 @@ namespace xsimd
             }
         }
 
+        // bitwise_lshift multiple (constant) specific implementations.
```
pitrou, line 336: I would say something like:

```suggestion
        // bitwise_lshift with constant bitshifts, for 16-bit elements.
```

include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -332,6 +333,40 @@ namespace xsimd
...
+        // Missing implementations are dispacthed to the `batch` overload in xsimd_api.
+        template <class T, class A, T... Vs, detail::enable_sized_integral_t<T, 2> = 0>
+        XSIMD_INLINE batch<T, A> bitwise_lshift(
+            batch<T, A> const& self, batch_constant<T, A, Vs...> shifts, requires_arch<avx2> req) noexcept
+        {
+            using uint_t = typename std::make_unsigned<T>::type;
+
+            XSIMD_IF_CONSTEXPR(utils::all_equals(shifts))
```
pitrou, line 349: Add a rationale here? For example:

```suggestion
            // AVX2 only supports 16-bit shifts with a uniform bitshift value,
            // otherwise emulate using 32-bit shifts.
            XSIMD_IF_CONSTEXPR(utils::all_equals(shifts))
```

include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -332,6 +333,40 @@ namespace xsimd
...
+            {
+                return bitwise_lshift<shifts.get(0), A>(self, req);
+            }
+            return bitwise_cast<T>(
+                utils::bitwise_lshift_as_twice_larger<uint_t, uint32_t>(
+                    bitwise_cast<uint_t>(self),
+                    batch_constant<uint_t, A, static_cast<uint_t>(Vs)...> {}));
+        }
+
+        template <class T, class A, T... Vs, detail::enable_sized_integral_t<T, 1> = 0>
```
pitrou:
```suggestion
        // bitwise_lshift with constant bitshifts, for 8-bit elements.
        template <class T, class A, T... Vs, detail::enable_sized_integral_t<T, 1> = 0>
```

include/xsimd/types/xsimd_api.hpp
```diff
@@ -353,6 +353,41 @@ namespace xsimd
...
+        // The ``batch_constant`` overload is a rare but useful optimization.
+        // Running the detection here is less error prone than to add a fallback to all
+        // architectures.
+
+        template <class A, class B, class C, class = void>
+        struct has_bitwise_lshift_batch_const : std::false_type
+        {
+        };
+
+        template <class A, class B, class C>
```
pitrou: Naming those A, B and C makes the signature cryptic, how about meaningful parameter names?

JohanMabille: We should probably do a global pass on the APIs to get something homogeneous.

JohanMabille:
> I'm all in for the second and third effect, but still puzzled by the first case :-)

Not sure to get you.

include/xsimd/types/xsimd_api.hpp
serge-sans-paille: What about
`A`=`Arch`
`B`=`BatchType`
`C`=`IntOrBatchType`

include/xsimd/types/xsimd_api.hpp
serge-sans-paille: I know we've been having issues with error-prone overloads in the past, but I'm not a big fan of that approach.

I'll let it pass this time, but I think we should settle on something better.

include/xsimd/arch/utils/shifts.hpp
```diff
@@ -0,0 +1,89 @@
...
+                if (bit_index == 8 * sizeof(I))
+                {
+                    return ~I { 0 };
+                }
+                return static_cast<I>((I { 1 } << bit_index) - I { 1 });
+            }
+
+            template <class T, class A, T... Vs>
+            constexpr bool all_equals(batch_constant<T, A, Vs...> c)
+            {
```
serge-sans-paille, line 50: What about:
```
template <class T, class A, T V, T... Vs>
constexpr bool all_equals(batch_constant<T, A, V, Vs...> c)
{
   return (c == std::integral_constant<T, V>{}).all();
}

File: include/xsimd/arch/utils/shifts.hpp [hunk 10]
```diff
@@ -0,0 +1,89 @@
...
+
+                bool out = true;
+                for (std::size_t k = 0; k < sizeof...(Vs); ++k)
+                {
+                    out &= c.get(k) == c.get(0);
+                }
+                return out;
+            }
+
+            template <class T, class T2, class A, T... Vs>
```
serge-sans-paille: `T2` is actually an internal detail, it does not 'escape` the function, nor appear in the output signature. You could compute it as an internal type

include/xsimd/types/xsimd_api.hpp
serge-sans-paille: ack

include/xsimd/types/xsimd_api.hpp
AntoinePrv: I agree, batch_constant versions are tedious because they break full ordering of which dispatch is the best.
We cannot automate which what is better: dispatching to the runtime version, or dispatching to a lower architecture version. The implicit default (because of inheritance) is to dispatch to a lower architecture.
This is especially tricky when a new implementation is added somewhere in the chain, then it can impact the reasoning for other functions not defined in the same file. Hopefully new implementations are not discovered too often.

In this case, I first tried implementing all functions. Coming back to it much later, it became a big effort to track what dispatches to what. This solution was much simpler because it changes the default dispatch from lower architecture to runtime alternative. However I am now realizing this has the same pitfalls. I agree we should ideally stick to the default behavior. If C++17 is not too far away, this could more easily be done with a simplification of overloads.
 
> I'll let it pass this time, but I think we should settle on something better.

Note that I plan to add a symetric API for `bitwise_rshift` for AVX2, and possibly to compute the negative shifts in neon.

In fact we should probably add a placeholder API for batch_constant for all functions, even the ones with no constant implementation. That way if something gets added, users would get the benefit without code changes. Otherwise they need to track their `.as_batch()` calls and which new xsimd constant API was added.

serge-sans-paille: > In fact we should probably add a placeholder API for batch_constant for all functions, even the ones with no constant implementation. That way if something gets added, users would get the benefit without code changes. Otherwise they need to track their .as_batch() calls and which new xsimd constant API was added.

I like this idea. I'm more and more leaning forward code being generated from a description document for all those variants. We could imagine a textual description (say in YAML, but that's an implementation detail) from which we could derive documentation, test cases examples and `xsimd_api.h`.

serge-sans-paille:
LGTM, even if it opens a pandora box I'm not a big fan of. let's say it's a good motivation to come up with a more generic design, as hinted by @AntoinePrv in one of the comment above.

serge-sans-paille:
@AntoinePrv please squash / cleanup the history and we're good

AntoinePrv:
Just need to fix a small dispatch typo and will squash-rebase (got bitten by #1267 during refactoring)

PR #1221, opened 2025-11-25
Fix sfinae for fix-1212

DiamonDinoia:
This addresses https://github.com/xtensor-stack/xsimd/pull/1212#issuecomment-3548448379


serge-sans-paille:
Thanks!

PR #1225, opened 2025-11-28
CMake < 3.10 will be removed in a future release.

atupone:


serge-sans-paille:
Thanks! I have nothing against this change, can you share an official reference pointing at the support drop?

DiamonDinoia:
https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html, Changed in version 3.31: Compatibility with versions of CMake older than 3.10 is deprecated.

serge-sans-paille:
Thanks @DiamonDinoia !

serge-sans-paille:
And of course @atupone

PR #1233, opened 2025-12-22
Fix default values for AVX and AVX512 state depending on OSXSAVE

mhandb:
This pull request updates the initialization of AVX and AVX512 OS support detection in `xsimd_cpuid.hpp`. The change ensures that AVX and AVX512 are only considered enabled if the OS has explicitly enabled OSXSAVE.

Indeed, AVX and AVX512 strictly require the OS to enable OXSAVE otherwise we can face "illegal instruction"
https://wiki.osdev.org/SSE#AVX

For now I am using a local patch, but I would like to avoid to maintain it.

Let me know if this PR is good enough or if it needs any change.


serge-sans-paille:
Thanks :bow:

PR #1239, opened 2025-12-30
Feature/restore riscv support

serge-sans-paille:


serge-sans-paille:
@JohanMabille it's a bit frightening that we missed that :-/

Issue #1232, opened 2025-12-18
Many test failures on aarch64-linux


doronbehar:
Hello, on NixOS we encounter many build errors when trying to update from 13.2.0 to 14.0.0. Here's a full build log:

https://gist.github.com/doronbehar/518771da28f673916997b6757ed4a9e2

serge-sans-paille:
Thanks for reporting, I'll have a look. That's on aarch64 with GCC 14.3, but I can't see the actual compilation flags specific to the build, could you run the build step in verbose mode and share the log so that I can try to reproduce locally? thanks!

doronbehar:
I hope this won't be too exhaustive.. There are roughly 6 levels of verbosity, and I chose the highest verbosity:

https://gist.github.com/doronbehar/c50e32bc02dc5a9bc7da9b25ed820878

ggardet:
There is a similar issue on openSUSE Tumbleweed aarch64 with GCC 15.2.1: https://build.opensuse.org/package/live_build_log/openSUSE:Factory:ARM/xsimd/standard/aarch64

`cmake` command is: 
```
[...]

PR #1241, opened 2026-01-05
[ci] Add gcc 14 validation for arm

serge-sans-paille:
Related to #1232


serge-sans-paille:
Good news, I can reproduce on CI. Now let's understand!

PR #1242, opened 2026-01-08
fix-gcc-10

DiamonDinoia:
On my machine with gcc-10 I get the error:
```
/mnt/home/mbarbone/repos/xsimd/include/xsimd/types/../arch/./xsimd_avx512f.hpp:2740:59: error: '_mm512_cvtsi512_si32' was not declared in this scope; did you mean '_mm512_castsi512_si128'?
 2740 |                 return static_cast<T>(_mm512_cvtsi512_si32(self) & 0xFF);
      |                                       ~~~~~~~~~~~~~~~~~~~~^~~~~~
      |                                       _mm512_castsi512_si128
/mnt/home/mbarbone/repos/xsimd/include/xsimd/types/../arch/./xsimd_avx512f.hpp:2744:59: error: '_mm512_cvtsi512_si32' was not declared in this scope; did you mean '_mm512_castsi512_si128'?
 2744 |                 return static_cast<T>(_mm512_cvtsi512_si32(self) & 0xFFFF);
      |                                       ~~~~~~~~~~~~~~~~~~~~^~~~~~
      |                                       _mm512_castsi512_si128
/mnt/home/mbarbone/repos/xsimd/include/xsimd/types/../arch/./xsimd_avx512f.hpp:2748:59: error: '_mm512_cvtsi512_si32' was not declared in this scope; did you mean '_mm512_castsi512_si128'?
 2748 |                 return static_cast<T>(_mm512_cvtsi512_si32(self));
      |                                       ~~~~~~~~~~~~~~~~~~~~^~~~~~
      |                                       _mm512_castsi512_si128
```
This fixes it for me.

### DIFF
--- a/include/xsimd/arch/common/xsimd_common_swizzle.hpp
+++ b/include/xsimd/arch/common/xsimd_common_swizzle.hpp
@@ -167,6 +167,41 @@ namespace xsimd
                 return cross_impl<0, sizeof...(Vs), sizeof...(Vs) / 2, Vs...>::value;
             }
 
+            /**
+             * @brief Internal: Check if a swizzle pattern crosses lane boundaries
+             *
+             * @tparam LaneSizeBytes Size of a lane in bytes (must be > 0)
+             * @tparam ElemT Element type to determine element size
+             * @tparam U Type of the index values
+             * @tparam Vs... Index values for the swizzle pattern
+             *
+             * @return true if any element accesses data from a different lane
+             *
+             * This is an internal helper. Architecture-specific code can call this directly
+             * with explicit lane sizes (e.g., detail::is_cross_lane_with_lane_size<16, float, ...>()
+             * for 128-bit lanes).
+             */
+            template <std::size_t LaneSizeBytes, typename ElemT, typename U, U... Vs>
+            XSIMD_INLINE constexpr bool is_cross_lane_with_lane_size() noexcept
+            {
+                static_assert(std::is_integral<U>::value, "swizzle mask values must be integral");
+                static_assert(sizeof...(Vs) >= 1, "need at least one value");
+                static_assert(LaneSizeBytes > 0, "lane size must be positive");
+
+                constexpr std::size_t lane_elems = LaneSizeBytes / sizeof(ElemT);
+                constexpr U values[] = { Vs... };
+                constexpr std::size_t N = sizeof...(Vs);
+
+                for (std::size_t i = 0; i < N; ++i)
+                {
+                    std::size_t elem_lane = i / lane_elems;
+                    std::size_t target_lane = static_cast<std::size_t>(values[i]) / lane_elems;
+                    if (elem_lane != target_lane)
+                        return true;
+                }
+                return false;
+            }
+
             template <typename T, T... Vs>
             XSIMD_INLINE constexpr bool is_identity() noexcept { return detail::identity_impl<0, T, Vs...>(); }
             template <typename T, T... Vs>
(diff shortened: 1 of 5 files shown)


serge-sans-paille:
Thanks. It would be great to have a CI reproducer first, could you have a look or do you want me to handle it?

DiamonDinoia:
Is https://github.com/xtensor-stack/xsimd/pull/1243 okay? https://github.com/DiamonDinoia/xsimd/actions/runs/20831227045/job/59845642039

DiamonDinoia:
I could cherry pick the commit here in case you want to merge them in one go

serge-sans-paille:
> I could cherry pick the commit here in case you want to merge them in one go

yes, please do, first the ci commit then the fix :-)

DiamonDinoia:
The fact that it breaks it is a good thing for me. I was chasing bugs in finufft with gcc-10. This might be the cause. https://github.com/flatironinstitute/finufft/pull/780

DiamonDinoia:
```
diff --git a/test/test_shuffle.cpp b/test/test_shuffle.cpp
index b082109..f500072 100644
--- a/test/test_shuffle.cpp
+++ b/test/test_shuffle.cpp
@@ -672,10 +672,8 @@ struct shuffle_test
             }
         };
 
-        std::array<value_type, size> ref_lo;
-        for (size_t i = 0; i < size; ++i)
-            ref_lo[i] = (i & 1) ? rhs[i / 2] : lhs[i / 2];
-        B b_ref_lo = B::load_unaligned(ref_lo.data());
+        // Use zip_lo as a stable reference for the expected interleave.
+        B b_ref_lo = xsimd::zip_lo(b_lhs, b_rhs);
 
         INFO("zip_lo");
         B b_res_lo = xsimd::shuffle(b_lhs, b_rhs, xsimd::make_batch_constant<mask_type, zip_lo_generator, arch_type>());
@@ -689,12 +687,8 @@ struct shuffle_test
             }
         };
 
-        std::array<value_type, size> ref_hi;
-        for (size_t i = 0; i < size; ++i)
-        {
-            ref_hi[i] = (i & 1) ? rhs[size / 2 + i / 2] : lhs[size / 2 + i / 2];
-        }
-        B b_ref_hi = B::load_unaligned(ref_hi.data());
+        // Use zip_hi as a stable reference for the expected interleave.
+        B b_ref_hi = xsimd::zip_hi(b_lhs, b_rhs);
```
the issue seems to be gcc-10 auto vectorization not xsimd explicit one. 

Adding that this code too causes issues: https://marco.godbolt.org/z/bjo6o8TfG

serge-sans-paille:
if that's a GCC issue, specific to gcc-10, we could add something specific, but is that worth the effort?

DiamonDinoia:
gcc-10 has a problem with avx512 but there is also an issue that I introduced when changing the avx512 swizzle: https://github.com/flatironinstitute/finufft/pull/780#issuecomment-3729886876

I should fix that. For gcc-10 if this does not pass CI, I am not sure what is best. I just need xsimd to work with gcc-10 as is not too old of a compiler otherwise I need a #error in finufft if avx512 is used to compile finufft.

test/test_shuffle.cpp
```diff
@@ -672,10 +672,15 @@ struct shuffle_test
             }
         };
 
+#if defined(__GNUC__) && (__GNUC__ == 10) && XSIMD_WITH_AVX512F
```
serge-sans-paille: clang also defines `__GNUC__`, probably not as `__GNUC__ == 10` but `&& !defined (__clang__)` would be safer.

include/xsimd/arch/common/xsimd_common_swizzle.hpp
```diff
@@ -167,6 +167,49 @@ namespace xsimd
...
+                constexpr std::size_t N = sizeof...(Vs);
+                constexpr std::size_t lane_elems = 16 / sizeof(ElemT);
+                return cross_impl128<0, N, lane_elems, U, Vs...>::value;
+            }
+
+            // overload accepting an element type first to compute 128-bit lane size
+            template <typename ElemT, typename U, U... Vs>
+            XSIMD_INLINE constexpr bool is_cross_lane() noexcept
+            {
+                static_assert(std::is_integral<U>::value, "swizzle mask values must be integral");
```
serge-sans-paille: you could probably also static_assert one `sizeof...(U)` to be on the safe side.

include/xsimd/arch/common/xsimd_common_swizzle.hpp
```diff
@@ -167,6 +167,49 @@ namespace xsimd
...
+                static constexpr bool value = false;
+            };
+
+            template <typename ElemT, typename U, U... Vs>
+            XSIMD_INLINE constexpr bool is_cross_lane_128() noexcept
+            {
+                static_assert(sizeof...(Vs) >= 1, "Need at least one lane");
+                constexpr std::size_t N = sizeof...(Vs);
+                constexpr std::size_t lane_elems = 16 / sizeof(ElemT);
+                return cross_impl128<0, N, lane_elems, U, Vs...>::value;
```
serge-sans-paille: Now that we're in C++14, which has better `constexpr` support you could probably write this in a more procedural style, using a temporary array and a for loop to perform the check?

serge-sans-paille:
@DiamonDinoia , this looks good to me, CI is green, can I merge?

DiamonDinoia:
Yes! thanks

serge-sans-paille:
Thank *you*

PR #1243, opened 2026-01-08
adding gcc-10 to CI

DiamonDinoia:


DiamonDinoia:
cherry picked in #1242

PR #1247, opened 2026-01-19
iota batch constant and a few overloads

serge-sans-paille:


serge-sans-paille:
cc @AntoinePrv

serge-sans-paille:
Looks good, but I'll be waiting for @AntoinePrv feedback before landing.

serge-sans-paille:
OK. I'll leave it to you to provide an alias :-)

Issue #611, opened 2021-10-19
MSVC compile error for NEON (win/arm64)


niyas-sait:
XSIMD library cannot be built using MSVC compiler on a windows/arm64 target.

### Steps to reproduce

Visual Studio 2019 or higher is required for compilation.

Compilation can be done natively on a win/arm64 machine or can be cross-compiled from an x86/x64 machine.

From a command terminal

```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvarsx86_arm64.bat

cmake -G "NMake Makefiles" -D DOWNLOAD_GTEST=1 -D CMAKE_BUILD_TYPE=Debug .

nmake test_xsimd
```

The issue has been noticed while building the apache arrow library and would be required to build an apache arrow for win/arm64 using MSVC.

stwiegand:
Any news on this issue, since nowadays ARM64 on Windows is a thing?

JohanMabille:
We should probably revive #612 , but I cannot give a deadline for now. Any help on that would be appreciated.

serge-sans-paille:
proved obsolete by #1248.

jgiannuzzi:
@serge-sans-paille I'm afraid this issue is not proved obsolete by the new CI job. It's about building with the MSVC compiler targeting arm64, not about building with gcc/mingw. Could you please reopen the issue?

serge-sans-paille:
As noted in https://github.com/xtensor-stack/xsimd/pull/1248#issuecomment-3795659200, I mistyped the label, but it's indeed msvc being run.
And this build: https://github.com/xtensor-stack/xsimd/actions/runs/21322768338/job/61375021654 makes me think everything is good?

```
-- The C compiler identification is MSVC 19.44.35222.0
-- The CXX compiler identification is MSVC 19.44.35222.0
```
...
or so I thought, but no, It's targeting amd64, my bad. Reopening

serge-sans-paille:
@jgiannuzzi https://github.com/xtensor-stack/xsimd/actions/runs/21334459960/job/61404179357?pr=1250 from https://github.com/xtensor-stack/xsimd/pull/1250 looks good, right?

jgiannuzzi:
@serge-sans-paille yes that looks good to me! Thank you!

serge-sans-paille:
[...]

PR #1248, opened 2026-01-24
[ci] Setup Windows arm64 runner

serge-sans-paille:


serge-sans-paille:
The fact that this CI passes proves that #611 is no longer a problem - cool!

serge-sans-paille:
reverted and reopened as #1249

Issue #1245, opened 2026-01-12
Additional cpu features detection


AntoinePrv:
Extend the calls to `cpuid` and design a cleaner interface to expose more information not immediately used in this library, but likely relevant for its users, such availability of BMI2, cache sizes... 
Example interface https://learn.microsoft.com/en-us/cpp/intrinsics/cpuid-cpuidex?view=msvc-170.

The context is from Apache Arrow's own detection features not being well maintained https://github.com/apache/arrow/issues/45331.
Alternatives such as https://github.com/google/cpu_features would require adding yet another dependency to Arrow.

This was discussed with @pitrou and @JohanMabille who agree in principle that this addition could have general benefits for users.

What do you think @serge-sans-paille?

serge-sans-paille:
Current implementation is perfectible, I second the idea.

AntoinePrv:
Closing as done, requests for new features can be done in new issues.

PR #1251, opened 2026-01-30
Refactor cpuid

AntoinePrv:
This is a no-addition, non-breaking refactor of existing CPU id features as first class citizen as part of #1245.

- Two individual and simple classes parsing for XCR0 and CPUID
- The classes/header is safe on all platform, avoiding 
- `supported_arch` keeps the same structure and merges use of both class at the moment but both class could be combined a user-friendly `x86_cpu_features`.

### DIFF
--- a/include/xsimd/config/xsimd_cpu_features_x86.hpp
+++ b/include/xsimd/config/xsimd_cpu_features_x86.hpp
@@ -1,0 +1,584 @@
+/***************************************************************************
+ * Copyright (c) Johan Mabille, Sylvain Corlay, Wolf Vollprecht and         *
+ * Martin Renou                                                             *
+ * Copyright (c) QuantStack                                                 *
+ * Copyright (c) Serge Guelton                                              *
+ *                                                                          *
+ * Distributed under the terms of the BSD 3-Clause License.                 *
+ *                                                                          *
+ * The full license is in the file LICENSE, distributed with this software. *
+ ****************************************************************************/
+
+#ifndef XSIMD_CPU_FEATURES_X86_HPP
+#define XSIMD_CPU_FEATURES_X86_HPP
+
+#include <array>
+#include <cassert>
+#include <cstdint>
+#include <cstring>
+
+#include "../utils/bits.hpp"
+#include "./xsimd_config.hpp"
+
+#if XSIMD_TARGET_X86 && defined(_MSC_VER)
+#include <intrin.h> // Contains the definition of __cpuidex
+#endif
+
+namespace xsimd
+{
+    namespace detail
+    {
+        using x86_reg32_t = std::uint32_t;
+
+        using cpuid_reg_t = std::array<x86_reg32_t, 4>;
+
+        /**
+         * CPU Identification (CPUID) instruction results.
+         *
+         * The CPUID instruction provides detailed information about the processor,
+         * including supported instruction set extensions (SSE, AVX, AVX-512, etc.).
+         * This function is well defined on all architectures but will return all zeros
+         * on all non-x86 architectures.
+         *
+         * @param leaf The value inputted to the EAX register.
+         * @param subleaf The value inputted to the ECX register.
+         *
+         * @see https://en.wikipedia.org/wiki/CPUID
+         */
+        inline cpuid_reg_t x86_cpuid(int leaf, int subleaf = 0) noexcept;
+
+        inline x86_reg32_t x86_xcr0_low() noexcept;
+
+        template <typename E>
+        using x86_reg32_bitset = utils::uint_bitset<E, x86_reg32_t>;
+
+        template <x86_reg32_t leaf, x86_reg32_t subleaf, typename A, typename B, typename C, typename D>
+        class x86_cpuid_regs : private x86_reg32_bitset<A>, x86_reg32_bitset<B>, x86_reg32_bitset<C>, x86_reg32_bitset<D>
+        {
+        private:
+            using eax_bitset = x86_reg32_bitset<A>;
+            using ebx_bitset = x86_reg32_bitset<B>;
+            using ecx_bitset = x86_reg32_bitset<C>;
+            using edx_bitset = x86_reg32_bitset<D>;
+
+            /* Parse CPUINFO register value into individual bit components.*/
+            constexpr explicit x86_cpuid_regs(const cpuid_reg_t& regs) noexcept
+                : eax_bitset(regs[0])
+                , ebx_bitset(regs[1])
+                , ecx_bitset(regs[2])
+                , edx_bitset(regs[3])
+            {
+            }
+
+        public:
+            using eax = A;
+            using ebx = B;
+            using ecx = C;
+            using edx = D;
+
+            inline static x86_cpuid_regs read()
+            {
+                return x86_cpuid_regs(detail::x86_cpuid(leaf, subleaf));
+            }
+
+            constexpr x86_cpuid_regs() noexcept = default;
+
+            using eax_bitset::all_bits_set;
+            using ebx_bitset::all_bits_set;
+            using ecx_bitset::all_bits_set;
+            using edx_bitset::all_bits_set;
+        };
+
+        template <typename T>
+        using make_x86_cpuid_regs = x86_cpuid_regs<T::leaf, T::subleaf,
+                                                   typename T::eax,
+                                                   typename T::ebx,
+                                                   typename T::ecx,
+                                                   typename T::edx>;
+    }
+
+    struct x86_cpuid_leaf1_traits
+    {
+        static constexpr detail::x86_reg32_t leaf = 1;
+        static constexpr detail::x86_reg32_t subleaf = 0;
+
+        enum class eax
+        {
+        };
+        enum class ebx
+        {
+        };
+        enum class ecx
+        {
+            /* Streaming SIMD Extensions 3. */
+            sse3 = 0,
+            /* Supplemental Streaming SIMD Extensions 3. */
+            ssse3 = 9,
+            /* Fused multiply-add with 3 operands (FMA3). */
+            fma3 = 12,
+            /* Streaming SIMD Extensions 4.1. */
+            sse4_1 = 19,
+            /* Streaming SIMD Extensions 4.2. */
+            sse4_2 = 20,
+            /* OS has enabled XSAVE/XRSTOR for extended processor state management. */
+            osxsave = 27,
+            /* Advanced Vector Extensions (256-bit SIMD). */
+            avx = 28,
+        };
+        enum class edx
+        {
+            /* Streaming SIMD Extensions 2. */
+            sse2 = 26,
+        };
+    };
+
+    /**
+     * Processor Info and Feature Bits.
+     *
+     * Utility class that can read and parse the registers for the first leaf level
+     * of the CPUID instruction.
+     * This is well defined on all architectures but will return all false on all
+     * non-x86 architectures.
+     *
+     * @see https://en.wikipedia.org/wiki/CPUID
+     */
+    using x86_cpuid_leaf1 = detail::make_x86_cpuid_regs<x86_cpuid_leaf1_traits>;
+
+    struct x86_cpuid_leaf7_traits
+    {
+        static constexpr detail::x86_reg32_t leaf = 7;
+        static constexpr detail::x86_reg32_t subleaf = 0;
+
+        enum class eax
+        {
+        };
+        enum class ebx
+        {
+            /* Advanced Vector Extensions 2 (integer 256-bit SIMD). */
+            avx2 = 5,
+            /* AVX-512 Foundation instructions. */
+            avx512f = 16,
+            /* AVX-512 Doubleword and Quadword instructions. */
+            avx512dq = 17,
+            /* AVX-512 Integer Fused Multiply-Add instructions. */
+            avx512ifma = 21,
+            /* AVX-512 Prefetch instructions. */
+            avx512pf = 26,
+            /* AVX-512 Exponential and Reciprocal instructions. */
+            avx512er = 27,
+            /* AVX-512 Conflict Detection instructions. */
+            avx512cd = 28,
+            /* AVX-512 Byte and Word instructions. */
+            avx512bw = 30,
+        };
+        enum class ecx
+        {
+            /* AVX-512 Vector Bit Manipulation instructions. */
+            avx512vbmi = 1,
+            /* AVX-512 Vector Bit Manipulation instructions 2. */
+            avx512vbmi2 = 6,
+            /* AVX-512 Vector Neural Network instructions. */
+            avx512vnni_bw = 11,
+        };
+        enum class edx
+        {
+        };
+    };
+
+    /**
+     * Extended Feature Bits (EAX=7, ECX=0).
+     *
+     * Utility class that can read and parse the registers for the extended
+     * feature bits leaf of the CPUID instruction.
+     * This is well defined on all architectures but will return all false on all
+     * non-x86 architectures.
+     *
+     * @see https://en.wikipedia.org/wiki/CPUID
+     */
+    using x86_cpuid_leaf7 = detail::make_x86_cpuid_regs<x86_cpuid_leaf7_traits>;
+
+    struct x86_cpuid_leaf7sub1_traits
+    {
+        static constexpr detail::x86_reg32_t leaf = 7;
+        static constexpr detail::x86_reg32_t subleaf = 1;
+
+        enum class eax
+        {
+            /* AVX (VEX-encoded) Vector Neural Network instructions. */
+            avxvnni = 4,
+        };
+        enum class ebx
+        {
+        };
+        enum class ecx
+        {
+        };
+        enum class edx
+        {
+        };
+    };
+
+    /**
+     * Extended Feature Bits (EAX=7, ECX=1).
+     *
+     * Utility class that can read and parse the registers for the extended
+     * feature bits, subleaf 1, of the CPUID instruction.
+     * This is well defined on all architectures but will return all false on all
+     * non-x86 architectures.
+     *
+     * @see https://en.wikipedia.org/wiki/CPUID
+     */
+    using x86_cpuid_leaf7sub1 = detail::make_x86_cpuid_regs<x86_cpuid_leaf7sub1_traits>;
+
+    struct x86_cpuid_leaf80000001_traits
+    {
+        static constexpr detail::x86_reg32_t leaf = 0x80000001;
+        static constexpr detail::x86_reg32_t subleaf = 0;
+
+        enum class eax
+        {
+        };
+        enum class ebx
+        {
+        };
+        enum class ecx
+        {
+            /* AMD Fused multiply-add with 4 operands (FMA4). */
+            fma4 = 16,
+        };
+        enum class edx
+        {
+        };
+    };
+
+    /**
+     * Extended Processor Info and Feature Bits.
+     *
+     * Utility class that can read and parse the registers for the extended
+     * processor info leaf of the CPUID instruction.
+     * This is well defined on all architectures but will return all false on all
+     * non-x86 architectures.
+     *
+     * @see https://en.wikipedia.org/wiki/CPUID
+     */
+    using x86_cpuid_leaf80000001 = detail::make_x86_cpuid_regs<x86_cpuid_leaf80000001_traits>;
+
+    /*
+     * Extended Control Register 0 (XCR0).
+     *
+     * Operating systems can explicitly disable the usage of instruction set (such
+     * as SSE or AVX extensions) by setting an appropriate flag in XCR0 register.
+     * This utility parses such bit values.
+     *
+     * @see https://docs.kernel.org/admin-guide/hw-vuln/gather_data_sampling.html
+     */
+    class x86_xcr0
+    {
+    public:
+        enum class xcr0
+        {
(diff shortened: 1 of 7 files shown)


include/xsimd/xsimd_cpu_features_x86.hpp
```diff
@@ -0,0 +1,307 @@
...
+#include <intrin.h> // Contains the definition of __cpuidex
+#endif
+
+namespace xsimd
+{
+    namespace detail
+    {
+        inline void get_cpuid(int reg[4], int level, int count = 0) noexcept;
+
+        inline std::uint32_t get_xcr0_low() noexcept;
```
serge-sans-paille: it would be good to somehow ensure that this type is the same as `ref_t`

serge-sans-paille: And this could be a private static method for better encapsulation right?

include/xsimd/xsimd_cpu_features_x86.hpp
```diff
@@ -0,0 +1,307 @@
...
+
+        constexpr bool avx512_state_os_enabled() const noexcept
+        {
+            // Check all SSE, AVX, and AVX52 bits even though AVX512 must
+            // imply AVX and SSE
+            return utils::bit_is_set<1, 2, 6>(m_low);
+        }
+
+    private:
+        std::uint32_t m_low = {};
```
serge-sans-paille: should be a `reg_t` ?

include/xsimd/xsimd_cpu_features_x86.hpp
```diff
@@ -0,0 +1,307 @@
...
+     *
+     * @see https://docs.kernel.org/admin-guide/hw-vuln/gather_data_sampling.html
+     */
+    class x86_xcr0
+    {
+    public:
+        using reg_t = std::uint32_t;
+
+        /** Parse a XCR0 value into individual components. */
+        constexpr explicit x86_xcr0(reg_t low) noexcept
```
serge-sans-paille: if `read` is the only way to build this class, this should be `private`

include/xsimd/utils/bits.hpp
```diff
@@ -0,0 +1,40 @@
...
+        template <typename I, typename... Args>
+        constexpr I make_bit_mask(I bit, Args... bits)
+        {
+            return make_bit_mask<I>(bit) | make_bit_mask<I>(static_cast<I>(bits)...);
+        }
+
+        template <int... Bits, typename I>
+        constexpr bool bit_is_set(I value)
+        {
+            constexpr I mask = make_bit_mask<I>(static_cast<I>(Bits)...);
```
serge-sans-paille, line 38: Add a `FIXME: use fold expression once we move to C++17` ?

include/xsimd/xsimd_cpu_features_x86.hpp
```diff
@@ -0,0 +1,307 @@
...
+
+        constexpr bool avxvnni() const noexcept
+        {
+            return utils::bit_is_set<4>(m_regs.reg7a[0]);
+        }
+
+        constexpr bool fma4() const noexcept
+        {
+            return utils::bit_is_set<16>(m_regs.reg8[2]);
+        }
```
serge-sans-paille: Open question: should we have a macro that makes the generation of those field cleaner to the eye?

include/xsimd/xsimd_cpu_features_x86.hpp
```diff
@@ -0,0 +1,307 @@
...
+
+    private:
+        cpu_id_regs m_regs = {};
+    };
+
+    namespace detail
+    {
+        inline void get_cpuid(int reg[4], int level, int count) noexcept
```
serge-sans-paille: Same here: could be private for better encapsulation.

include/xsimd/types/xsimd_all_registers.hpp
```diff
@@ -9,6 +9,8 @@
  * The full license is in the file LICENSE, distributed with this software. *
  ****************************************************************************/
 
+#include "../config/xsimd_inline.hpp"
```
serge-sans-paille: why do you need this?

serge-sans-paille (commented):
Some early review, a few open questions but nothing looks bad here. Way to go!

include/xsimd/types/xsimd_all_registers.hpp
AntoinePrv: It for header including what they use (otherwise I get nasty errors in my editor).
But this one did not belong here but in `xsimd_register.hpp` where `XSIMD_INLINE` is used.

include/xsimd/xsimd_cpu_features_x86.hpp
AntoinePrv: I think it's useful to leave at least a default constructor, so that one can create it somewhere and only conditionally read the value etc.

include/xsimd/xsimd_cpu_features_x86.hpp
AntoinePrv: Either way. I am personally not big on macros. I find it easier read this way (could also be made to fit on one line) and find&replace is quite enough a change is required.

include/xsimd/xsimd_cpu_features_x86.hpp
AntoinePrv: On the contrary, I was wondering whether to make it part of public API.
It's a well defined function, unlikely to change, and that could be useful for users to get features we do not expose (not sure we'll go to cover 100% of cpuid).

include/xsimd/config/xsimd_cpu_features_x86.hpp
```diff
@@ -0,0 +1,311 @@
...
+#if !XSIMD_TARGET_X86
+            (void)level;
+            (void)count;
+            return {}; // All bits to zero
+
+#elif defined(_MSC_VER)
+            __cpuidex(reg.data(), level, count);
+
+#elif defined(__INTEL_COMPILER)
+            __cpuid(reg.data(), level);
```
AntoinePrv: Should we change to use inline ASM for intel compiler? Missing the `count` option here.

include/xsimd/config/xsimd_cpu_features_x86.hpp
```diff
@@ -0,0 +1,311 @@
...
+
+        inline xcr0_reg_t get_xcr0_low() noexcept
+        {
+#if !XSIMD_TARGET_X86
+            return {}; // All bits to zero
+
+#elif defined(_MSC_VER) && _MSC_VER >= 1400
+            return static_cast<xcr0_reg_t>(_xgetbv(0));
+
+#elif defined(__GNUC__)
```
AntoinePrv, line 552: What about `__clang__` and `__INTEL_COMPILER`? Should we reproduce the `get_cpuid` pattern?

AntoinePrv:
@serge-sans-paille what do you think of this?

I wanted to add an `x86_cpu_features` that mixed `xcr0`, `cpuid`, as well as future opinionated features (*e.g.* override `avx2` detection from cpuid if the specific implementation is known not to be performant).
However today that felt very redundant, perhaps in the future if we simplify the logic of `supported_arch`.

include/xsimd/xsimd_cpu_features_x86.hpp
JohanMabille: It looks like the default constructor role is fulfilled by the static `safe_default` method. I would find it cleaner to:
- move the constructor takin a `reg_t`  to the private section.
- keep the `safe_default` method OR replace it with a default constructor.

include/xsimd/config/xsimd_cpu_features_x86.hpp
```diff
@@ -0,0 +1,270 @@
...
+
+            reg_t reg1 = {};
+            reg_t reg7 = {};
+            reg_t reg7a = {};
+            reg_t reg8 = {};
+        };
+
+        /** Parse CpuInfo register values into individual components. */
+        constexpr explicit x86_cpu_id(const cpu_id_regs& regs) noexcept
+            : m_regs(regs)
```
JohanMabille: I think the `cpu_id_regs` struct is implementation detail and should live in the private section. Same for the constructor, objects of this class shoujld only be constructible via the read method  (or this one can become the default constructor instead).

include/xsimd/config/xsimd_cpu_features_x86.hpp
JohanMabille: Yes, I think so. Also, IIRC, `__clang__` comes with `__GNUC__` defined too, and Intel compiler defines `__GNUC__` or `_MSC_VER` depending on the platform. So we may simplify this logic (but first we need to verify that my asumption is right).

serge-sans-paille:
@AntoinePrv please ping me when you need a review!

AntoinePrv:
@serge-sans-paille this is ready. This is intended to make a clean abstraction for CPUID.
In follow-up PRs, I will:
- Add some sort of similar abstraction for arm64
- Propose a generic user-facing interface.

include/xsimd/config/xsimd_cpu_features_x86.hpp
```diff
@@ -0,0 +1,298 @@
...
+         *
+         * AVX and AVX512 strictly require OSXSAVE to be enabled by the OS.
+         * If OSXSAVE is disabled (e.g., via bcdedit /set xsavedisable 1), AVX state won't
+         * be preserved across context switches, so AVX cannot be used.
+         * SSE is therefore the only value safe to assume.
+         */
+        constexpr static x86_xcr0 safe_default() noexcept
+        {
+            reg_t low = {};
+            low = utils::set_bit<static_cast<reg_t>(bit::sse)>(low);
```
serge-sans-paille: Coild be a call to `make_mask` directly, or (better ?) use an overload of `set_bit` with no parameter.

include/xsimd/config/xsimd_cpu_features_x86.hpp
```diff
@@ -0,0 +1,298 @@
...
+        {
+            // Check both SSE and AVX bits even though AVX must imply SSE
+            return bit_is_set<bit::sse, bit::avx>(m_low);
+        }
+
+        constexpr bool avx512_enabled() const noexcept
+        {
+            // Check all SSE, AVX, and AVX512 bits even though AVX512 must
+            // imply AVX and SSE
+            return bit_is_set<bit::sse, bit::avx, bit::zmm_hi256>(m_low);
```
serge-sans-paille: it's a bit strange to have a function named `bit_is_set` that takes several bits as parameter. Maybe `all_bits_set<...>` ?

include/xsimd/config/xsimd_cpu_features_x86.hpp
```diff
@@ -0,0 +1,298 @@
...
+     *
+     * @see https://en.wikipedia.org/wiki/CPUID
+     */
+    class x86_cpu_id
+    {
+    public:
+        /**
+         * Read the CpuId registers from the CPU if on the correct architecture.
+         *
+         * This is only safe to call if bit 18 of CR4.OSXSAVE has been set.
```
serge-sans-paille, line 320: Instead of a comment, you could make this an assert

include/xsimd/config/xsimd_cpu_features_x86.hpp
```diff
@@ -0,0 +1,298 @@
...
+         * @see cpu_id::osxsave
+         */
+        inline static x86_cpu_id read()
+        {
+            cpu_id_regs regs = {};
+            // TODO(C++20): Use designated initializer
+            regs.reg1 = detail::get_cpuid(0x1);
+            regs.reg7 = detail::get_cpuid(0x7);
+            regs.reg7a = detail::get_cpuid(0x7, 0x1);
+            regs.reg8 = detail::get_cpuid(0x80000001);
```
serge-sans-paille: Open question: could we avoid some `get_cpuid` call in some situation? For instance `reg8` is only used for `fma4`, and `reg7` for avx2 or later, which mean we could skip filling `reg7` if we don't have avx, and `fma4` if have `fma3`.

include/xsimd/config/xsimd_cpu_features_x86.hpp
serge-sans-paille: I assumle so, feel free to give it a try!

include/xsimd/config/xsimd_cpu_features_x86.hpp
```diff
@@ -0,0 +1,298 @@
...
+    namespace detail
+    {
+        inline cpuid_reg_t get_cpuid(int level, int count) noexcept
+        {
+            cpuid_reg_t reg = {};
+
+#if !XSIMD_TARGET_X86
+            (void)level;
+            (void)count;
+            return {}; // All bits to zero
```
serge-sans-paille: you could avoid this `return` and let the `return reg` at the end do the job. I wouldn't mind having

```
#if XSIMD_TARGET_X86
// current implementation
#else
inline cpuid_reg_t get_cpuid(int , int ) noexcept {
return {}; // All bits to zero
}
```

include/xsimd/config/xsimd_cpu_features_x86.hpp
serge-sans-paille: It's true that clang defines `__GNUC__`. And it happens to be true for icx too, see https://godbolt.org/z/56j57vhnT

include/xsimd/config/xsimd_cpuid.hpp
```diff
@@ -122,138 +114,39 @@ namespace xsimd
...
-
-#else /* _MSC_VER < 1400 */
-#error "_MSC_VER < 1400 is not supported"
-#endif /* _MSC_VER && _MSC_VER >= 1400 */
-                    return xcr0;
-                };
-
-                auto get_cpuid = [](int reg[4], int level, int count = 0) noexcept
-                {
+                // Safe on all platforms, we simply be false
```
serge-sans-paille: that english sentence is... difficult to understand

include/xsimd/config/xsimd_cpuid.hpp
```diff
@@ -122,138 +114,39 @@ namespace xsimd
...
+                const auto cpuid = xsimd::x86_cpu_id::read();
+                const auto xcr0 = cpuid.osxsave() ? x86_xcr0::read() : x86_xcr0::safe_default();
 
-#if defined(_MSC_VER)
-                    __cpuidex(reg, level, count);
+                sse2 = cpuid.sse2() && xcr0.sse_enabled();
```
serge-sans-paille: some random thoughts: it looks like we actually want `supported_arch` to have a filed member that matches the data structure of `cpuid`, and have a similar layout for `xcr0` so that we could just write something like `*this = cpuid & xcr0` and implement this very efficiently.

include/xsimd/utils/bits.hpp
```diff
@@ -0,0 +1,48 @@
...
+#ifndef XSIMD_CPUID_UTILS_HPP
+#define XSIMD_CPUID_UTILS_HPP
+
+namespace xsimd
+{
+    namespace utils
+    {
+        template <typename I>
+        constexpr I make_bit_mask(I bit)
+        {
```
serge-sans-paille, line 22: you could `static_assert` that `8 * bit < sizeof(I)`, altough I guess the compiler would issue a warning in that case.

include/xsimd/config/xsimd_cpuid.hpp
AntoinePrv: On the contrary, I'm thinking of pushing `x86_cpu_id` `x86_xcr0` even further from any assumptions: become basically just ints with a named enum describing individual bits. Closer to what there is in `x86_xcr0`.

Then have an `x86_cpu_features` with our assumptions and names.

> so that we could just write something like `*this = cpuid & xcr0` and implement this very efficiently.

I'm not sure how that would work. If we have similar named methods, without reflection in C++, we will have to repeat all checks, no?

serge-sans-paille: We can postpone this, it's more of an optimization, so it should not prevent you from moving forward with the restructuring. I'll submit a followup.

include/xsimd/config/xsimd_cpu_features_x86.hpp
serge-sans-paille: Let's postpone this as a further optimization then?

AntoinePrv:
@serge-sans-paille in 1/2 day I can push another commit (easily revertable) with things I've been trying lately.
Basically the idea is to wrap individual register with strongly typed bitset exposed by an enum (essentially what is done for xcr0).

serge-sans-paille:
ok, ping me when you need an extra pair of eyes.

AntoinePrv:
@serge-sans-paille with this last commit we have
- `uint_bitset` is a generic type that wrap any integer + enum into a strongly typed and descriptive bit set
- It's used for all reading all registers
  Raw classes such as `x86_xcr0` or `x86_cpuid_leaf1` can be used as to read and parse registers with strongly typed enums, no more, no less.
- `x86_cpu_features` is an advanced class that combine CPUID results with XCR0 permissions. It is also lazy, reading CPUID registers as needed, including short circuiting if a permission would return false anyways. We can also use the class in the future to override if certain platforms are known to give inaccurate replies.

It's more LOC (though there is more documentation now, and many empty classes / brace lines), but it's also more declarative from now on. We just need to add more members to the enums to read other things.
`x86_cpu_features` on the contrary feels a bit repetitive, though I am not sure if there is a way to intelligently match permissions and cpu features without more boilerplate.

If you like the design I'd be happy to add tests for all classes with know implications (e.g. `sse2` is always true on `x86_64`, `avx2` implies `avx` and `sse4.2`...).

include/xsimd/config/xsimd_cpuid.hpp
```diff
@@ -122,138 +114,38 @@ namespace xsimd
...
+                const auto x86_cpu = xsimd::x86_cpu_features();
 
-#if defined(_MSC_VER)
-                    __cpuidex(reg, level, count);
+                sse2 = x86_cpu.sse2();
+                sse3 = x86_cpu.sse3();
+                ssse3 = x86_cpu.ssse3();
+                sse4_1 = x86_cpu.sse4_1();
+                sse4_2 = x86_cpu.sse4_2();
+                fma3_sse42 = x86_cpu.fma3();
```
serge-sans-paille, line 126: do you think you could  have `supported_arch` inherit from `x86_cpu` to avoid this sequence of assignment?

AntoinePrv: Absolutely, this piece of code was meant as temporary.
Though I planned to this this in follow-up PR, once this and `arm64_cpu_features` is done.

include/xsimd/config/xsimd_cpu_features_x86.hpp
```diff
@@ -0,0 +1,584 @@
...
+            using ebx = B;
+            using ecx = C;
+            using edx = D;
+
+            inline static x86_cpuid_regs read()
+            {
+                return x86_cpuid_regs(detail::x86_cpuid(leaf, subleaf));
+            }
+
+            constexpr x86_cpuid_regs() noexcept = default;
```
JohanMabille, line 84: I'm still not convinced by the need for this default constructor.

include/xsimd/utils/bits.hpp
```diff
@@ -0,0 +1,91 @@
...
+            constexpr I mask = make_bit_mask<I>(static_cast<I>(Bits)...);
+            return (value & mask) == mask;
+        }
+
+        template <int Bit, typename I>
+        constexpr I set_bit(I value)
+        {
+            constexpr I mask = make_bit_mask<I>(static_cast<I>(Bit));
+            return value | mask;
+        }
```
JohanMabille, line 47: Why having these free functions instead of having their implementation in the `uint_bitset` class directly? Do you plan to reuse them elsewhere?

include/xsimd/config/xsimd_cpu_features_x86.hpp
AntoinePrv: IMHO types without a default constructor are a pain to work with.
This default constructor is well defined, it returns all zeros bitsets.
In fact we use it in this very file for the lazy reading in `x86_cpu_features`:
  - Create well-defined but empty `x86_cpuid_regs` when creating `x86_cpu_features`.
  - Replace them as with a "read" version as needed.

I think I understand you point of view, that is may be weird to have the default ctor for a type intended to read a register, that does not do that. Perhpas this is more of an aggregate parser...

include/xsimd/utils/bits.hpp
AntoinePrv: Good point, they came before `uint_bitset`, that is why they are here.
Either way. I could see `constexpr` bit manipulation utilities used in the rest of xsimd instead of magic numbers.

JohanMabille: OK let's keep them for now then.

serge-sans-paille:
LGTM once squashed appropriately. I may do an extra iteration in a follow-up PR, but let's move on.

AntoinePrv:
> LGTM once squashed appropriately. I may do an extra iteration in a follow-up PR, but let's move on.

I can add a similar feature for arm64 that I had started before that and then we can discuss possible changes to `supported_architectures` if you prefer

EDIT: in follow-up PRs

pitrou:
Is this ready for merging?

AntoinePrv:
@serge-sans-paille any remaining issues?

JohanMabille:
Let's get this in and tag a release.

Issue #1252, opened 2026-01-30
Are SVE/I8MM ever detected at runtime?


AntoinePrv:
Related to working #1245, I am wondering if dynamic dispatch to SVE (to name one) is ever possible.

If I understand correctly, for a give function `add` we would like to dispatch, it would get compiled to
- `add_neon.cpp` (assuming we also dispatch it for simplicity here) with `XSIMD_WITH_NEON` (`__ARM_NEON`...)
- `add_sve.cpp` `XSIMD_WITH_SVE` (`__ARM_FEATURE_SVE`...)
- `add_dispatch.cpp` where we use `xsimd::dispatch` with no compilation options in particular (since it needs to run on all platforms).

However in the supported architecture code, getting `sve` value from `getauxval` is gated behind a check for compile-time activation of SVE.
Therefore in `add_dispatch.cpp`, SVE is never detected.

https://github.com/xtensor-stack/xsimd/blob/4d185a6de75ff51fd466064c8d91557ae459105a/include/xsimd/config/xsimd_cpuid.hpp#L115-L119

Am I understanding this correctly @serge-sans-paille @kou ?

kou:
How did you confirm it?

I tried the following program on QEMU:

```c
#include <stdio.h>
#include <sys/auxv.h>

int
main(void)
{
  unsigned long x = getauxval(AT_HWCAP);
#ifdef __ARM_FEATURE_SVE
  printf("__ARM_FEATURE_SVE: Defined\n");
#else
  printf("__ARM_FEATURE_SVE: Not defined\n");
#endif
  printf("%lu: %lu\n", x, x & HWCAP_SVE);
  return 0;
}
```

```console
$ docker run --platform linux/arm64 -it --rm arm64v8/ubuntu:24.04
$ apt update
$ apt install -y gcc
$ gcc -o xxx -O3 -march=armv8-a xxx.c && ./xxx
__ARM_FEATURE_SVE: Not defined
5100273659: 4194304
$ gcc -o xxx -O3 -march=native /host/xxx.c && ./xxx
__ARM_FEATURE_SVE: Defined
5100273659: 4194304
```

It seems that we can detect SVE without enabling SVE in compiling time.

AntoinePrv:
Thanks for the feedback!

> It seems that we can detect SVE without enabling SVE in compiling time.

Exactly what is intended I believe. The issue however is that in `xsimd` the runtime detection is removed if we are not compiling with SVE.

[...]

PR #1253, opened 2026-02-02
Fix runtime detection of SVE

serge-sans-paille:
It should obviously happen even if we don't compile for SVE. That's the goal of runtime detection :-/

Thanks a lot to @AntoinePrv for spotting this.


AntoinePrv (approved):

AntoinePrv:
Fixes #1252

PR #1254, opened 2026-02-02
Slighly less pessimistic detection of neon64

AntoinePrv:
- Detect Neon 64 even without `__ARM_NEON` since arm v8 is sufficient
- Remove hard to track nested `#if #else`


serge-sans-paille:
Clean and concise, thanks.

PR #1256, opened 2026-02-10
Add make_batch_constant from std::array in C++20

AntoinePrv:
Porposal to add `std::array` to `make_batch_constant` when building with C++ 20.


serge-sans-paille:
I'm fine with it, please just add a test case ;-)

serge-sans-paille:
(this requires a C++20 bot, but I'm fine with having this just to make sure we are C++20 compatible too)

AntoinePrv:
I added the test. There is already a CI job for testing with C++17 and C++20.

PR #1265, opened 2026-02-26
Prevent some invalid associativity in to_float implementation for avx…

### LINKED ISSUE #1264: Incorrect result for log() on AVX with `-ffast-math`
Here's the relevant code excerpt:

```cpp
LOG_EXPR(Arch::name());
auto batch = getBatch(expr.a, index);
LOG_EXPR(batch);
LOG_EXPR(xsimd::log(batch));
LOG_EXPR(batch.size);
```

And the dodgy output:
```
Arch::name() = avx2
batch = (0.60455, 0.635409, 0.891392, 0.871854)
xsimd::log(batch) = (0.189876, 0.23966, -0.114971, -0.137133)
batch.size = 4
```

The first two results are pretty wrong (they should be `-0.50327` and `-0.453487`).

This only happens for me when using `xsimd::avx` and compiling with `-mavx -ffast-math`.  If it's just `-mavx` alone then I don't get the issue.  It might happen for other architectures as well, AVX is just the one I have access to.

Any ideas?  I had a look through the code for `log()`, but couldn't see how the fast-math-specific bits would give this result.

serge-sans-paille:
…2 under -ffast-math

Fix  #1264


serge-sans-paille:
@DiamonDinoia I've cherry-picked your patch, removed the redundant guard and pushed it here

DiamonDinoia:
Sure, it looks good!

PR #1266, opened 2026-03-04
fix(neon): unsigned bitwise shifts are never called

AntoinePrv:
Shift public API expects the same type for data and shifts:

https://github.com/xtensor-stack/xsimd/blob/c3a8d37db2777c65c889e49143af36f32746d3da/include/xsimd/types/xsimd_api.hpp#L371

However, the neon implementations for unsigned data take the shifts as signed (likely because the underlying intrinsict use signed shifts).

https://github.com/xtensor-stack/xsimd/blob/c3a8d37db2777c65c889e49143af36f32746d3da/include/xsimd/arch/xsimd_neon.hpp#L2385

As a result the neon overload is never selected, falling back on `common` scalar implementation.

## Example

```cpp
#include <cstdint>
#include <xsimd/xsimd.hpp>

auto f(xsimd::batch<std::uint16_t> const& a, xsimd::batch<std::uint16_t> const& b)
{
    return xsimd::bitwise_rshift(a, b);
}
```

Before:
```asm
ldr     q0, [x0]
ldr     q1, [x1]
ushll.4s        v2, v0, #0
ushll2.4s       v0, v0, #0
ushll.4s        v3, v1, #0
movi.2d v4, #0000000000000000
usubw2.4s       v1, v4, v1
ushl.4s v0, v0, v1
neg.4s  v1, v3
ushl.4s v1, v2, v1
uzp1.8h v0, v1, v0
ret
```

After:
```asm
ldr     q0, [x0]
ldr     q1, [x1]
neg.8h  v1, v1
ushl.8h v0, v0, v1
ret
```

I also grep with `-O0` and the assembly was previously pointing to `common` as a `requires_arch` in the third parameter.

### DIFF
--- a/include/xsimd/arch/xsimd_neon.hpp
+++ b/include/xsimd/arch/xsimd_neon.hpp
@@ -2374,14 +2374,12 @@ namespace xsimd
             }
 
             template <class A, class T>
-            XSIMD_INLINE bool all_positive(batch<T, A> const& b) noexcept
+            XSIMD_INLINE bool shifts_all_positive(batch<T, A> const& b) noexcept
             {
-                for (std::size_t k = 0; k < b.size; ++k)
-                {
-                    if (b.get(k) < 0)
-                        return false;
-                }
-                return true;
+                std::array<T, batch<T, A>::size> tmp = {};
+                b.store_unaligned(tmp.begin());
+                return std::all_of(tmp.begin(), tmp.end(), [](T x)
+                                   { return x >= 0; });
             }
         }
 
@@ -2397,7 +2395,7 @@ namespace xsimd
         XSIMD_INLINE batch<T, A> bitwise_lshift(batch<T, A> const& lhs, batch<T, A> const& rhs, requires_arch<neon>) noexcept
         {
             // Blindly converting to signed since out of bounds shifts are UB anyways
-            assert(detail::all_positive(rhs));
+            assert(detail::shifts_all_positive(rhs));
             return vshlq_u8(lhs, vreinterpretq_s8_u8(rhs));
         }
 
@@ -2411,7 +2409,7 @@ namespace xsimd
         XSIMD_INLINE batch<T, A> bitwise_lshift(batch<T, A> const& lhs, batch<T, A> const& rhs, requires_arch<neon>) noexcept
         {
             // Blindly converting to signed since out of bounds shifts are UB anyways
-            assert(detail::all_positive(rhs));
+            assert(detail::shifts_all_positive(rhs));
             return vshlq_u16(lhs, vreinterpretq_s16_u16(rhs));
         }
 
@@ -2425,7 +2423,7 @@ namespace xsimd
         XSIMD_INLINE batch<T, A> bitwise_lshift(batch<T, A> const& lhs, batch<T, A> const& rhs, requires_arch<neon>) noexcept
         {
             // Blindly converting to signed since out of bounds shifts are UB anyways
-            assert(detail::all_positive(rhs));
+            assert(detail::shifts_all_positive(rhs));
             return vshlq_u32(lhs, vreinterpretq_s32_u32(rhs));
         }
 
@@ -2439,7 +2437,7 @@ namespace xsimd
         XSIMD_INLINE batch<T, A> bitwise_lshift(batch<T, A> const& lhs, batch<T, A> const& rhs, requires_arch<neon>) noexcept
         {
             // Blindly converting to signed since out of bounds shifts are UB
-            assert(detail::all_positive(rhs));
+            assert(detail::shifts_all_positive(rhs));
             return vshlq_u64(lhs, vreinterpretq_s64_u64(rhs));
         }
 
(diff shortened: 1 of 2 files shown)


AntoinePrv:
## Note on 64 bits

The fact that bitwise_lshift for 64 bits is in `neon` and bitwise_rshift for 64 bit sits alone in `neon64` also seems suspicious.

include/xsimd/arch/xsimd_neon.hpp
```diff
@@ -2382,9 +2382,10 @@ namespace xsimd
         }
 
         template <class A, class T, detail::enable_sized_unsigned_t<T, 1> = 0>
-        XSIMD_INLINE batch<T, A> bitwise_lshift(batch<T, A> const& lhs, batch<as_signed_integer_t<T>, A> const& rhs, requires_arch<neon>) noexcept
+        XSIMD_INLINE batch<T, A> bitwise_lshift(batch<T, A> const& lhs, batch<T, A> const& rhs, requires_arch<neon>) noexcept
         {
-            return vshlq_u8(lhs, rhs);
+            // Blindly converting to signed since out of bounds shifts are UB anyways
```
serge-sans-paille, line 2397: maybe just add an assert?

include/xsimd/arch/xsimd_neon.hpp
```diff
@@ -2394,9 +2395,10 @@ namespace xsimd
         }
 
         template <class A, class T, detail::enable_sized_unsigned_t<T, 2> = 0>
-        XSIMD_INLINE batch<T, A> bitwise_lshift(batch<T, A> const& lhs, batch<as_signed_integer_t<T>, A> const& rhs, requires_arch<neon>) noexcept
+        XSIMD_INLINE batch<T, A> bitwise_lshift(batch<T, A> const& lhs, batch<T, A> const& rhs, requires_arch<neon>) noexcept
         {
-            return vshlq_u16(lhs, rhs);
+            // Blindly converting to signed since out of bounds shifts are UB anyways
```
serge-sans-paille, line 2411: same here

serge-sans-paille (approved):
LGTM with asserts added if you feel like doing so

serge-sans-paille:
> ## Note on 64 bits
> 
> The fact that bitwise_lshift for 64 bits is in `neon` and bitwise_rshift for 64 bit sits alone in `neon64` also seems suspicious.

What puzzles me is... how is

```
vshlq_u64(lhs, vnegq_s64(rhs));
```

an implementation of a right shift

serge-sans-paille:
> > ## Note on 64 bits
> > The fact that bitwise_lshift for 64 bits is in `neon` and bitwise_rshift for 64 bit sits alone in `neon64` also seems suspicious.
> 
> What puzzles me is... how is
> 
> ```
> vshlq_u64(lhs, vnegq_s64(rhs));
> ```
> 
> an implementation of a right shift

And the answer is because left shift seems to support shifting right if the operand is negative, fine.
And it's in neon64 because `vnegq_s64` requires neon64. But we could emulate this on neon32 appropriately, I'll do that.

serge-sans-paille:
> And the answer is because left shift seems to support shifting right if the operand is negative, fine. And it's in neon64 because `vnegq_s64` requires neon64. But we could emulate this on neon32 appropriately, I'll do that.

see https://github.com/xtensor-stack/xsimd/pull/1268

AntoinePrv:
@serge-sans-paille all good now

include/xsimd/arch/xsimd_neon.hpp
```diff
@@ -2371,6 +2372,17 @@ namespace xsimd
                     return bitwise_lshift(lhs, n, ::xsimd::detail::int_sequence<Is...>());
                 }
             }
+
+            template <class A, class T>
+            XSIMD_INLINE bool all_positive(batch<T, A> const& b) noexcept
```
serge-sans-paille: It's just `xsimd::all(b >= 0)`

AntoinePrv: That does not work because `kernel::all` on which it depends is declared below.
Would also require compare utilities to be declared.

This is not the only case where I find it hard to reuse other functions, with possibly circular include loops.
With all the SFINAE, forward declaring all of them is not very maintainable.
(In C++17 we could get rid of all SFINAE in favor of `if constexpr` and have simple signatures that we can easily forward declare).

serge-sans-paille: I agree with the C++17 simplification. But that's for another day. meanwhile you could at least store to a buffer and call `std::all_of` ?

AntoinePrv: @serge-sans-paille I'm happy to do the changes, but not sure it will be for the best.
What is the objective? Less assertion code? No extra visible symbol (`all_positive`)?

If we want to get rid of `all_positive` then we would need at least this in all functions. Possibly more with `#ifdef NDEBUG` to avoid warnings given `[[maybe_unused]]` is C++17.
```c++
std::array<T, batch<T, A>> tmp = {};
rhs.store(tmp.data());
assert(std::all_of(tmp.begin(), tmp.end(), [](auto x) { return x >= 0; }));
```

Or do you mean to use `std::all_of` inside `all_positive`?

serge-sans-paille:
Thanks!

PR #1276, opened 2026-03-19
fix: harden reassociation barriers for fast-math

DiamonDinoia:
There is a bug in nearbyint `-fassociative-math` breaks it as it does not define `__FAST_MATH__`.
Also the barrier used was causing a stack spill. 

I centralized a barrier function that we can use everywhere in the code and used it in the places I know it helps. 

Now we can just use `reassociation_barrier` to avoid compiler reordering of instructions. 

Let me know if you like the internal API and if you need changes to it. I find that this is the solution that minimizes ifdef boilerplate and allows to dispatch to all archs. (With c++17 this will be simpler).

Cheers,
Marco

### DIFF
--- a/include/xsimd/arch/common/xsimd_common_details.hpp
+++ b/include/xsimd/arch/common/xsimd_common_details.hpp
@@ -111,6 +111,69 @@ namespace xsimd
 
         namespace detail
         {
+            // Prevent -ffast-math from reassociating floating-point
+            // arithmetic across this point.  The reason string
+            // documents *why* at each call site; unused at runtime.
+            //
+            // Zero-cost register constraints per target:
+            //   x86  "+x"  (XMM/YMM/ZMM, also scalar float/double)
+            //   ARM  "+w"  (V-reg / SVE Z-reg, also scalar float/double)
+            //   PPC  "+wa" (VS register, also scalar float/double)
+            //   RISC-V "+f" (F/D register, scalar float/double)
+            //   RISC-V RVV "+vr" (V register; GCC 15+ / Clang 20+)
+            //
+            // On unknown targets the "+m" fallback spills; it is
+            // only emitted when the compiler can actually reassociate.
+            template <class T>
+            XSIMD_INLINE void reassociation_barrier(T& x, const char*) noexcept
+            {
+#if XSIMD_REASSOCIATIVE_MATH && XSIMD_WITH_INLINE_ASM && !defined(__EMSCRIPTEN__)
+#if XSIMD_WITH_SSE2
+                __asm__ volatile("" : "+x"(x));
+#elif XSIMD_WITH_NEON || XSIMD_WITH_SVE
+                __asm__ volatile("" : "+w"(x));
+#elif XSIMD_WITH_VSX
+                __asm__ volatile("" : "+wa"(x));
+#else
+                __asm__ volatile("" : "+m"(x));
+#endif
+#else
+                (void)x;
+#endif
+            }
+
+            // RISC-V scalar float/double: use F/D registers instead of
+            // spilling through "+m".  These overloads also serve
+            // emulated batches on RISC-V via the std::array overload.
+#if XSIMD_REASSOCIATIVE_MATH && XSIMD_WITH_INLINE_ASM && defined(__riscv)
+            XSIMD_INLINE void reassociation_barrier(float& x, const char*) noexcept
+            {
+                __asm__ volatile("" : "+f"(x));
+            }
+            XSIMD_INLINE void reassociation_barrier(double& x, const char*) noexcept
+            {
+                __asm__ volatile("" : "+f"(x));
+            }
+#endif
+
+            template <class T, size_t N>
+            XSIMD_INLINE void reassociation_barrier(std::array<T, N>& arr, const char* reason) noexcept
+            {
+                for (auto& v : arr)
+                    reassociation_barrier(v, reason);
+            }
+
+            template <class T, class A>
+            XSIMD_INLINE void reassociation_barrier(batch<T, A>& b, const char* reason) noexcept
+            {
+#if XSIMD_REASSOCIATIVE_MATH && XSIMD_WITH_RVV && XSIMD_WITH_INLINE_ASM && ((__GNUC__ >= 15) || (__clang_major__ >= 20))
+                __asm__ volatile("" : "+vr"(b.data.value.value));
+                (void)reason;
+#else
+                reassociation_barrier(b.data, reason);
+#endif
+            }
+
             template <class F, class A, class T, class... Batches>
             XSIMD_INLINE batch<T, A> apply(F&& func, batch<T, A> const& self, batch<T, A> const& other) noexcept
             {
--- a/include/xsimd/arch/xsimd_common_fwd.hpp
+++ b/include/xsimd/arch/xsimd_common_fwd.hpp
@@ -101,6 +101,11 @@ namespace xsimd
         // Forward declarations for pack-level helpers
         namespace detail
         {
+            template <class T>
+            XSIMD_INLINE void reassociation_barrier(T& x, const char*) noexcept;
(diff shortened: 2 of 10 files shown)


DiamonDinoia:
@serge-sans-paille this is also ready for review.

.github/workflows/windows.yml
```diff
@@ -92,6 +92,70 @@ jobs:
...
+    - name: Setup
+      run: |
+        mkdir _build
+        cd _build && cmake .. -DBUILD_TESTS=ON -DDOWNLOAD_DOCTEST=ON -DBUILD_BENCHMARK=ON -DBUILD_EXAMPLES=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_CXX_FLAGS="/arch:AVX2" -G Ninja
+    - name: Build
+      run: |
+        cd _build && cmake --build .
+    - name: Testing xsimd
+      run: |
+        cd _build && ./test/test_xsimd
```
serge-sans-paille: It would be nice to split that part in a separate commit (even PR!). It's independent from the fast-math issue, right?

.github/workflows/windows.yml
```diff
@@ -92,6 +92,70 @@ jobs:
...
+        python3 -m pip install ninja
+    - name: Checkout xsimd
+      uses: actions/checkout@v3
+    - name: Setup
+      run: |
+        mkdir _build
+        cd _build && cmake .. -DBUILD_TESTS=ON -DDOWNLOAD_DOCTEST=ON -DBUILD_BENCHMARK=ON -DBUILD_EXAMPLES=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_CXX_FLAGS="/arch:AVX2" -G Ninja
+    - name: Build
+      run: |
+        cd _build && cmake --build .
```
serge-sans-paille: nit: could be `cmake --build -B _build`

.github/workflows/windows.yml
```diff
@@ -92,6 +92,70 @@ jobs:
...
+      with:
+        arch: amd64
+    - name: Check clang-cl
+      run: |
+        command -v clang-cl
+        clang-cl --version
+    - name: Setup Ninja
+      run: |
+        python3 -m pip install --upgrade pip setuptools wheel
+        python3 -m pip install ninja
```
serge-sans-paille: given the redundancy between that job and the one above, it seems better to use github strategy.

include/xsimd/arch/xsimd_common_fwd.hpp
```diff
@@ -101,6 +110,12 @@ namespace xsimd
         // Forward declarations for pack-level helpers
         namespace detail
         {
+            template <class T>
+            XSIMD_INLINE void reassociation_barrier(T& x, memory_barrier_tag) noexcept;
+
+            template <class T, class A>
+            XSIMD_INLINE void reassociation_barrier(T& x, A const&) noexcept;
```
serge-sans-paille: Why this dispatching based on the architecture?

serge-sans-paille: Based on the usage, I wonder if we should enforce a string literal as last argument, that explains the reason why we generate that barrier. the string would be ignored, but it enforces the documentation.

serge-sans-paille (changes requested):

include/xsimd/arch/xsimd_common_fwd.hpp
DiamonDinoia: Because arm/neon have different asm instructions or might not be supported at all. I just thought this would be a way to avoid ifdefs. Potentially reassociation_barrier can be guarded by xsimd_with_arch and we get a different function with different ifdefs. I have not investigated where to put such function (in which header) and how the code looks like. This felt xsimd idiomatic. 

I am okay with a [[maybe_unused]] string literal argument to this function for documentation purposes.

.github/workflows/windows.yml
DiamonDinoia: yes we can, though the fast math part tough will fail without this PR.

DiamonDinoia:
I think I can simplify this a bit more. A is not needed anymore if I we go this route.

include/xsimd/arch/common/xsimd_common_details.hpp
```diff
@@ -111,6 +111,48 @@ namespace xsimd
...
+            // Uses the tightest register-class constraint for the target so
+            // the value stays in its native SIMD register (no spill):
+            //   x86  (SSE/AVX/AVX-512) : "+x"  – XMM / YMM / ZMM
+            //   ARM  (NEON / SVE)      : "+w"  – vector / SVE Z-reg
+            //   PPC  (VSX)             : "+wa" – VS register
+            //   other / MSVC           : address + memory clobber (fallback)
+            template <class T>
+            XSIMD_INLINE void reassociation_barrier(T& x, const char*) noexcept
+            {
+#if XSIMD_WITH_INLINE_ASM
```
serge-sans-paille: Shouldn't we make this empty if we're not under fast-math?

DiamonDinoia:
Well, if we are not under fast math this should essentially be a no op.
Also, fast math does not detect if only associative math is enabled . Which
also breaks it. Now it becames checking all flags of the compiler that can
reorder floating points.

serge-sans-paille:
Well, I don't think we can say that

```
__asm__ volatile("" : : "r"(&x) : "memory");
```

is a no-op. Should we allow for user-defined `XSIMD_REASSOCIATIVE` which would be forced by `__FAST_MATH__` but could also be set by the user?
I don't want fast-math to impact non-fast math code generation.

DiamonDinoia:
Let me think about this for a second. MSVC does not allow inline asm. I should update the comment there. There might be better alternatives for scalars, WASM, and RISC-V that don't impact performance.

DiamonDinoia:
I tried to have the no op in most cases and the spill asm is guarded with macros. I think is the best solution as users should not worry about FP reordering in practice, except corner cases. We could document those if needed.

DiamonDinoia:
Hi @serge-sans-paille what do you think now?

DiamonDinoia:
This also supersedes #550 and #551

test/test_xsimd_api.cpp
```diff
@@ -591,12 +591,16 @@ struct xsimd_api_float_types_functions
     void test_exp()
     {
         value_type val(2);
+#ifdef __FAST_MATH__
```
serge-sans-paille: or XSIMD_REASSOCIATIVE_MATH ?

include/xsimd/arch/common/xsimd_common_details.hpp
```diff
@@ -111,6 +111,72 @@ namespace xsimd
...
+            //   PPC  "+wa" (VS register, also scalar float/double)
+            //   RISC-V "+f" (F/D register, scalar float/double)
+            //   RISC-V RVV "+vr" (V register; GCC 15+ / Clang 20+)
+            //
+            // On unknown targets the "+m" fallback spills; it is
+            // guarded by XSIMD_REASSOCIATIVE_MATH so it is only
+            // emitted when the compiler can actually reassociate.
+            template <class T>
+            XSIMD_INLINE void reassociation_barrier(T& x, const char*) noexcept
+            {
```
serge-sans-paille, line 129: I really would rather have the whole body be guarded by `XSIMD_REASSOCIATIVE_MATH`.

I would love having a section on fast-math in the manual, where the documentation of that variable would be showcased. Can be in a follow-up commit though.

DiamonDinoia: I don't love this idea. But okay.

serge-sans-paille:
Thanks!

PR #1280, opened 2026-03-26
Update README after C++14 requirement

DiamonDinoia:
Removes the outdated note telling users to enable C++14 manually, since the requirement is now published through the CMake interface.


JohanMabille (approved):

JohanMabille:
Thanks!

PR #1285, opened 2026-04-01
Enable MSVC (cl.exe) build on Windows ARM64

pdeep854:
# PR: Enable MSVC (cl.exe) build on Windows ARM64

## Summary

This PR restores full build compatibility for xsimd with the MSVC compiler (`cl.exe`) on
Windows ARM64 (`_M_ARM64`). Prior to this change, the project failed to compile with MSVC
on ARM64 due to fundamental differences in how MSVC exposes ARM NEON intrinsics compared
to GCC and Clang.

All changes are **strictly additive and backward-compatible**: every existing GCC/Clang code
path is preserved unchanged inside `#else` branches. No behaviour is altered for any
non-MSVC-ARM64 target.

## Testing

The build was verified locally on a Windows ARM64 machine using Visual Studio 2022:

```cmd
cmake .. -DBUILD_TESTS=ON -DDOWNLOAD_DOCTEST=ON -DBUILD_BENCHMARK=ON ^
         -DBUILD_EXAMPLES=ON -DCMAKE_BUILD_TYPE=Release ^
         -G "Visual Studio 17 2022"
cmake --build . --config Release
```

**Result:** All three targets build successfully with zero errors:
```
benchmark_xsimd.vcxproj -> ...\Release\benchmark_xsimd.exe
mandelbrot.vcxproj      -> ...\Release\mandelbrot.exe
test_xsimd.vcxproj      -> ...\Release\test_xsimd.exe
```

image

PR #1288, opened 2026-04-03
Fix windows CI arm runner

AntoinePrv:


serge-sans-paille:
There's a conflict with #1285 . I prefer your approach as it's more modern C++, but let's be polite and also honor the author of #1285 who already put effort into this. Would you be ok to wait for me to merge their PR, and then you'd update it with yours?

AntoinePrv:
@serge-sans-paille  sorry, I did not saw the other PR. I noticed the windows-arm was not running ARM in #1287.
The author can also pick from this PR if they want, either is fine for me.

From a CI perspective, it seems the Github action `ilammy/msvc-dev-cmd` can be called with `arm64`, which minimize the workflow changes.

PR #1290, opened 2026-04-04
fix: avoid per-instantiation vtable issues with MinGW GCC

DiamonDinoia:
Replace INFO() macro calls with concrete StringContextScope class to work around MinGW GCC generating a unique vtable per template instantiation. INFO() creates ContextScope<Lambda> which triggers this issue; the concrete class has a single vtable definition shared across all instantiations.

- Add StringContextScope struct inheriting from doctest::detail::ContextScopeBase
- Add make_context_info() helper to create context without template lambdas
- Update CHECK_BATCH_EQ, CHECK_SCALAR_EQ, CHECK_VECTOR_EQ macros to use new helper

### DIFF
--- a/test/test_utils.hpp
+++ b/test/test_utils.hpp
@@ -386,27 +386,47 @@ namespace detail
         b.store_unaligned(dst.data() + i);
     }
 
+    // Non-template context scope to avoid per-instantiation vtable issues with MinGW GCC.
+    // INFO() creates a ContextScope<Lambda> with a unique vtable per template instantiation.
+    // This concrete class has a single vtable definition shared across all instantiations.
+    struct StringContextScope : doctest::detail::ContextScopeBase
+    {
+        std::string msg_;
+        explicit StringContextScope(std::string msg)
+            : msg_(std::move(msg))
+        {
+        }
+        void stringify(std::ostream* os) const override { *os << msg_; }
+    };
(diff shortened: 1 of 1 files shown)


test/test_utils.hpp
```diff
@@ -386,27 +386,46 @@ namespace detail
...
-#define CHECK_BATCH_EQ(b1, b2)                            \
-    do                                                    \
-    {                                                     \
-        INFO(#b1 ":", b1);                                \
-        INFO(#b2 ":", b2);                                \
-        CHECK_UNARY(::detail::expect_batch_near(b1, b2)); \
+#define CHECK_BATCH_EQ(b1, b2)                             \
+    do                                                     \
+    {                                                      \
+        auto _ctx1 = ::detail::make_context_info(#b1, b1); \
```
serge-sans-paille, line 414: please add a small comment here and below to explain why we're doing something smart :-)

serge-sans-paille:
the error looks unrelated? o_O

DiamonDinoia:
The failure seems that in test/test_shuffle.cpp:623 — a -Werror=uninitialized error in the transpose() test for batch<short int> on AVX512 with GCC 13. It's complaining about b_matrix being used uninitialized. I don't think this is related to this PR.

PR #1293, opened 2026-04-09
[test] Simplify shuffle_test while pleasing gcc static analysis

serge-sans-paille:


test/test_shuffle.cpp
```diff
@@ -621,12 +621,10 @@ struct shuffle_test
     {
         B b_lhs = B::load_unaligned(lhs.data());
         std::array<B, size> b_matrix;
```
DiamonDinoia: b_matrix{} and ref_matrix{} would please gcc

test/test_shuffle.cpp
```diff
@@ -621,12 +621,10 @@ struct shuffle_test
...
-        for (size_t i = 0; i < size; ++i)
-            b_matrix[i] = b_lhs;
+        std::fill(b_matrix.begin(), b_matrix.end(), b_lhs);
         std::array<value_type, size * size> ref_matrix;
         for (size_t i = 0; i < size; ++i)
-            for (size_t j = 0; j < size; ++j)
-                ref_matrix[i * size + j] = lhs[i];
+          std::fill(&ref_matrix[i*size], &ref_matrix[(i + 1)*size], lhs[i]);
```
DiamonDinoia: I would do 
auto ref_base = ref_matrix.data();
`std::fill(ref_base+i*size, ref_base+(i + 1)*size, lhs[i]);`

I personally don't like using &() when we have access to the pointer.

DiamonDinoia:
Weird, I cannot close the conversations. They are addressed so all good from me :)

serge-sans-paille:
@DiamonDinoia can you check the mingw issue? It seems to be related to #1290

DiamonDinoia:
I think that on MINGW32 (32-bit x86, GCC 15.2.0), the i386 ABI only guarantees 4-byte stack alignment, but SSE types (__m128i etc.) require 16-byte alignment. GCC normally inserts stack realignment prologues when it detects SSE usage, but std::fill on an array of SIMD batch types may generate different code paths (e.g., vectorized stores, rep instructions, or aligned movaps) compared to the manual loop which GCC likely compiled with unaligned movdqu stores. If GCC's inlined std::fill emits aligned SSE stores on a misaligned stack frame, that's a segfault.

I am not sure. Try:

```cpp 
  std::array<B, size> b_matrix = {};                                                                                                                              
  for (size_t i = 0; i < size; ++i)
      b_matrix[i] = b_lhs;

  std::array<value_type, size * size> ref_matrix = {};                                                                                                            
  auto ref_matrix_iter = ref_matrix.begin();
  for (size_t i = 0; i < size; ++i, ref_matrix_iter += size)                                                                                                      
      std::fill(ref_matrix_iter, ref_matrix_iter + size, lhs[i]);                                                                                                  
```

AntoinePrv:
At this rate, maybe the CI will be done by Monday 😓

serge-sans-paille:
I fear the avx512 CI is now hanging in test execution :-/

serge-sans-paille:
I may revert some commits if we end up breaking CI that much... I'll do that in another PR though.

AntoinePrv:
I don't think it comes from the changes. If I recall, yesterday my PR was massively hanging too, including 2h on a download.

AntoinePrv:
@serge-sans-paille @JohanMabille I think the CI is back in shape if you want to restart this.

AntoinePrv:
Wow, I relaunched the tests, same time out 🤯

PR #1294, opened 2026-04-09
Add xsimd::get<>() for optimized compile-time element extraction

DiamonDinoia:
Add a free function xsimd::get<I>(batch) API mirroring std::get<I>(tuple) for fast compile-time element extraction from SIMD batches.

Per-architecture optimized kernel::get overloads using the fastest available intrinsics:
- SSE2: shuffle/shift + scalar convert
- SSE4.1: pextrd/pextrq/pextrb/pextrw, bitcast + pextrd for float
- AVX: vextractf128/vextracti128 + SSE4.1 delegate
- AVX-512: vextracti64x4/vextractf32x4 + AVX delegate
- NEON: vgetq_lane_* (single instruction for all types)
- NEON64: vgetq_lane_f64

Also fixes a latent bug in the common fallback for complex batch compile-time get (wrong buffer type).

### DIFF
--- a/test/test_batch.cpp
+++ b/test/test_batch.cpp
@@ -158,6 +158,21 @@ struct batch_test
         CHECK_EQ(res.first(), lhs[0]);
     }
 
+    template <size_t... Is>
+    void test_get_impl(batch_type const& res, std::index_sequence<Is...>) const
+    {
+        array_type extracted = { xsimd::get<Is>(res)... };
+        CHECK_EQ(extracted, lhs);
+        CHECK_BATCH_EQ(batch_type::load_unaligned(extracted.data()), res);
+    }
+
+    void test_get() const
+    {
+        batch_type res = batch_lhs();
+        CHECK_EQ(xsimd::get<0>(res), res.first());
+        test_get_impl(res, std::make_index_sequence<size> {});
+    }
+
     void test_arithmetic() const
     {
         // +batch
@@ -984,6 +999,11 @@ TEST_CASE_TEMPLATE("[batch]", B, BATCH_TYPES)
     SUBCASE("first element")
     {
         Test.test_first_element();
+    }
+
+    SUBCASE("get")
+    {
+        Test.test_get();
     }
 
     SUBCASE("arithmetic")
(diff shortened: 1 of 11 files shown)


DiamonDinoia:
Nice thanks for fixing CI!

This is ready for review. Once approved I will rewrite the history. I don't want to trigger a useless CI run.

test/test_batch_complex.cpp
```diff
@@ -182,6 +182,26 @@ struct batch_complex_test
...
+    void check_get_element(batch_type const& res) const
+    {
+        CHECK_EQ(xsimd::get<I>(res), lhs[I]);
+    }
+
+    template <size_t... Is>
+    void check_get_all(batch_type const& res, std::index_sequence<Is...>) const
+    {
+        int dummy[] = { (check_get_element<Is>(res), 0)... };
+        (void)dummy;
```
serge-sans-paille: you could check that loading the generated array ends up being equal to `res`, right?

serge-sans-paille (changes requested):

serge-sans-paille (changes requested):
Please fix the testing so that we have a decent confidence in the getter when index != 0

DiamonDinoia:
> Please fix the testing so that we have a decent confidence in the getter when index != 0

Yes, I will! I also noticed some small changes I should make. I just did not have time to get to this still.

DiamonDinoia:
I like how it is now. I tried to minimize new code by re-using existing APIs. Tests check all values.

test/test_batch.cpp
```diff
@@ -158,6 +158,21 @@ struct batch_test
         CHECK_EQ(res.first(), lhs[0]);
     }
 
+    template <size_t... Is>
+    void test_get_impl(batch_type const& res, std::index_sequence<Is...>) const
+    {
+        array_type extracted = { xsimd::get<Is>(res)... };
```
serge-sans-paille, line 164: Exactly what I had in mind, thanks!

AntoinePrv:
@serge-sans-paille @DiamonDinoia this PR was merged without being properly up to date with `master`, and ending up failing in `master`.
I know the CI can be a bit slow but let's try our best to keep PR up to date and the CI 🟢

PR #1298, opened 2026-04-13
Try SVE CI on ARM

AntoinePrv:
Even with QEmu for running SVE it may be generally faster on a ARM runner


AntoinePrv:
Wow, I managed to crash both GCC *and* clang! Not sure what is going on, especially since I got it all working locally (Docker Linux + Qemu on MacOS ARM).

Edit: did not check the same versions. For clang it is starting from clang 21.

PR #1301, opened 2026-04-14
Nicer factoring of getauxval based cpu features

AntoinePrv:
- **Add a getauxval backend**
- **Split implementation**

### DIFF
--- a/include/xsimd/config/xsimd_cpu_features_arm.hpp
+++ b/include/xsimd/config/xsimd_cpu_features_arm.hpp
@@ -37,45 +37,51 @@ namespace xsimd
     class arm_cpu_features : private linux_hwcap_backend_default
     {
     public:
-        arm_cpu_features() noexcept = default;
+        inline bool neon() const noexcept;
+        inline bool neon64() const noexcept;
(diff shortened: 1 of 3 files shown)


include/xsimd/config/xsimd_cpu_features_arm.hpp
```diff
@@ -36,87 +34,54 @@ namespace xsimd
...
+        return static_cast<bool>(XSIMD_WITH_NEON);
 #endif
-        }
+    }
 
-        constexpr bool neon64() const noexcept
-        {
-            return static_cast<bool>(XSIMD_WITH_NEON64);
-        }
+    inline bool arm_cpu_features::neon64() const noexcept
```
JohanMabille, line 59: Why not keeping constexpr?

AntoinePrv: There is just this one that would be `constexpr`. I made them homogeneous to clearly indicate that *in general* thsi API is for runtime information.

JohanMabille (approved):

PR #1302, opened 2026-04-14
Check SVE and RVV vector length

AntoinePrv:
- **Add SVE vector length**
- **Add RVV vector length**
- **Fix SVE and RVV dynamic dispatch sizes**


AntoinePrv:
@serge-sans-paille @JohanMabille
Previously the size was not checked so we would dispatch to `sve<512>` even if the architecture did not support it.

JohanMabille (approved):

PR #1304, opened 2026-04-15
Dynamic dispatch VSX

AntoinePrv:
I am not very knowledgeable PowerPC but I wonder this might have been suffering from the same bug (dynamic dispatch is in fact static) as some other.
Use `getauxval` if available or fallback to previous `XSIMD_WITH_VSX` (since I understand it is mandatory on newer PPC).
Deferring to @serge-sans-paille

### DIFF
--- a/.github/workflows/cross-ppc.yml
+++ b/.github/workflows/cross-ppc.yml
@@ -35,7 +35,7 @@ jobs:
       uses: actions/checkout@v6
     - name: Setup
       run: |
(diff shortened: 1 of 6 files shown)


JohanMabille (approved):

JohanMabille:
The failure on emulated SVE is weird

AntoinePrv:
That is the usual one, we've been seeing it all across.

PR #1306, opened 2026-04-16
perf: add native AVX2 uint64/int64 mul kernel

DiamonDinoia:
Previously batch<[u]int64_t, avx2> mul fell through to AVX, which has no integer mul, which in turn fell through to SSE4.1 — splitting each 256-bit register into two 128-bit halves (vextracti128/vinserti128) and running the mul_epu32 sequence twice.

Add a sizeof(T)==8 specialization using _mm256_mul_epu32 directly, mirroring the SSE4.1 pattern with 256-bit intrinsics. Generates 8 ymm ops: 2 vpshufd, 3 vpmuludq, 2 vpaddq, 1 vpsllq — no lane splitting.

AVX512F (without DQ) also benefits since it forwards to the AVX2 kernel.

### DIFF
--- a/include/xsimd/arch/xsimd_avx2.hpp
+++ b/include/xsimd/arch/xsimd_avx2.hpp
@@ -912,6 +912,16 @@ namespace xsimd
             {
                 return _mm256_mullo_epi32(self, other);
             }
+            else XSIMD_IF_CONSTEXPR(sizeof(T) == 8)
+            {
+                return _mm256_add_epi64(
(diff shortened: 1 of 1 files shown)


DiamonDinoia:
@serge-sans-paille this is a small one :)

serge-sans-paille:
yep, looks good!

serge-sans-paille (approved):

serge-sans-paille:
You should ahve the right to merge now, feel free to do so once green.

PR #1307, opened 2026-04-17
Add IBM Z (s390x) support

Andreas-Krebbel:
This PR adds support for IBM Z. It leverages the compiler intrinsics available with GCC and Clang:
https://www.ibm.com/docs/en/open-xl-c-cpp-zos/2.2.0?topic=reference-vector-programming-support

The intrinsics are similar to what is available for IBM Power with altivec (VSX). I've started with an initial version of my patch before VSX has been added to XSIMD. However, while refreshing my changes to the latest upstream level I've used the VSX implementation as a base.

I've tested the patch on an IBM z17 (latest machine generation) with GCC and Clang:
````
[100%] Built target test_xsimd
[doctest] doctest version is "2.4.9"
[doctest] run with "--help" for options
===============================================================================
[doctest] test cases:     363 |     363 passed | 0 failed | 0 skipped
[doctest] assertions: 1681618 | 1681618 passed | 0 failed |
[doctest] Status: SUCCESS!
[100%] Built target xtest

### CODE REVIEW

**Comment by serge-sans-paille** [comment]
We need to setup CI for this. @Andreas-Krebbel could you do that, probably based on how it's done for vsx?

**Comment by Andreas-Krebbel** [comment]
> We need to setup CI for this. @Andreas-Krebbel could you do that, probably based on how it's done for vsx?

I can give it a try. Yes.

**Comment by Andreas-Krebbel** [comment]
I have tested the workflow locally and it seems to work fine. I see less testcases being executed though:
````
| [doctest] doctest version is "2.4.9"
| [doctest] run with "--help" for options
| ===============================================================================
| [doctest] test cases:  83 |  83 passed | 0 failed | 0 skipped
| [doctest] assertions: 781 | 781 passed | 0 failed |
| [doctest] Status: SUCCESS!
````

File: include/xsimd/arch/xsimd_vxe.hpp [hunk 1]
```diff
@@ -0,0 +1,797 @@
...
+        struct builtin_scalar<long>
+        {
+            using type = long long;
+        };
+
+#ifdef __CHAR_UNSIGNED__
+        template <>
+        struct builtin_scalar<char>
+        {
+            using type = unsigned char;
```
serge-sans-paille, line 66: could be `using type = typename std::conditional<std::is_signed<char>::value, signed char, unsigned char>::type` and that would avoid the `#ifdef`, right?

include/xsimd/arch/xsimd_vxe.hpp
```diff
@@ -0,0 +1,797 @@
...
+#endif
+
+        template <typename T>
+        using builtin_t = typename builtin_scalar<T>::type;
+
+        // bitwise_cast
+        template <class A, class T_in, class T_out>
+        XSIMD_INLINE batch<T_out, A> bitwise_cast(batch<T_in, A> const& self, batch<T_out, A> const&, requires_arch<vxe>) noexcept
+        {
+            return (typename batch<T_out, A>::register_type)(self.data);
```
serge-sans-paille, line 83: nit: could be a `static_cast`, but we're not very diligent on this, so feel free to ignore.

include/xsimd/arch/xsimd_vxe.hpp
```diff
@@ -0,0 +1,797 @@
...
+                uv16qi perm_imag = (uv16qi) { 8, 9, 10, 11, 12, 13, 14, 15, 24, 25, 26, 27, 28, 29, 30, 31 };
+                v2df real = vec_perm((v2df)hi.data, (v2df)lo.data, perm_real);
+                v2df imag = vec_perm((v2df)hi.data, (v2df)lo.data, perm_imag);
+                return { batch<double, A>(real), batch<double, A>(imag) };
+            }
+
+            template <class A>
+            XSIMD_INLINE batch<float, A> complex_low(batch<std::complex<float>, A> const& self, requires_arch<vxe>) noexcept
+            {
+                uv16qi perm = (uv16qi) { 0, 1, 2, 3, 16, 17, 18, 19, 4, 5, 6, 7, 20, 21, 22, 23 };
```
serge-sans-paille, line 142: gcc constructor expression. that's a gnu extension, but I guess it's fine. I've been using it for PowerPC too.

include/xsimd/arch/xsimd_vxe.hpp
```diff
@@ -0,0 +1,797 @@
...
+
+        // floor
+        template <class A, class T, class = std::enable_if_t<std::is_floating_point<T>::value>>
+        XSIMD_INLINE batch<T, A> floor(batch<T, A> const& self, requires_arch<vxe>) noexcept
+        {
+            return vec_floor(self.data);
+        }
+        // round
+        // vec_round rounds ties to even instead of zero
+#if defined __has_builtin && __has_builtin(__builtin_s390_vfi)
```
serge-sans-paille, line 782: This seems to be supported on modern clang, good.

include/xsimd/config/xsimd_cpuid.hpp
```diff
@@ -72,6 +71,7 @@ namespace xsimd
             ARCH_FIELD_EX(detail::rvv<128>, rvv128)
             ARCH_FIELD(wasm)
             ARCH_FIELD(vsx)
+            ARCH_FIELD(vxe)
```
serge-sans-paille, line 74: you should also add a piece of code for runtime detection. See how it's done for power there: https://github.com/xtensor-stack/xsimd/blob/master/include/xsimd/config/xsimd_cpu_features_ppc.hpp

include/xsimd/types/xsimd_vxe_register.hpp
```diff
@@ -0,0 +1,86 @@
...
+        };                                                           \
+    };                                                               \
+    XSIMD_DECLARE_SIMD_REGISTER(T, vxe, __vector Tv)
+
+        // The VXE vector intrinsics do not support long, unsigned long,
+        // and char data types.  batches of these types are vectors of
+        // equivalent types.
+        XSIMD_DECLARE_SIMD_BOOL_VXE_REGISTER(signed char, signed char, char);
+        XSIMD_DECLARE_SIMD_BOOL_VXE_REGISTER(unsigned char, unsigned char, char);
+#ifdef __CHAR_UNSIGNED__
```
serge-sans-paille, line 67: If you could avoid using a macro using a type alais instead, that would be cool.

serge-sans-paille (commented):
Some nit picking, just missing the runtime detection and we're good.

It also means we should improve CI to detect that...

serge-sans-paille:
The error in arch consistency is related to qemu emulation capability, more on https://www.qemu.org/docs/master/system/target-s390x.html

include/xsimd/arch/xsimd_vxe.hpp
Andreas-Krebbel: A static_cast didn't work with GCC since it doesn't allow a cast between vector types with different signness. clang was ok with it though.

include/xsimd/arch/xsimd_vxe.hpp
Andreas-Krebbel: Yes, this seems to work. Thanks!

include/xsimd/config/xsimd_cpuid.hpp
Andreas-Krebbel: Done

include/xsimd/types/xsimd_vxe_register.hpp
Andreas-Krebbel: It would work with GCC, but it didn't with clang. Clang does not seem to accept type aliases in combination with __vector bool.

serge-sans-paille:
This all looks good to me, thanks *a lot* for going through this and adressing the reviews. I'll leave it to you to squash and rebase and we're good!

serge-sans-paille:
@Andreas-Krebbel the risc-v issue is not on you, I'll accept your PR anyway and leave it to @AntoinePrv to fix it ;-)

AntoinePrv:
@serge-sans-paille, sure but please wait for it to be fixed before merging this one (#1322).

AntoinePrv:
@Andreas-Krebbel can you rebase your PR?

PR #1308, opened 2026-04-17
Remove unused standard includes

serge-sans-paille:


serge-sans-paille:
worked like a charm, yay

AntoinePrv:
@serge-sans-paille I think there are number of header tools in clang-tidy that we can use.
E.g. https://clang.llvm.org/extra/clang-tidy/checks/misc/include-cleaner.html

PR #1310, opened 2026-04-20
basic clang tidy setup

serge-sans-paille:


serge-sans-paille:
@JohanMabille I'm super happy with those checks, the only problem is that this PR only enables it for x86_64 and common files.

JohanMabille (approved):
That's better than nothing and that gives us hints about what to search for and fix for other archs (assuming we've been systematic in our coding style ;))

PR #1314, opened 2026-04-21
Factor integer traits and disentangle traits headers

AntoinePrv:
Factor integer trait at a higher level to avoid repeated work of explicitly defining something for all sizes.
Basically avoid repeating the following kind of logic all over:
```cpp
struct sized_int<1> {
    using type = int8_t;
};
struct sized_int<2> {
    using type = int16_t;
};
...
```
This is added in `xsimd/type_traits.hpp`.
This PR as a lot of noise as I needed to fix a few circular header inclusion.

This PR is still imperfect. IMHO we still have some work work to properly factor utilities and ensure all header inclusion are non circular, order independent, and do not depend on leaking behavior (or transitively dependent).

### DIFF
--- a/include/xsimd/types/xsimd_utils.hpp
+++ b/include/xsimd/types/xsimd_utils.hpp
@@ -23,15 +23,10 @@
 #include "xtl/xcomplex.hpp"
 #endif
 
+#include "./xsimd_batch_fwd.hpp"
+
 namespace xsimd
 {
-
-    template <class T, class A>
-    class batch;
-
-    template <class T, class A>
-    class batch_bool;
-
     /**************
      * index      *
      **************/
--- a/include/xsimd/utils/xsimd_type_traits.hpp
+++ b/include/xsimd/utils/xsimd_type_traits.hpp
@@ -1,0 +1,127 @@
+/***************************************************************************
+ * Copyright (c) Johan Mabille, Sylvain Corlay, Wolf Vollprecht and         *
+ * Martin Renou                                                             *
+ * Copyright (c) QuantStack                                                 *
+ * Copyright (c) Serge Guelton                                              *
+ *                                                                          *
+ * Distributed under the terms of the BSD 3-Clause License.                 *
+ *                                                                          *
+ * The full license is in the file LICENSE, distributed with this software. *
+ ****************************************************************************/
+
+#ifndef XSIMD_TYPE_TRAITS_HPP
+#define XSIMD_TYPE_TRAITS_HPP
+
+#include <cstddef>
+#include <cstdint>
+#include <type_traits>
+
+namespace xsimd
+{
+    namespace detail
+    {
+        template <std::size_t S>
+        struct sized_num_types;
+
+        template <>
+        struct sized_num_types<1>
+        {
+            using signed_type = std::int8_t;
+            using unsigned_type = std::uint8_t;
+            using floating_point_type = void;
+        };
+
+        template <>
+        struct sized_num_types<2>
+        {
+            using signed_type = std::int16_t;
+            using unsigned_type = std::uint16_t;
+            using floating_point_type = void;
+        };
+
+        template <>
+        struct sized_num_types<4>
+        {
+            using signed_type = std::int32_t;
+            using unsigned_type = std::uint32_t;
+            using floating_point_type = float;
+        };
+
+        template <>
+        struct sized_num_types<8>
+        {
+            using signed_type = std::int64_t;
+            using unsigned_type = std::uint64_t;
+            using floating_point_type = double;
+        };
+    }
+
+    /**
+     * @ingroup batch_traits
+     *
+     * Signed integer type with exactly @c S bytes (1, 2, 4, or 8).
+     *
+     * @tparam S size in bytes.
+     */
+    template <std::size_t S>
+    using sized_int_t = typename detail::sized_num_types<S>::signed_type;
+
+    /**
+     * @ingroup batch_traits
+     *
+     * Unsigned integer type with exactly @c S bytes (1, 2, 4, or 8).
+     *
+     * @tparam S size in bytes.
+     */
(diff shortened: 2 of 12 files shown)


include/xsimd/utils/xsimd_type_traits.hpp
```diff
@@ -0,0 +1,127 @@
...
+            using floating_point_type = float;
+        };
+
+        template <>
+        struct sized_num_types<8>
+        {
+            using signed_type = std::int64_t;
+            using unsigned_type = std::uint64_t;
+            using floating_point_type = double;
+        };
```
serge-sans-paille, line 56: I like this!

include/xsimd/utils/xsimd_type_traits.hpp
```diff
@@ -0,0 +1,127 @@
...
+    template <std::size_t S>
+    using sized_fp_t = typename detail::sized_num_types<S>::floating_point_type;
+
+    namespace detail
+    {
+        template <typename T, typename = void>
+        struct widen;
+
+        template <typename T>
+        struct widen<T, std::enable_if_t<std::is_floating_point<T>::value>>
```
serge-sans-paille, line 96: nit: you could inherit from `detail::sized_num_types<S>::floating_point_type|

include/xsimd/utils/xsimd_type_traits.hpp
```diff
@@ -0,0 +1,127 @@
...
+        };
+
+        template <typename T>
+        struct widen<T, std::enable_if_t<!std::is_floating_point<T>::value && std::is_signed<T>::value>>
+        {
+            using type = xsimd::sized_int_t<sizeof(T) * 2>;
+        };
+
+        template <typename T>
+        struct widen<T, std::enable_if_t<!std::is_floating_point<T>::value && std::is_unsigned<T>::value>>
```
serge-sans-paille, line 108: !std::is_floating_point seems redundant here

include/xsimd/types/xsimd_utils.hpp
```diff
@@ -23,15 +23,10 @@
 #include "xtl/xcomplex.hpp"
 #endif
 
+#include "./xsimd_batch_fwd.hpp"
```
serge-sans-paille, line 26: good point.

serge-sans-paille (commented):
While you're at it, maybe you could make all internal includes absolute instead of relative?

AntoinePrv:
@serge-sans-paille I definitely want to do a pass on header but I think this is PR is ugly enough (and I am already spending too much time on xsimd 😅 )

JohanMabille:
> While you're at it, maybe you could make all internal includes absolute instead of relative?

I think we should keep them relativem this makes it easier for projects vendoring xsimd.

include/xsimd/utils/xsimd_type_traits.hpp
JohanMabille: Not sure to understand the benefit of it since we would still need to define `type` as `base::floating_point_type` in the struct.

include/xsimd/utils/xsimd_type_traits.hpp
```diff
@@ -0,0 +1,127 @@
...
+ * Copyright (c) QuantStack                                                 *
+ * Copyright (c) Serge Guelton                                              *
+ *                                                                          *
+ * Distributed under the terms of the BSD 3-Clause License.                 *
+ *                                                                          *
+ * The full license is in the file LICENSE, distributed with this software. *
+ ****************************************************************************/
+
+#ifndef XSIMD_TYPE_TRAITS_HPP
+#define XSIMD_TYPE_TRAITS_HPP
```
JohanMabille, line 13: This file should be named xsimd_type_traits.hpp for consistency, and should probably live in xsimd/types. Only `xsimd.hpp` should be at the "root" of the project

JohanMabille (changes requested):

include/xsimd/utils/xsimd_type_traits.hpp
AntoinePrv: > This file should be named xsimd_type_traits.hpp for consistency

Ok, though I think we should get rid of that convention in the next major release.

> and should probably live in xsimd/types. Only xsimd.hpp should be at the "root" of the project

@JohanMabille 
This is weird to me. Why would everything need to live in `type`? IMHO this participate in the general confusion about header organization and weird "../../type" inclusion.
How about "xsimd/utils"?

AntoinePrv:
> I think we should keep them relativem this makes it easier for projects vendoring xsimd.

I think we should settle on either `#include "xsimd/header.h"` or `#include "./header.h"` but `#include "header.h"` should be forbidden.

include/xsimd/utils/xsimd_type_traits.hpp
JohanMabille: > though I think we should get rid of that convention in the next major release.

Not sure to understand why, having a single "include-all" header at the root of the project is a convenient pattern; having many headers at the same level would be weird.

> How about "xsimd/utils"?

Works for me, as long as it is not at the sane level as `xsimd.hpp`

AntoinePrv: > > though I think we should get rid of that convention in the next major release.
> 
> Not sure to understand why, having a single "include-all" header at the root of the project is a convenient pattern; having many headers at the same level would be weird.
> 
> > How about "xsimd/utils"?
> 
> Works for me, as long as it is not at the sane level as `xsimd.hpp`

I meant the `xsimd_` prefix convention for filenames which I think is needlessly verbose and very much C-style.

JohanMabille: We can see to remove it in next major release, but for now let's stay consistent.

JohanMabille (approved):

PR #1317, opened 2026-04-23
Fix windows on Neon

AntoinePrv:
Basically Windows ARM64 was never built. This is a first.

This removed all the wrapper macros and dispatcher hierarchy. The manual wrapper are a bit longer but at least are transparent for the reader, easier to search, and portable (also https://github.com/apache/arrow/pull/49756 is reporting the previous implentation may not nicely inline).

The wrapper also now start with `x_` because MSVC does not allow `::intrinsic` so we needed to disambiguate names.

Beyond that, there was other small fixes where MSVC was not happy. I had to move some code around as well to reused other functions.

Close #1288 #1285 #1250 #612 #611


JohanMabille (approved):

serge-sans-paille:
@JohanMabille : I'm quite unhappy with that merge: the patch stack is not clean at all, which leads to a commit message that's full of noise.
It's quite a lot of changes and I would rather had time to review it properly ;-)
I'll do a post-merge review this evening.

JohanMabille:
Hey, sorry for that, I was a little too enthusiastic about having NEON working on Windows, and the possibility to release and move to C++17.  We can still amend the commit message and make it cleaner; happy to commit any change according to your review before releasing.

serge-sans-paille:
> Hey, sorry for that, I was a little too enthusiastic about having NEON working on Windows, and the possibility to release and move to C++17. We can still amend the commit message and make it cleaner; happy to commit any change according to your review before releasing.

I can understand this, I'm very happy with that update too. Between this, the support for s390x and the recent improvements, we're heading toward something very nice. And once C++17 makes the cut, it will be

\o\ |o| /o/

.github/workflows/windows.yml
```diff
@@ -145,7 +145,7 @@ jobs:
     - name: Setup compiler
       uses: ilammy/msvc-dev-cmd@v1
       with:
-        arch: amd64
+        arch: arm64
```
serge-sans-paille, line 148: I'd rather have this change in a seperate commit (in the same patch stack, that's fine)

include/xsimd/arch/xsimd_neon.hpp
```diff
@@ -306,6 +47,200 @@ namespace xsimd
...
+            XSIMD_INLINE uint8x16_t x_vreinterpretq(uint32x4_t a) noexcept { return vreinterpretq_u8_u32(a); }
+            template <class R, class T, std::enable_if_t<std::is_same<R, uint8_t>::value && std::is_same<T, int32_t>::value, int> = 0>
+            XSIMD_INLINE uint8x16_t x_vreinterpretq(int32x4_t a) noexcept { return vreinterpretq_u8_s32(a); }
+            template <class R, class T, std::enable_if_t<std::is_same<R, uint8_t>::value && std::is_same<T, uint64_t>::value, int> = 0>
+            XSIMD_INLINE uint8x16_t x_vreinterpretq(uint64x2_t a) noexcept { return vreinterpretq_u8_u64(a); }
+            template <class R, class T, std::enable_if_t<std::is_same<R, uint8_t>::value && std::is_same<T, int64_t>::value, int> = 0>
+            XSIMD_INLINE uint8x16_t x_vreinterpretq(int64x2_t a) noexcept { return vreinterpretq_u8_s64(a); }
+            template <class R, class T, std::enable_if_t<std::is_same<R, uint8_t>::value && std::is_same<T, float>::value, int> = 0>
+            XSIMD_INLINE uint8x16_t x_vreinterpretq(float32x4_t a) noexcept { return vreinterpretq_u8_f32(a); }
```
serge-sans-paille, line 74: this would have been simpler to write (and faster to compile) through class specialization.

include/xsimd/arch/xsimd_neon.hpp
```diff
@@ -3153,37 +3095,32 @@ namespace xsimd
...
         {
             using register_type = typename batch<T, A>::register_type;
-            // Adding modulo to avoid warning.
-            const detail::neon_dispatcher::binary dispatcher = {
-                std::make_tuple(wrap::rotate_left_u8<N>, wrap::rotate_left_s8<N>, wrap::rotate_left_u16<N % 8>, wrap::rotate_left_s16<N % 8>,
-                                wrap::rotate_left_u32<N % 4>, wrap::rotate_left_s32<N % 4>, wrap::rotate_left_u64<N % 2>, wrap::rotate_left_s64<N % 2>,
-                                wrap::rotate_left_f32<N % 4>)
-            };
-            return dispatcher.apply(register_type(a), register_type(a));
+            return wrap::x_rotate_left<N, project_num_t<T>>(register_type(a), register_type(a));
```
serge-sans-paille, line 3123: not related to that PR, but I'm not a big fan of than function name `project_num_t`. I do understand why it's there though. Maybe `map_to_sized_type_t` ?

include/xsimd/arch/xsimd_neon64.hpp
```diff
@@ -879,186 +888,73 @@ namespace xsimd
...
          **************/
 
-        WRAP_REDUCER_INT_EXCLUDING_64(vmaxvq)
-        WRAP_REDUCER_FLOAT(vmaxvq)
-
         namespace wrap
         {
-            XSIMD_INLINE uint64_t vmaxvq_u64(uint64x2_t a) noexcept
+            // TODO(c++17): Make a single function with if constexpr switch
+            template <class T, std::enable_if_t<std::is_same<T, uint8_t>::value, int> = 0>
```
serge-sans-paille, line 934: why do you need this `T` parameter + the enable_if ? it looks like you could just write

```
XSIMD_INLINE uint8_t x_vmaxvq(uint8x16_t a) noexcept { return vmaxvq_u8(a); }
```

oh, of course that's becuase of windows that does not have strong types for each `uint8x16_t` etc type.`

I think there should be a big comment on top of the file to explain this.

test/test_utils.hpp
```diff
@@ -399,10 +399,29 @@ namespace detail
         void stringify(std::ostream* os) const override { *os << msg_; }
     };
 
+    template <typename T, typename std::enable_if<std::is_arithmetic<T>::value, int>::type = 0>
+    std::string to_string_full_precision(T value)
+    {
+        // TODO(C++17): use std::to_chars
+        char buf[64];
```
serge-sans-paille, line 406: shouldn't this `64` be `std::numeric_limits<T>::max_digits10 + 1/*sign, unsure of that one*/ +1 /*null*/ ` ?

test/test_utils.hpp
```diff
@@ -399,10 +399,29 @@ namespace detail
...
+        std::snprintf(
+            buf, sizeof(buf),
```
serge-sans-paille, line 408: nit:    `std::size(buf)`

serge-sans-paille (commented):
I've done my review as I would have done it if we were earlier in the process. Feel free to apply some or all of the comments.

include/xsimd/arch/xsimd_neon.hpp
AntoinePrv: I had in mind a single function with `if constexpr`, and backported it to C++14.
This way we do not have to change the call sites.

include/xsimd/arch/xsimd_neon.hpp
AntoinePrv: I agree it not the best name, especially since the pattern shows up a lot.
It's a mathematical projection, in that `project_num_t<project_num_t<T>>` is `project_num_t<T>`.
`map_to_sized_type_t` is good though.

include/xsimd/arch/xsimd_neon64.hpp
AntoinePrv: > oh, of course that's because of windows that does not have strong types for each `uint8x16_t` etc type.`

Exactly! I general I found that directly on scalar types (`uint8_t`, ...) and getting the vector type via a single type trait is a pattern that works quite well and avoid having to adapt to each compilers.

Sending a patch for the comments.

test/test_utils.hpp
AntoinePrv: Sending patch

test/test_utils.hpp
AntoinePrv: Sending patch

PR #1318, opened 2026-04-24
Make test_rounding to go over all the inputs

Andreas-Krebbel:
test_rounding calculates how many batches need to be run to cover all the input values in nb_batches. However, when the loops go over the number of batches, the batch count is increased by the batch size. This is probably not what was intended here.

What currently happens is that e.g. for a 128 bit float vector only the first four values are actually tested, leaving out some of the interesting ones coming later.

I've tested the change on x86 and s390x. Both are still clean.

On Power however, this reveals an oddity with the vec_round builtin on float vectors, where it uses a different rounding mode than for doubles. I'll try to address this with a separate PR.

### DIFF
--- a/test/test_rounding.cpp
+++ b/test/test_rounding.cpp
@@ -56,12 +56,12 @@ struct rounding_test
             std::transform(input.cbegin(), input.cend(), expected.begin(),
                            [](const value_type& v)
                            { return std::ceil(v); });
-            for (size_t i = 0; i < nb_batches; i += size)
+            for (size_t i = 0; i < nb_batches; i++)
             {
                 batch_type in, out, ref;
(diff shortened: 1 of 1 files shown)


test/test_rounding.cpp
```diff
@@ -56,12 +56,12 @@ struct rounding_test
             std::transform(input.cbegin(), input.cend(), expected.begin(),
                            [](const value_type& v)
                            { return std::ceil(v); });
-            for (size_t i = 0; i < nb_batches; i += size)
+            for (size_t i = 0; i < nb_batches; i++)
```
serge-sans-paille: /o\

serge-sans-paille:
thanks!

Issue #1323, opened 2026-04-24
Failed to build with gcc-8.4.0 -std=c++14


hdu-sdlzx:
# Problem

The <version> header is introduced in C++20.

# Build Log

```
$ g++ test.cpp -I./include -std=c++14
In file included from ./include/xsimd/types/../arch/./xsimd_common.hpp:16,
                 from ./include/xsimd/types/../arch/xsimd_isa.hpp:140,
                 from ./include/xsimd/types/xsimd_batch.hpp:520,
                 from ./include/xsimd/xsimd.hpp:77,
                 from test.cpp:1:
./include/xsimd/types/../arch/././common/xsimd_common_bit.hpp:8:10: fatal error: version: No such file or directory
 #include <version>
          ^~~~~~~~~
compilation terminated.
```

serge-sans-paille:
You fixed that one in #1324 🙇

PR #1324, opened 2026-04-24
fix build for old compilers (#1323)

hdu-sdlzx:
version header and std::countl_one are introduced in C++20

### DIFF
--- a/include/xsimd/arch/common/xsimd_common_bit.hpp
+++ b/include/xsimd/arch/common/xsimd_common_bit.hpp
@@ -5,6 +5,8 @@
 #ifndef XSIMD_BIT_HPP
 #define XSIMD_BIT_HPP
 
+#if __cplusplus > 202002L
+
 #include <version>
 
 #if __cpp_lib_bitops >= 201907L
@@ -22,6 +24,8 @@ namespace xsimd
         using std::popcount;
     }
 }
+
+#endif
 
 #else
 


include/xsimd/arch/common/xsimd_common_bit.hpp
```diff
@@ -5,6 +5,8 @@
 #ifndef XSIMD_BIT_HPP
 #define XSIMD_BIT_HPP
 
+#if __cplusplus > 202002L
+
 #include <version>
 
 #if __cpp_lib_bitops >= 201907L
```
serge-sans-paille, line 12: Does this check and the include of '<version>' still makes sense then?

serge-sans-paille:
Thanks. It's quite strange that our CI didn't catch it though.

include/xsimd/arch/common/xsimd_common_bit.hpp
JohanMabille: I guess it is for compilers that were late to implement some features of the standard?

hdu-sdlzx:
The CI failure looks unrelated to this PR, try to fix it with #1327

include/xsimd/arch/common/xsimd_common_bit.hpp
hdu-sdlzx: This makes sense when someone uses a compiler that does not fully support C++20.

Looks like clang(libcxx) implemented version header ([P0754R2](https://wg21.link/P0754R2)) in clang 7 and bit operations ([P0553R4](https://wg21.link/P0553R4)) in clang 9.

Reference: https://libcxx.llvm.org/Status/Cxx20.html

serge-sans-paille: ok, let's live with it for a while then.

PR #1325, opened 2026-04-24
IBM Z (s390x) fixes

Andreas-Krebbel:
- Fix an issue which so far prevented the full set of tests being compiled in when running in the github workflow.
- While doing this, align cross-s390x.yml with the latest cross-ppc.yml changes.
- Add a cpu_feature testcase for s390x VXE and add a few pieces which were still missing.

Tested with GCC and Clang on IBM z17. Workflow tested locally. It runs all the tests now successfully.

### DIFF
--- a/.github/workflows/cross-s390x.yml
+++ b/.github/workflows/cross-s390x.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         target:
-          - { platform: 's390x',     dir: 's390x-linux-gnu',   flags: '-mzvector -march=z14', full: 'OFF' }
+          - { platform: 's390x',     dir: 's390x-linux-gnu',   full: 'OFF' }
         sys:
           - { compiler: 'gcc',   version: '14' }
     steps:
@@ -34,7 +34,7 @@ jobs:
       uses: actions/checkout@v6
     - name: Setup
       run: |
-        cmake -B _build \
+        cmake -B build/ \
               -DBUILD_TESTS=ON -DDOWNLOAD_DOCTEST=ON \
               -DBUILD_BENCHMARK=${{ matrix.target.full }} -DBUILD_EXAMPLES=${{ matrix.target.full }} \
               -DCMAKE_BUILD_TYPE=Release \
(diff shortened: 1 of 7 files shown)


.github/workflows/cross-s390x.yml
```diff
@@ -34,15 +34,22 @@ jobs:
...
+        cmake -B build/ \
               -DBUILD_TESTS=ON -DDOWNLOAD_DOCTEST=ON \
               -DBUILD_BENCHMARK=${{ matrix.target.full }} -DBUILD_EXAMPLES=${{ matrix.target.full }} \
               -DCMAKE_BUILD_TYPE=Release \
               -DCMAKE_C_FLAGS="${{ matrix.target.flags }}" \
               -DCMAKE_CXX_FLAGS="${{ matrix.target.flags }}" \
               -DCMAKE_TOOLCHAIN_FILE=${{ github.workspace }}/.github/toolchains/${{ matrix.sys.compiler }}-${{ matrix.target.dir }}.cmake
     - name: Build
-      run: cmake --build _build --verbose -j1
+      run: cmake --build build/ --verbose -j1
```
serge-sans-paille, line 45: I'm curious: why this change?

.github/workflows/cross-s390x.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         target:
-          - { platform: 's390x',     dir: 's390x-linux-gnu',   flags: '-mzvector -march=z14', full: 'OFF' }
+          - { platform: 's390x',     dir: 's390x-linux-gnu',   flags: '', full: 'OFF' }
```
serge-sans-paille: remove `flags` then? Or keep it so that we can update in one place in the future? Up to you.

serge-sans-paille (approved):

.github/workflows/cross-s390x.yml
Andreas-Krebbel: This is how cross-ppc.yml does it and I thought it would make sense to keep the two files as close as possible.

.github/workflows/cross-s390x.yml
Andreas-Krebbel: The flags have been redundant since I already had it in test/CMakeLists.txt so they appeared twice on the command line in the github workflow:
````
    elseif(${CMAKE_SYSTEM_PROCESSOR} MATCHES "s390x")
        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=z14 -mzvector")
````
Power seems to do it the other way around.

I'll remove flags from cross-s390x.yml then I guess.

PR #1328, opened 2026-04-26
Provide cmake-level configuration switch for emulated build

serge-sans-paille:


serge-sans-paille:
@AntoinePrv : thanks to @hdu-sdlzx 's hint, this should fix the emulated test.

PR #1329, opened 2026-04-27
Tentative support for avx extensions to 128 bit registers

serge-sans-paille:


AntoinePrv (commented):
We need to implement a CPU feature detection that checks for the presence of AVX in CPUID but is unaffected by `xcr0::avx` (contrary to current `avx` method).

serge-sans-paille:
> We need to implement a CPU feature detection that checks for the presence of AVX in CPUID but is unaffected by `xcr0::avx` (contrary to current `avx` method).

is this independent from this patch?

AntoinePrv:
Could very much be, yes

serge-sans-paille:
+ avx2_128

serge-sans-paille:
@DiamonDinoia / @AntoinePrv : ready for review!

include/xsimd/arch/xsimd_avx_128.hpp
```diff
@@ -0,0 +1,163 @@
...
+        template <class A, class T, class ITy, class = std::enable_if_t<std::is_floating_point<T>::value && sizeof(T) == sizeof(ITy)>>
+        XSIMD_INLINE batch<T, A> swizzle(batch<T, A> const& self, batch<ITy, A> mask, requires_arch<avx_128>) noexcept
+        {
+            XSIMD_IF_CONSTEXPR(std::is_same<T, float>::value)
+            {
+                return _mm_permutevar_ps(self, mask);
+            }
+            else
+            {
+                // FIXME: _mm_permutevar_pd fails validation, but it shouldn't o_O
```
DiamonDinoia: Maybe:  Each pair of source elements must lie in the same 128-bit region as the destination.

Blatant advertise for my own tool: https://diamondinoia.com/simdref/#_mm_permutevar_pd

AntoinePrv (commented):
I won't have time to inspect the details but the general structure is looking good!

DiamonDinoia:
Things looks good to me, I verified that we use the correct intrinsics. One suggestion
```diff
-                // FIXME: _mm_permutevar_pd fails validation, but it shouldn't o_O                                                                             
-                return swizzle(self, mask, sse4_2 {});
-                // return _mm_permutevar_pd(self, mask);                                                                                                       
+                // VPERMILPD's variable control reads bit 1 of each 64-bit selector
+                // (bit 0 is ignored), so a {0,1} index needs to become {0,2}.                                                                                 
+                // Negation is a cheap alternative to a left shift by 1.                                                                                       
+                return _mm_permutevar_pd(self, -(mask & batch<ITy, A>(1)));

```

serge-sans-paille:
> Things looks good to me, I verified that we use the correct intrinsics. One suggestion
> 
> ```diff
> -                // FIXME: _mm_permutevar_pd fails validation, but it shouldn't o_O                                                                             
> -                return swizzle(self, mask, sse4_2 {});
> -                // return _mm_permutevar_pd(self, mask);                                                                                                       
> +                // VPERMILPD's variable control reads bit 1 of each 64-bit selector
> +                // (bit 0 is ignored), so a {0,1} index needs to become {0,2}.                                                                                 
> +                // Negation is a cheap alternative to a left shift by 1.                                                                                       
> +                return _mm_permutevar_pd(self, -(mask & batch<ITy, A>(1)));
> ```

I've applied your patch, but without the `& 1` which seems useless (mask can only be 0 or 1 in that context, right?)

DiamonDinoia:
> > Things looks good to me, I verified that we use the correct intrinsics. One suggestion
> > ```diff
> > -                // FIXME: _mm_permutevar_pd fails validation, but it shouldn't o_O                                                                             
> > -                return swizzle(self, mask, sse4_2 {});
> > -                // return _mm_permutevar_pd(self, mask);                                                                                                       
> > +                // VPERMILPD's variable control reads bit 1 of each 64-bit selector
> > +                // (bit 0 is ignored), so a {0,1} index needs to become {0,2}.                                                                                 
> > +                // Negation is a cheap alternative to a left shift by 1.                                                                                       
> > +                return _mm_permutevar_pd(self, -(mask & batch<ITy, A>(1)));
> > ```
> 
> I've applied your patch, but without the `& 1` which seems useless (mask can only be 0 or 1 in that context, right?)

Yes, the difference would be only if the user passes an invalid mask (3 for example). But, I don't think we allow that.

DiamonDinoia (approved):

serge-sans-paille:
Let's do the avx512 part now :fearful:

PR #1332, opened 2026-04-28
feat: add runtime batch_bool mask overloads for load_masked/store_masked

DiamonDinoia:
Add runtime-mask overloads of xsimd::load_masked and xsimd::store_masked across AVX2, AVX-512, SSE, SVE, RVV, and NEON. The generic common-path fallback is collapsed to a whole-vector select, and the unaligned page-cross fast path is dropped since the underlying intrinsics suppress faults on masked-off lanes regardless of alignment.

### DIFF
--- a/docs/source/api/data_transfer.rst
+++ b/docs/source/api/data_transfer.rst
@@ -12,7 +12,7 @@ Data Transfers
 From memory:
 
 +---------------------------------------+----------------------------------------------------+
-| :cpp:func:`load`                      | load values from memory (optionally masked)        |
+| :cpp:func:`load`                      | load values from memory (optionally masked) [#m]_  |
 +---------------------------------------+----------------------------------------------------+
 | :cpp:func:`load_aligned`              | load values from aligned memory                    |
 +---------------------------------------+----------------------------------------------------+
@@ -32,7 +32,7 @@ From a scalar:
 To memory:
 
 +---------------------------------------+----------------------------------------------------+
-| :cpp:func:`store`                     | store values to memory (optionally masked)         |
+| :cpp:func:`store`                     | store values to memory (optionally masked) [#m]_   |
 +---------------------------------------+----------------------------------------------------+
 | :cpp:func:`store_aligned`             | store values to aligned memory                     |
 +---------------------------------------+----------------------------------------------------+
@@ -84,3 +84,11 @@ The following empty types are used for tag dispatching:
 
 .. doxygenstruct:: xsimd::unaligned_mode
    :project: xsimd
+
+.. rubric:: Footnotes
+
+.. [#m] Masked ``load`` / ``store`` come in two flavours. The
+   :cpp:class:`batch_bool_constant` overload encodes the mask in the type and
+   is resolved at compile time. The runtime :cpp:class:`batch_bool` overload
+   accepts a mask computed at runtime. Prefer the compile-time mask whenever
+   the selection is known at compile time.
--- a/docs/source/api/data_transfer.rst
+++ b/docs/source/api/data_transfer.rst
@@ -90,5 +90,5 @@ The following empty types are used for tag dispatching:
 .. [#m] Masked ``load`` / ``store`` come in two flavours. The
    :cpp:class:`batch_bool_constant` overload encodes the mask in the type and
    is resolved at compile time. The runtime :cpp:class:`batch_bool` overload
-   accepts a mask computed at runtime. Prefer the compile-time mask whenever
-   the selection is known at compile time.
+   accepts a mask computed at runtime. For performance reasons, prefer the
+   compile-time mask whenever possible.
--- a/include/xsimd/arch/common/xsimd_common_memory.hpp
+++ b/include/xsimd/arch/common/xsimd_common_memory.hpp
@@ -441,12 +441,8 @@ namespace xsimd
         XSIMD_INLINE batch<T, A>
         load_masked(T const* mem, batch_bool<T, A> mask, convert<T>, Mode, requires_arch<common>) noexcept
         {
-            // Per-lane validity contract: only active lanes of ``mem`` are
-            // required to be addressable. An unconditional whole-vector load
-            // would touch inactive lanes and trip ASan/Valgrind on partial
-            // buffers, so stay scalar. Arches with hardware predicated loads
-            // (AVX2 32/64-bit, AVX-512, SVE, RVV) override this with a single
-            // intrinsic that suppresses inactive-lane reads in hardware.
+            // Scalar fallback: only active lanes are touched. Arches with
+            // hardware predicated loads override this.
             constexpr std::size_t size = batch<T, A>::size;
             alignas(A::alignment()) std::array<T, size> buffer;
             for (std::size_t i = 0; i < size; ++i)
@@ -465,12 +461,8 @@ namespace xsimd
         XSIMD_INLINE void
         store_masked(T* mem, batch<T, A> const& src, batch_bool<T, A> mask, Mode, requires_arch<common>) noexcept
         {
-            // Per-lane validity contract (matches native masked-store APIs):
-            // only active lanes of ``mem`` are touched. A load+select+store
-            // RMW would both read and write inactive bytes, breaking that
-            // contract — stay scalar. Arches with hardware predicated stores
-            // override this with a single intrinsic that suppresses inactive
-            // lanes in hardware.
+            // Scalar fallback: only active lanes are touched. Arches with
+            // hardware predicated stores override this.
             constexpr std::size_t size = batch<T, A>::size;
             alignas(A::alignment()) std::array<T, size> src_buf;
             src.store_aligned(src_buf.data());
--- a/include/xsimd/arch/common/xsimd_common_memory.hpp
+++ b/include/xsimd/arch/common/xsimd_common_memory.hpp
@@ -437,11 +437,46 @@ namespace xsimd
             return detail::load_masked_common(mem, mask, cvt, mode, detail::masked_memory_uses_fp_bitcast<A, T_in, T_out> {});
         }
 
+        template <class A, class T, class Mode>
+        XSIMD_INLINE batch<T, A>
+        load_masked(T const* mem, batch_bool<T, A> mask, convert<T>, Mode, requires_arch<common>) noexcept
+        {
+            // Per-lane validity contract: only active lanes of ``mem`` are
+            // required to be addressable. An unconditional whole-vector load
+            // would touch inactive lanes and trip ASan/Valgrind on partial
+            // buffers, so stay scalar. Arches with hardware predicated loads
+            // (AVX2 32/64-bit, AVX-512, SVE, RVV) override this with a single
+            // intrinsic that suppresses inactive-lane reads in hardware.
+            constexpr std::size_t size = batch<T, A>::size;
+            alignas(A::alignment()) std::array<T, size> buffer;
+            for (std::size_t i = 0; i < size; ++i)
+                buffer[i] = mask.get(i) ? mem[i] : T(0);
+            return batch<T, A>::load_aligned(buffer.data());
+        }
+
         template <class A, class T_in, class T_out, bool... Values, class alignment>
         XSIMD_INLINE void
         store_masked(T_out* mem, batch<T_in, A> const& src, batch_bool_constant<T_in, A, Values...> mask, alignment mode, requires_arch<common>) noexcept
         {
             detail::store_masked_common(mem, src, mask, mode, detail::masked_memory_uses_fp_bitcast<A, T_in, T_out> {});
+        }
+
+        template <class A, class T, class Mode>
+        XSIMD_INLINE void
+        store_masked(T* mem, batch<T, A> const& src, batch_bool<T, A> mask, Mode, requires_arch<common>) noexcept
+        {
+            // Per-lane validity contract (matches native masked-store APIs):
+            // only active lanes of ``mem`` are touched. A load+select+store
+            // RMW would both read and write inactive bytes, breaking that
+            // contract — stay scalar. Arches with hardware predicated stores
+            // override this with a single intrinsic that suppresses inactive
+            // lanes in hardware.
+            constexpr std::size_t size = batch<T, A>::size;
+            alignas(A::alignment()) std::array<T, size> src_buf;
+            src.store_aligned(src_buf.data());
+            for (std::size_t i = 0; i < size; ++i)
+                if (mask.get(i))
+                    mem[i] = src_buf[i];
         }
 
         template <class A, class T_in, class T_out>
--- a/include/xsimd/arch/common/xsimd_common_memory.hpp
+++ b/include/xsimd/arch/common/xsimd_common_memory.hpp
@@ -442,7 +442,7 @@ namespace xsimd
         load_masked(T const* mem, batch_bool<T, A> mask, convert<T>, Mode, requires_arch<common>) noexcept
         {
             // Scalar fallback: only active lanes are touched. Arches with
-            // hardware predicated loads override this.
+            // hardware predicated loads should override this.
             constexpr std::size_t size = batch<T, A>::size;
             alignas(A::alignment()) std::array<T, size> buffer;
             for (std::size_t i = 0; i < size; ++i)
@@ -462,7 +462,7 @@ namespace xsimd
         store_masked(T* mem, batch<T, A> const& src, batch_bool<T, A> mask, Mode, requires_arch<common>) noexcept
         {
             // Scalar fallback: only active lanes are touched. Arches with
-            // hardware predicated stores override this.
+            // hardware predicated stores should override this.
             constexpr std::size_t size = batch<T, A>::size;
             alignas(A::alignment()) std::array<T, size> src_buf;
             src.store_aligned(src_buf.data());
--- a/include/xsimd/arch/xsimd_avx2.hpp
+++ b/include/xsimd/arch/xsimd_avx2.hpp
@@ -126,10 +126,12 @@ namespace xsimd
             {
                 XSIMD_IF_CONSTEXPR(sizeof(T) == 4)
                 {
+                    static_assert(sizeof(int) == 4, "_mm256_maskload_epi32 requires a 4-byte int");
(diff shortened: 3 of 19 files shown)


serge-sans-paille:
Coild you split the head / tail part in another PR? This one is already quite dense...

include/xsimd/arch/common/xsimd_common_memory.hpp
```diff
@@ -374,6 +375,25 @@ namespace xsimd
...
+        load_masked(T const* mem, batch_bool<T, A> mask, convert<T>, Mode, requires_arch<common>) noexcept
+        {
+            // Per-lane validity contract: only active lanes of ``mem`` are
+            // required to be addressable. An unconditional whole-vector load
+            // would touch inactive lanes and trip ASan/Valgrind on partial
+            // buffers, so stay scalar. Arches with hardware predicated loads
+            // (AVX2 32/64-bit, AVX-512, SVE, RVV) override this with a single
+            // intrinsic that suppresses inactive-lane reads in hardware.
+            constexpr std::size_t size = batch<T, A>::size;
+            alignas(A::alignment()) std::array<T, size> buffer {};
```
serge-sans-paille: to make it worse, building a mask is not always a  single operation depending on the target...

serge-sans-paille: this array assignment forces everything to zero, while some stores are not needed, and the compiler is notable to optimize this away in the generic case

include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -175,6 +172,20 @@ namespace xsimd
...
+        // fall back to the scalar common path: AVX2 has no native maskload for
+        // those widths, and a load-then-blend would break fault-suppression at
+        // page boundaries (the main reason callers ask for a masked load).
+        // Both aligned_mode and unaligned_mode route to the same intrinsic —
+        // masked-off lanes do not fault regardless of alignment.
+        template <class A, class T, class Mode>
+        XSIMD_INLINE std::enable_if_t<std::is_integral<T>::value && (sizeof(T) == 4 || sizeof(T) == 8), batch<T, A>>
+        load_masked(T const* mem, batch_bool<T, A> mask, convert<T>, Mode, requires_arch<avx2>) noexcept
+        {
+            using int_t = std::conditional_t<sizeof(T) == 4, int32_t, long long>;
```
serge-sans-paille: why `long long` and not `int64_t` ?  Tehre's no garantee that `sizeof(long long) == 8`

include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -230,6 +239,20 @@ namespace xsimd
...
+        XSIMD_INLINE std::enable_if_t<std::is_integral<T>::value && (sizeof(T) == 4 || sizeof(T) == 8), void>
+        store_masked(T* mem, batch<T, A> const& src, batch_bool<T, A> mask, Mode, requires_arch<avx2>) noexcept
+        {
+            XSIMD_IF_CONSTEXPR(sizeof(T) == 4)
+            {
+                _mm256_maskstore_epi32(reinterpret_cast<int*>(mem), __m256i(mask), __m256i(src));
+            }
+            else
+            {
+                _mm256_maskstore_epi64(reinterpret_cast<long long*>(mem), __m256i(mask), __m256i(src));
```
serge-sans-paille: ok, I guess that's a constraint of the Intel intrinsic, at least static_assert that `sizeof(long long) ==8` and `sizeof(int) == 4` if you're using this to disntinguish between the two?

include/xsimd/arch/xsimd_sve.hpp
```diff
@@ -101,11 +101,28 @@ namespace xsimd
...
 
-        // load_masked
-        template <class A, class T, bool... Values, class Mode, detail::enable_arithmetic_t<T> = 0>
-        XSIMD_INLINE batch<T, A> load_masked(T const* mem, batch_bool_constant<float, A, Values...>, Mode, requires_arch<sve>) noexcept
+        // load_masked (compile-time mask): build a runtime predicate from
+        // the constant mask and reuse the runtime-mask path. ``pmask`` only
+        // constructs a 128-bit chunk predicate (svdupq_b{8,16,32,64}), which
+        // is replication-based and does not correctly express a per-lane
+        // mask on SVE wider than 128 bits — going through ``as_batch_bool``
+        // gives the right predicate for every vector width. ``int32``/
```
serge-sans-paille: Do you know if the pmask approach would be faster? If so we could still `if constexpr` its usage when the sve size allows it.

include/xsimd/types/xsimd_api.hpp
```diff
@@ -1550,6 +1550,36 @@ namespace xsimd
...
+     * @ingroup batch_data_transfer
+     *
+     * Creates a batch from the buffer \c ptr using a runtime mask. Elements
+     * corresponding to \c false in the mask are not accessed in memory and are
+     * zero-initialized in the resulting batch. No type conversion is performed:
+     * \c ptr must point to \c T. Lanes whose mask bit is \c false do not fault,
+     * so partial loads across a page boundary are safe. \c stream_mode is not
+     * supported.
+     *
+     * \warning Runtime-mask loads carry a significant performance penalty on
```
serge-sans-paille: I don't think we should go into details here:
1. it's difficult to maintain this kind of documentation (what about newly added architectures)
2. we already have the case for other operations and we don't specify it.

I think it's important to communicate that info, but until we have an automated way to do so, better not just throw documentation at it.

include/xsimd/types/xsimd_batch.hpp
```diff
@@ -722,6 +728,26 @@ namespace xsimd
...
 
+    template <class T, class A>
+    template <class Mode>
+    XSIMD_INLINE batch<T, A> batch<T, A>::load(T const* mem, batch_bool<T, A> mask, Mode mode) noexcept
+    {
+        detail::static_check_supported_config<T, A>();
+        static_assert(std::is_same<Mode, aligned_mode>::value || std::is_same<Mode, unaligned_mode>::value,
+                      "supported load mode");
+        constexpr uint64_t full_mask = details::full_mask(size);
+        const auto bits = mask.mask();
```
serge-sans-paille: I'm unsure we want that extra call to `mask` which may be costly, plus the extra tests... if masking is supported, is it beneifical? If it's not, we're already slow...

include/xsimd/types/xsimd_batch.hpp
```diff
@@ -745,6 +771,29 @@ namespace xsimd
...
+    template <class Mode>
+    XSIMD_INLINE void batch<T, A>::store(T* mem,
+                                         batch_bool<T, A> mask,
+                                         Mode mode) const noexcept
+    {
+        detail::static_check_supported_config<T, A>();
+        static_assert(std::is_same<Mode, aligned_mode>::value || std::is_same<Mode, unaligned_mode>::value,
+                      "supported store mode");
+        constexpr uint64_t full_mask = details::full_mask(size);
+        const auto bits = mask.mask();
```
serge-sans-paille: same here

include/xsimd/arch/common/xsimd_common_memory.hpp
DiamonDinoia: Addressed in the latest push — switched the common-arch fallback to use `mask.get(i)` directly instead of materialising `mask.mask()` once and shifting per lane. The `.mask()` call is now gone from the hot path on architectures that fall back to common, so the per-target cost of building the bit mask no longer matters here.

DiamonDinoia: Addressed — dropped the value-init `{}` on the buffer and instead write every lane unconditionally in one pass: `buffer[i] = mask.get(i) ? mem[i] : T(0);`. No double-write on inactive lanes anymore.

include/xsimd/arch/xsimd_avx2.hpp
DiamonDinoia: `long long` is what the Intel intrinsic `_mm256_maskload_epi64` takes (`long long const*`). Using `int64_t*` would force a `reinterpret_cast` at every call site. Added `static_assert(sizeof(long long) == 8, ...)` next to the helpers so the assumption is pinned at compile time — the dispatcher uses the pointer-width to pick the right intrinsic.

include/xsimd/arch/xsimd_sve.hpp
DiamonDinoia: Good question — `pmask` (`svdupq_b{8,16,32,64}`) is replication-based: it builds a 128-bit chunk and replicates it across the SVE register, so it only expresses a per-lane mask correctly when the SVE vector length is exactly 128 bits. For 256/512/1024/2048-bit SVE it would silently produce the wrong predicate. The current path through `as_batch_bool` is correct for every VL and lowers to a single `cmpne` against the integer-domain mask. We could in principle gate `pmask` on `__ARM_FEATURE_SVE_BITS == 128`, but on my qemu build that case lowers to the same `ptrue + cmpne` pair, so it would just be conditional code without a measured win. Happy to add the gate if you have a benchmark that shows a delta on a 128-bit-VL system.

include/xsimd/types/xsimd_api.hpp
DiamonDinoia: Agreed — collapsed the four runtime-mask doxygen blocks down to one short paragraph each (one for `load`, one for `store`), with `\overload` on the unaligned variant. Removed the per-architecture rundown of which targets do/do not have native maskload, since that information rots quickly as the project picks up new arches.

include/xsimd/types/xsimd_batch.hpp
DiamonDinoia: Agreed — dropped the `bits == 0` / `bits == full_mask` early-out in both `batch::load(ptr, batch_bool, mode)` and `batch::store(ptr, batch_bool, mode)`. The runtime-mask member now just forwards straight to `kernel::load_masked` / `kernel::store_masked`. Targets with native predicated instructions (AVX2/AVX-512/SVE/RVV) absorb the all-zero / all-one mask via the hardware predicate, and on the common scalar fallback the per-lane loop handles those cases for free. The extra `mask.mask()` call and the two compares are gone.

include/xsimd/types/xsimd_batch.hpp
DiamonDinoia: Same fix — early-out and `static_assert` are gone from `batch::store(T*, batch_bool, Mode)`; it now forwards directly to `kernel::store_masked`.

DiamonDinoia:
@serge-sans-paille I am now happy with how this looks. I tried to simplify things when possible :)

@claude did a couple of rounds of review

docs/source/api/data_transfer.rst
```diff
@@ -84,3 +84,11 @@ The following empty types are used for tag dispatching:
 
 .. doxygenstruct:: xsimd::unaligned_mode
    :project: xsimd
+
+.. rubric:: Footnotes
+
+.. [#m] Masked ``load`` / ``store`` come in two flavours. The
+   :cpp:class:`batch_bool_constant` overload encodes the mask in the type and
+   is resolved at compile time. The runtime :cpp:class:`batch_bool` overload
+   accepts a mask computed at runtime. Prefer the compile-time mask whenever
```
serge-sans-paille: rewording: For performance reason, prefer the compile-time mask whenever possible.

include/xsimd/arch/common/xsimd_common_memory.hpp
```diff
@@ -437,13 +437,40 @@ namespace xsimd
             return detail::load_masked_common(mem, mask, cvt, mode, detail::masked_memory_uses_fp_bitcast<A, T_in, T_out> {});
         }
 
+        template <class A, class T, class Mode>
+        XSIMD_INLINE batch<T, A>
+        load_masked(T const* mem, batch_bool<T, A> mask, convert<T>, Mode, requires_arch<common>) noexcept
+        {
+            // Scalar fallback: only active lanes are touched. Arches with
+            // hardware predicated loads override this.
```
serge-sans-paille: "should override" ?

include/xsimd/arch/common/xsimd_common_memory.hpp
```diff
@@ -437,13 +437,40 @@ namespace xsimd
...
         {
             detail::store_masked_common(mem, src, mask, mode, detail::masked_memory_uses_fp_bitcast<A, T_in, T_out> {});
         }
 
+        template <class A, class T, class Mode>
+        XSIMD_INLINE void
+        store_masked(T* mem, batch<T, A> const& src, batch_bool<T, A> mask, Mode, requires_arch<common>) noexcept
+        {
+            // Scalar fallback: only active lanes are touched. Arches with
+            // hardware predicated stores override this.
```
serge-sans-paille: same here

include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -117,18 +117,34 @@ namespace xsimd
...
+        // 8/16-bit integers fall back to the common scalar path.
         namespace detail
         {
-            XSIMD_INLINE __m256i maskload(const int32_t* mem, __m256i mask) noexcept
+            template <class T>
+            XSIMD_INLINE __m256i maskload(T const* mem, __m256i mask) noexcept
             {
-                return _mm256_maskload_epi32(mem, mask);
+                XSIMD_IF_CONSTEXPR(sizeof(T) == 4)
+                {
```
serge-sans-paille, line 128: Could you `static_assert(sizeof(int) == 4)` here? It's likely that because the condition is not always constexpr (until we reach C++17) you might need to just assert.

include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -117,18 +117,34 @@ namespace xsimd
...
+                    return _mm256_maskload_epi32(reinterpret_cast<int const*>(mem), mask);
+                }
+                else
+                {
+                    return _mm256_maskload_epi64(reinterpret_cast<long long const*>(mem), mask);
```
serge-sans-paille, line 135: same here for `long long`

include/xsimd/arch/xsimd_avx2.hpp
```diff
@@ -117,18 +117,34 @@ namespace xsimd
...
             }
 
-            XSIMD_INLINE __m256i maskload(const long long* mem, __m256i mask) noexcept
+            template <class T>
+            XSIMD_INLINE void maskstore(T* mem, __m256i mask, __m256i src) noexcept
             {
-                return _mm256_maskload_epi64(reinterpret_cast<long long const*>(mem), mask);
+                XSIMD_IF_CONSTEXPR(sizeof(T) == 4)
+                {
+                    _mm256_maskstore_epi32(reinterpret_cast<int*>(mem), mask, src);
```
serge-sans-paille, line 145: and there

include/xsimd/arch/xsimd_avx512bw.hpp
```diff
@@ -378,6 +378,53 @@ namespace xsimd
...
+        // load_masked / store_masked: native vmovdqu8 / vmovdqu16 predication for
+        // 8/16-bit, replacing the common scalar fallback. No aligned masked 8/16
+        // intrinsic exists and masked moves never fault, so loadu fits both modes.
+        template <class A, class T, class Mode,
+                  class = std::enable_if_t<std::is_integral<T>::value && (sizeof(T) == 1 || sizeof(T) == 2)>>
+        XSIMD_INLINE batch<T, A> load_masked(T const* mem, batch_bool<T, A> mask, convert<T>, Mode, requires_arch<avx512bw>) noexcept
+        {
+            XSIMD_IF_CONSTEXPR(sizeof(T) == 1)
+            {
+                return _mm512_maskz_loadu_epi8((__mmask64)mask.mask(), mem);
```
serge-sans-paille, line 390: interestingly there's no  `_mm512_maskz_load_epi8` :-)

include/xsimd/arch/xsimd_isa.hpp
```diff
@@ -74,14 +74,23 @@
...
+#include "./xsimd_avx512vl_128.hpp"
+#include "./xsimd_avx512vl_256.hpp"
+#endif
+
 #if XSIMD_WITH_AVX512F
 #include "./xsimd_avx512f.hpp"
 #endif
 
 #if XSIMD_WITH_AVX512VL
 #include "./xsimd_avx512vl.hpp"
```
serge-sans-paille: any reason for not moving the remaining include above?

serge-sans-paille (commented):
A few minor nits but looks good to me otherwise.
No masked load / store on Neon, I haven't checked on VSX & VXE,  @Andreas-Krebbel ?

include/xsimd/arch/xsimd_avx512bw.hpp
DiamonDinoia: I mean anything is aligned to 8 bits no. They could provide either aligned or unaligned but this makes it explicit as unaligned is always safe :)

Andreas-Krebbel:
> A few minor nits but looks good to me otherwise. No masked load / store on Neon, I haven't checked on VSX & VXE, @Andreas-Krebbel ?

Power (VSX) should have masked loads/stores with Power10. s390x (VXE) does not have masked vector instructions. But I guess we could implement some special cases with vector store/load with length.

AntoinePrv:
I support **not** adding the feature when the hardware does not support it (e.g. Neon), as I understand it is been done here.

IIUC (at least with SVE) this is not "just another operation" but a fundamental different way to write algorithms. For instance using a single masked SIMD loop and skipping the trailer loop because the mask load will not touch masked memory.

So I'd rather statically dispatch between two implementations (maybe with a `xsimd::has_mask_load_v<Arch>`).

DiamonDinoia:
> I support **not** adding the feature when the hardware does not support it (e.g. Neon), as I understand it is been done here.
> 
> IIUC (at least with SVE) this is not "just another operation" but a fundamental different way to write algorithms. For instance using a single masked SIMD loop and skipping the trailer loop because the mask load will not touch masked memory.
> 
> So I'd rather statically dispatch between two implementations (maybe with a `xsimd::has_mask_load_v<Arch>`).

You know more about sve then me so I followed the suggestion.

JohanMabille (approved):
LGTM, and I think all comments from @serge-sans-paille have been addressed, so let's merge?

PR #1334, opened 2026-04-28
Feat/mulhilo

DiamonDinoia:
Some toolchains (notably certain GCC builds) define shift and mul
immediate intrinsics as macros that apply a textual C-style cast to
their operand. That cast does not traverse the multi-level alias
inheritance of simd_register (e.g. avx512bw -> avx512dq -> avx512cd
-> avx512f), so a batch<T, ISA> fails to convert to its native
register type in those contexts.

Declare the conversion operator on batch itself so the native type is
always one member-lookup away.


DiamonDinoia:
@AntoinePrv I optimized the cases I use in my RNG but it might be optimized further. What do you think?

include/xsimd/types/xsimd_batch.hpp
```diff
@@ -129,6 +129,18 @@ namespace xsimd
...
+         * types::simd_register when a batch is fed to an intrinsic defined as
+         * a macro (e.g. certain GCC shift/mullo imm intrinsics), because the
+         * textual C-style cast inside the macro does not traverse the alias
+         * inheritance chain. Declaring the operator here makes it visible on
+         * the batch type directly. */
+        XSIMD_INLINE operator register_type() const noexcept
+        {
+            return this->data;
+        }
+
```
AntoinePrv, line 143: I've been wondering if instead of conversion operator, we should rather aim for a method `batch::raw() -> register_type` (that we roll out progressively).
I have also found that relying on conversions makes it a bit harder to follow what is going on, and is definitely the sort of things where we see differences between compilers. What do you think?
CC @serge-sans-paille @JohanMabille

include/xsimd/types/xsimd_api.hpp
```diff
@@ -1744,6 +1745,60 @@ namespace xsimd
         return x * y;
     }
 
+    /**
+     * @ingroup batch_arithmetic
+     *
+     * Computes the low N bits of the 2N-bit product of integer batches \c x and \c y.
+     * Equivalent to ``mul(x, y)`` — the low half of the full product is identical for
+     * both signed and unsigned interpretations.
```
AntoinePrv, line 1753: Can we extend this a little bit, it was not immediately clear to me. Perhaps adding something like 

> The function behaves as if it computes the per-slot product of x and y using double the input width as an intermediate representation (e.g. 64 bits for 32-bit inputs), then returning the lower half of the result (the lower 32 bits in this example).

PR #1339, opened 2026-05-04
Add clang workflow for IBM Z (s390x)

Andreas-Krebbel:
Local test was successful.

### DIFF
--- a/.github/workflows/cross-s390x.yml
+++ b/.github/workflows/cross-s390x.yml
@@ -1,4 +1,4 @@
-name: IBM Z cross-compilation build
+name: IBM Z (s390x) cross-compilation build
 on: [push, pull_request]
 concurrency:
   group: ${{ github.workflow }}-${{ github.job }}-${{ github.ref }}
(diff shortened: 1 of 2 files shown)


Andreas-Krebbel:
I did run the workflow several times in a row without issues locally. Seems to run endlessly here, sorry for that.

serge-sans-paille:
I've restarted the hanging job manually

.github/workflows/cross-s390x.yml
```diff
@@ -43,6 +54,8 @@ jobs:
               -DCMAKE_TOOLCHAIN_FILE=${{ github.workspace }}/.github/toolchains/${{ matrix.sys.compiler }}-${{ matrix.target.dir }}.cmake
     - name: Build
       run: cmake --build build/ --verbose -j1
+    - name: Set CPU feature test expectations
```
serge-sans-paille, line 57: This is puzzling ^^

Andreas-Krebbel: Indeed. This used to be like this in cross-ppc.yml:
```
     - name: Set CPU feature test expectations
       run: |
```
I've changed that to the following since the above didn't work with 'act':
```
     - name: Set CPU feature test expectations
       run: /bin/true
```
Looks like it is supposed to be some placeholder to add stuff later. I could try removing it entirely, but I would be surprised if that's the culprit. It seems to be the qemu process hanging for some reason, but only if the test binary has been built with clang.

PR #1340, opened 2026-05-04
Add clang workflow for PPC

Andreas-Krebbel:
Trying to test this locally did not finish. The qemu process running the clang built test_xsimd binaries didn't finish. However, I could not reproduce it on real hardware. So this one is only to let the real github runners run against it.


JohanMabille (approved):
Worked like a charm, thanks!

PR #1342, opened 2026-05-06
Fix C++ version detection for MSVC

AntoinePrv:


AntoinePrv:
@JohanMabille

serge-sans-paille:
LGTM. The fact that we don't detect that in CI is a problem (and a potential regression pithole). Does that mean this PR should also add msvc in the compilers tested in `.github/workflows/cxx-versions.yml` ?

JohanMabille (approved):

AntoinePrv:
@serge-sans-paille I added MSVC for this build.

There is a bit of a self-referential problem here, tests were using `__cplusplus` to exclude checks for features with higher C++ requirement, so even with a MSVC build, they would just have been skiped as well

On top I find the CI starting to last long. I've started to think we should improve our strategy. For instance, mixing AVX2 check with C++17 *etc* -- same way we don't build the whole (compiler version, micro arch) matrix.
Perhaps even some smaller subset in PRs and more in main? I'll come back on this topic in another issue though.

PR #1344, opened 2026-05-06
Feat/mulhilo

DiamonDinoia:
Successor to #1334 (closed). Adds mulhi/mullo/mulhilo for integer batches; CI split for RVV cross-compile to use qemu 11 (arch container) at vlen=128 due to known qemu RVV slowdowns at vlen=128 in qemu < 11 (QEMU issue #2137). All 20 fork CI jobs green on this branch.

Took me forever to find the issue as I am using qemu 11 locally...


include/xsimd/types/xsimd_batch.hpp
```diff
@@ -130,6 +129,18 @@ namespace xsimd
...
         XSIMD_INLINE batch(register_type reg) noexcept;
 
+        /* Redeclare the conversion operator at the most-derived level. Some
+         * compilers fail to invoke the conversion inherited from
+         * types::simd_register when a batch is fed to an intrinsic defined as
+         * a macro (e.g. certain GCC shift/mullo imm intrinsics), because the
+         * textual C-style cast inside the macro does not traverse the alias
+         * inheritance chain. Declaring the operator here makes it visible on
+         * the batch type directly. */
+        XSIMD_INLINE operator register_type() const noexcept
```
serge-sans-paille, line 149: you could use `using types::simd_register<T, A>::operator register_type` instead.

test/test_batch.cpp
```diff
@@ -21,6 +21,49 @@
...
+    typename std::enable_if<std::is_integral<T>::value && (sizeof(T) <= 4), T>::type
+    mulhi_reference(T x, T y) noexcept
+    {
+        using W = typename std::conditional<std::is_signed<T>::value, int64_t, uint64_t>::type;
+        return static_cast<T>((static_cast<W>(x) * static_cast<W>(y)) >> (8 * sizeof(T)));
+    }
+
+#if defined(__SIZEOF_INT128__)
+    template <class T>
+    typename std::enable_if<std::is_integral<T>::value && (sizeof(T) == 8), T>::type
```
serge-sans-paille: I fear the `is_integral<T>`` test might fail here, as this type is not a standard type, see https://godbolt.org/z/3nd7KEj9K (works on gcc though)

test/test_batch.cpp
```diff
@@ -313,6 +356,145 @@ struct batch_test
...
+        auto run_case = [](array_type const& a, array_type const& b, const char* tag)
+        {
+            batch_type ba = batch_type::load_unaligned(a.data());
+            batch_type bb = batch_type::load_unaligned(b.data());
+
+            array_type lo_expected;
+            array_type hi_expected;
+            for (std::size_t i = 0; i < size; ++i)
+            {
+                lo_expected[i] = static_cast<value_type>(static_cast<UT>(a[i]) * static_cast<UT>(b[i]));
```
serge-sans-paille, line 376: Can you explain this intermediate cast to an unsigend type?

.github/workflows/cross-rvv-arch.yml
```diff
@@ -0,0 +1,85 @@
...
+# Empirically:
+#   qemu 8.2.2 (Ubuntu 24.04 apt)    : test_xsimd at vlen=128 times out
+#   qemu 9.2.1 (Ubuntu 25.04 plucky) : ditto
+#   qemu 10.0.8 (Debian trixie)      : ditto
+#   qemu 11.0.0 (Arch) + gcc 15.1    : 367 cases / 5664 asserts in <10 min
+#
+# So vlen=128 RVV coverage lives in this workflow, which runs the build
+# and test inside an `archlinux:latest` container (qemu 11 + gcc 15.1).
+# The matching ubuntu-runner workflow `cross-rvv.yml` keeps multi-compiler
+# matrix coverage (gcc-14, clang-17/18) for vlens >= 256, where the apt
```
serge-sans-paille, line 22: I guess there's no way to have all compilers running on archlinux?

docs/source/api/arithmetic_index.rst
```diff
@@ -40,6 +40,12 @@ Binary operations:
 +---------------------------------------+----------------------------------------------------+
 | :cpp:func:`mul`                       | per slot multiply                                  |
 +---------------------------------------+----------------------------------------------------+
+| :cpp:func:`mullo`                    | low N bits of the 2N-bit integer product           |
```
serge-sans-paille: indentation seems odd.

docs/source/api/arithmetic_index.rst
```diff
@@ -40,6 +40,12 @@ Binary operations:
...
++---------------------------------------+----------------------------------------------------+
+| :cpp:func:`mulhi`                    | high N bits of the 2N-bit integer product          |
```
serge-sans-paille: same here

include/xsimd/types/xsimd_api.hpp
```diff
@@ -1744,6 +1745,75 @@ namespace xsimd
...
+     * The function behaves as if it computes the per-slot product of \c x and \c y using
+     * double the input width as an intermediate representation (e.g. 64 bits for 32-bit
+     * inputs), then returns the lower half of the result (the lower 32 bits in this
+     * example).
+     * @tparam T integer element type of the batch.
+     * @param x batch involved in the product.
+     * @param y batch involved in the product.
+     * @return the low N bits of the product, lane-wise.
+     */
+    template <class T, class A, class = typename std::enable_if<std::is_integral<T>::value, void>::type>
```
serge-sans-paille: `void` is already the default second parameter for `std::enable_if`

include/xsimd/types/xsimd_api.hpp
```diff
@@ -1744,6 +1745,75 @@ namespace xsimd
...
+     * @param x batch involved in the product.
+     * @param y batch involved in the product.
+     * @return pair of batches ``{hi, lo}``.
+     */
+    template <class T, class A, class = typename std::enable_if<std::is_integral<T>::value, void>::type>
+    XSIMD_INLINE std::pair<batch<T, A>, batch<T, A>>
+    mulhilo(batch<T, A> const& x, batch<T, A> const& y) noexcept
+    {
+        detail::static_check_supported_config<T, A>();
+        return std::pair<batch<T, A>, batch<T, A>> { mulhi<T, A>(x, y), x * y };
```
serge-sans-paille: on arm, Neon has widening multiply, which is slightly similar to what you propose here, but with a different ordering. See for instance https://developer.arm.com/architectures/instruction-sets/intrinsics/vmull_high_s32 and https://developer.arm.com/architectures/instruction-sets/intrinsics/vmull_s32.

On Intel you can produce a similar behavior with `_mm256_madd_epi16` or `_mm_madd_epi16`  (which require some interleaving with zeros though).

Given that it's trivial to reproduce the behavior of the `mulhilo` you propose using `mulhi` and `mullo`, I wonder if we shouldn't propose a `widening_mulhi` and `widening_mullo` instead? Better names accepted

include/xsimd/arch/xsimd_sse2.hpp
```diff
@@ -1438,6 +1438,18 @@ namespace xsimd
...
+        template <class A>
+        XSIMD_INLINE batch<int16_t, A> mulhi(batch<int16_t, A> const& self, batch<int16_t, A> const& other, requires_arch<sse2>) noexcept
+        {
+            return _mm_mulhi_epi16(self, other);
+        }
+        template <class A>
+        XSIMD_INLINE batch<uint16_t, A> mulhi(batch<uint16_t, A> const& self, batch<uint16_t, A> const& other, requires_arch<sse2>) noexcept
+        {
+            return _mm_mulhi_epu16(self, other);
+        }
```
serge-sans-paille, line 1498: note for self: we would implement the 8 bit version of all those multiply, don't know if it makes sense though.

include/xsimd/arch/xsimd_sse4_1.hpp
```diff
@@ -359,6 +359,42 @@ namespace xsimd
             }
         }
 
+        // mulhi
+        template <class A>
+        XSIMD_INLINE batch<int32_t, A> mulhi(batch<int32_t, A> const& self, batch<int32_t, A> const& other, requires_arch<sse4_1>) noexcept
+        {
+            __m128i even = _mm_mul_epi32(self, other); // 64-bit products in lanes 0,2
+            __m128i odd = _mm_mul_epi32(_mm_shuffle_epi32(self, _MM_SHUFFLE(3, 3, 1, 1)),
```
serge-sans-paille: you could just `_mm_srli_epi32(self, 32)` instead, right?

serge-sans-paille: and same for `other`, of course

include/xsimd/arch/xsimd_sse4_1.hpp
```diff
@@ -359,6 +359,42 @@ namespace xsimd
...
+                                        _mm_shuffle_epi32(other, _MM_SHUFFLE(3, 3, 1, 1)));
+            // hi halves in the low 32 of each 64 lane of (even>>32), and in the high 32 of odd
+            __m128i even_hi = _mm_srli_epi64(even, 32);
```
serge-sans-paille, line 369: you could avoid this shift by using directly a call to `_mm_shuffle_ps ` on `even` and `odd`

include/xsimd/arch/xsimd_sse4_1.hpp
```diff
@@ -359,6 +359,42 @@ namespace xsimd
...
+            // blend: 32-bit lanes {even_hi[0], odd_hi[1], even_hi[2], odd_hi[3]}
+            return _mm_blend_epi16(even_hi, odd, 0xCC);
+        }
+        template <class A>
+        XSIMD_INLINE batch<uint32_t, A> mulhi(batch<uint32_t, A> const& self, batch<uint32_t, A> const& other, requires_arch<sse4_1>) noexcept
+        {
+            __m128i even = _mm_mul_epu32(self, other);
+            __m128i odd = _mm_mul_epu32(_mm_srli_epi64(self, 32), _mm_srli_epi64(other, 32));
+            __m128i even_hi = _mm_srli_epi64(even, 32);
```
serge-sans-paille, line 378: same here for blend vs. shuffle. You did use `_mm_srli_epi64` here though, so there is probably a reason I don't get ^^!

include/xsimd/arch/xsimd_neon.hpp
```diff
@@ -1037,6 +1037,54 @@ namespace xsimd
...
+        /*********
+         * mulhi *
+         *********/
+
+        template <class A>
+        XSIMD_INLINE batch<int8_t, A> mulhi(batch<int8_t, A> const& lhs, batch<int8_t, A> const& rhs, requires_arch<neon>) noexcept
+        {
+            int16x8_t lo = vmull_s8(vget_low_s8(lhs), vget_low_s8(rhs));
+            int16x8_t hi = vmull_s8(vget_high_s8(lhs), vget_high_s8(rhs));
+            return vcombine_s8(vshrn_n_s16(lo, 8), vshrn_n_s16(hi, 8));
```
serge-sans-paille: What about `vuzpq_s8` (and same below)

serge-sans-paille (commented):
First round of comments, I haven't reviewed the avx512 not the rvv part yet.

include/xsimd/types/xsimd_batch.hpp
DiamonDinoia: > you could use `using types::simd_register<T, A>::operator register_type` instead.

**Tried it, had to revert.** The using-declaration is evaluated at the
point of class-template instantiation, not lazily on use. But
`types::simd_register<T, A>` only carries `operator register_type` for
*supported* `(T, A)` pairs — for unsupported pairs (e.g. `batch<double,
neon>` on 32-bit ARM, where `double` isn't a native NEON element type)
the primary template is empty. With the using-decl in place, code that
merely *names* `batch<double, neon>` (xsimd's arch dispatcher walks the
list and instantiates each `batch<T, Arch>` it considers) fails to
compile:

```
error: 'operator xsimd::batch<double, xsimd::neon>::register_type' has
       not been declared in 'struct xsimd::types::simd_register<double,
       xsimd::neon>'
```

A redefined member operator is only instantiated when called, so
unsupported pairs stay well-formed up to the point a user actually
uses them. Restored the redefined inline operator and added a comment
explaining the constraint so this isn't re-suggested.

(Caught by the aarch64 cross-build / qemu run; it slips through the
x86 native run because the dispatcher there never walks `neon`.)

test/test_batch.cpp
DiamonDinoia: > I fear the `is_integral<T>` test might fail here, as this type is not a standard type, see https://godbolt.org/z/3nd7KEj9K (works on gcc though)

The reference helper is only ever instantiated with `T` ∈ `{int8_t, …,
int64_t, uint8_t, …, uint64_t}` — those are the `value_type`s of the
xsimd integer batches that exercise this path. None of them are
`__int128`, so `std::is_integral<T>` is well-defined and `true` on all
supported toolchains. `__int128` only appears as the *intermediate
widening type* `W` inside the body, gated on `__SIZEOF_INT128__`; we
never feed `__int128` through `std::is_integral`.

test/test_batch.cpp
DiamonDinoia: > Can you explain this intermediate cast to an unsigned type?

Signed integer overflow is UB in C++, and the test explicitly feeds
`vmin*vmin`, `vmin*-1`, etc. — pairs that overflow the signed range.
Casting to the corresponding unsigned type makes the multiplication
well-defined (modulo 2^N), and the low N bits of an N×N→2N product are
bit-identical for the signed and unsigned interpretations, so the
unsigned wrap value is the correct expected result for `mullo`.

I've added a short comment to that effect at the call site so a future
reader doesn't have to reconstruct this.

.github/workflows/cross-rvv-arch.yml
DiamonDinoia: > I guess there's no way to have all compilers running on archlinux?

There migth be a way. I am not an arch expert though. Maybe a follow-up
PR to clean this?

docs/source/api/arithmetic_index.rst
DiamonDinoia: > indentation seems odd. / same here

**Fixed.** The `mullo` and `mulhi` rows were one column short, breaking
the right-hand pipe. Padded both rows to match the rest of the table.

docs/source/api/arithmetic_index.rst
DiamonDinoia: > indentation seems odd. / same here

**Fixed.** The `mullo` and `mulhi` rows were one column short, breaking
the right-hand pipe. Padded both rows to match the rest of the table.

include/xsimd/types/xsimd_api.hpp
DiamonDinoia: > `void` is already the default second parameter for `std::enable_if`

**Fixed.** Dropped `, void` from the three sites (`mullo`, `mulhi`,
`mulhilo`) — now `std::enable_if<std::is_integral<T>::value>::type`.

include/xsimd/types/xsimd_api.hpp
DiamonDinoia: > on arm, Neon has widening multiply […] Given that it's trivial to
> reproduce the behavior of the `mulhilo` you propose using `mulhi` and
> `mullo`, I wonder if we shouldn't propose a `widening_mulhi` and
> `widening_mullo` instead?

Two separate questions, let me split them:

1. **Should `mulhilo` exist at all?** I'd argue yes, because there is a
   real perf gap: on most ISAs the natural way to compute the high half
   already produces the low half as a by-product (e.g. AArch64's
   `smull/umull` returns the full 2N product; SSE4.1 `pmuldq`/`pmuludq`
   likewise). Calling `mulhi` and `mullo` separately forces two
   independent multiply chains, whereas `mulhilo` lets the kernel
   return both halves from a single set of products. Concretely, that's
   what every counter-based RNG (Philox, Threefry, …) wants — they
   need the full 2N product — and right now they have to do it
   themselves with intrinsics. The fallback `{ mulhi(x,y), x*y }` is
   only the *generic* path; arch overrides can (and should) be added.

2. **Should we also add `widening_mulhi`/`widening_mullo`?** That's a
   different API — it changes the *element width* of the result, the
   way Neon's `vmull_*` does. It composes nicely with `mulhi`/`mullo`
   for users who want 2N output as a *wider batch* rather than two
   N-wide batches. I'm in favour of adding it as a follow-up PR,
   because:

   - the lane interleaving is non-trivial on x86 (SSE has no native
     widening multiply for 32×32→64 except `pmuldq`/`pmuludq`, which
     only do the even lanes — and AVX-512 finally adds `vpmullq` but
     not a true widening),
   - and the result type (`batch<wider_t, A>`) opens up a separate
     trait/dispatch question that I'd rather not couple with this PR.

So my proposal: keep `mulhi/mullo/mulhilo` in this PR (with the arch
overrides), and open a separate issue for `widening_mul*`.

include/xsimd/arch/xsimd_sse2.hpp
DiamonDinoia: > note for self: we would implement the 8 bit version of all those
> multiply, don't know if it makes sense though.

**Done — added at the SSE2 baseline** (so SSE4.1, AVX, … inherit it):

- `mul<uint8_t>` (the body) — split each 16-bit lane into even/odd
  bytes, do two 16-bit `pmullw`s, re-interleave with `pand`/`psllw`/`por`.
  No 8-bit-blend needed (avoids requiring SSE4.1).
- `mul<int8_t>` is a one-line `bitwise_cast` forwarder to the unsigned
  body, since the low N bits of an N×N→2N product are bit-identical
  for both signed and unsigned. Codegen for the two overloads is
  *byte-identical* (gcc 15 folds the cast):

  ```
  mul_i8_sse:  psrlw ; psrlw ; pmullw ; psrlw ; pand ; pand ; pmullw ; psllw ; pand ; por
  mul_u8_sse:  psrlw ; psrlw ; pmullw ; psrlw ; pand ; pand ; pmullw ; psllw ; pand ; por
  ```

- `mulhi<int8_t>` and `mulhi<uint8_t>` get separate bodies because the
  results *aren't* bit-identical (sign- vs zero-extension changes the
  product). Both use the standard widen→`pmullw`→narrow pattern.
  Codegen:

  ```
  mulhi_i8:  unpcklbw + unpckhbw ; psraw ×2 ; pmullw ×2 ; psraw ×2 ; packsswb
  mulhi_u8:  unpcklbw/unpckhbw against zero ; pmullw ×2 ; psrlw ×2 ; packuswb
  ```

AVX2 and AVX-512BW already had templated 8-bit `mul`. They now also
get **native-width** 8-bit `mulhi` overloads (added in this PR rather
than left as a follow-up): direct ports of the SSE2 algorithm to
`_mm256_*` and `_mm512_*` intrinsics. The per-128-bit-lane
`vpunpcklbw` / `vpunpckhbw` and `vpacksswb` / `vpackuswb` form an
inverse pair within each lane, so no inter-lane `vpermq` /
`vpermq2`-style fix-up is needed — the byte ordering is preserved end
to end. Codegen on Haswell / Skylake-X:

```
mulhi_i8 (avx2):     vpunpcklbw + vpunpckhbw ×2 ; vpsraw ×4 ; vpmullw ×2 ; vpsraw ×2 ; vpacksswb
mulhi_u8 (avx2):     vpunpcklbw + vpunpckhbw against zero ; vpmullw ×2 ; vpsrlw ×2 ; vpackuswb
mulhi_i8 (avx512bw): same shape on ZMM
mulhi_u8 (avx512bw): same shape on ZMM
```

include/xsimd/arch/xsimd_sse4_1.hpp
DiamonDinoia: > you could just `_mm_srli_epi32(self, 32)` instead, right? / and same for `other`, of course

`_mm_srli_epi32(x, 32)` would clear every 32-bit lane to zero (the SSE
intrinsic saturates past-width shifts to 0). I think you meant
`_mm_srli_epi64(x, 32)` — same as the unsigned overload right below
already does. **Done** — switched the signed `mulhi<int32>` to use
`_mm_srli_epi64(self, 32)` and `_mm_srli_epi64(other, 32)` to produce
the odd-lane operands, matching the `uint32` overload. `pmuldq` reads
the low 32 bits of each 64-bit lane as signed, and `srli_epi64` puts
the original odd lane there with zero-extended high bits, so signedness
is preserved.

Codegen comparison (`g++ -O3 -msse4.2 -mno-avx`):

```
before (signed):  pshufd ; pshufd ; pmuldq ; pmuldq ; psrlq ; pblendw
after  (signed):  psrlq  ; psrlq  ; pmuldq ; pmuldq ; psrlq ; pblendw
unsigned (unchanged):  psrlq ; psrlq ; pmuludq ; pmuludq ; psrlq ; pblendw
```

Same op count, but the signed and unsigned overloads are now structural
duplicates — easier to read and easier to keep in sync.

include/xsimd/arch/xsimd_sse4_1.hpp
DiamonDinoia: > you could avoid this shift by using directly a call to `_mm_shuffle_ps` on `even` and `odd`. / same here for blend vs. shuffle. You did use `_mm_srli_epi64` here though, so there is probably a reason I don't get ^^!

I tried this and it doesn't fall out as a single instruction:
`_mm_shuffle_ps(even, odd, mask)` always picks two lanes from `even`
then two from `odd`, so the best we get in one shuffle is
`[E0hi, E1hi, O0hi, O1hi]` — but the `mulhi` lane order has to be
`[E0hi, O0hi, E1hi, O1hi]` (lane *i* = high half of `x[i]*y[i]`).
Recovering the right interleave costs a second shuffle, which is the
same op-count as the current `srli_epi64` + `blend_epi16` and adds an
int<->fp domain crossing on most µarchs.

Codegen confirms (gcc 15, `-O3 -msse4.2 -mno-avx`):

```
current (srli + pblendw):
    movdqa ; psrlq ; pmuldq ; psrlq ; pmuldq ; psrlq ; pblendw     (7 ops)

shufps + fixup pshufd:
    movdqa ; psrlq ; pmuldq ; psrlq ; pmuldq ; shufps ; pshufd     (7 ops)

blendps (same shape, different blend):
    movdqa ; psrlq ; pmuldq ; psrlq ; pmuldq ; psrlq ; blendps     (7 ops)
```

Three 7-op sequences. Same instruction count, but the `shufps + pshufd`
variant pins **two** ops to port-5 (Skylake/Alderlake) on top of the
two `pmuldq`s, which already use port 5 — the `srli + pblendw` keeps
the trailing pair on the shift/blend ports (0/1/5) instead. So no win,
and a likely loss under back-to-back issue. Keeping the current
sequence.

include/xsimd/arch/xsimd_neon.hpp
DiamonDinoia: > What about `vuzpq_s8` (and same below)

**Done.** I am not super familiar with neon/arm so this slipped.
Switched all six byte/half/word `mulhi` overloads to
`vuzpq_*(reinterpret(lo), reinterpret(hi)).val[1]`, dropping the pair
of `vshrn_n_*` + `vcombine_*`.

Codegen on aarch64 (`aarch64-linux-gnu-g++-15 -O3`):

```
mulhi int8 :   smull v30.8h, ... ; smull v31.8h, ... ; uzp2 v31.16b, ... ; str
mulhi int16:   smull v30.4s, ... ; smull v31.4s, ... ; uzp2 v31.8h , ... ; str
mulhi int32:   smull v30.2d, ... ; smull v31.2d, ... ; uzp2 v31.4s , ... ; str
```

Two widening multiplies + a single `uzp2` per overload. The high lane
of each widened product lives at every odd index of `[lo | hi]`, so
`uzp2` extracts them directly.

include/xsimd/types/xsimd_api.hpp
serge-sans-paille: ok for postponing widening mul .

API note: we already have `zip_lo` and `zip_hi` so we should probably have `mul_hi` and `mul_lo` (and then `mul_hilo`.

Your argument for `mul_hi` wrt. performance is strange, because we don't specialize the kernel, looks just like an helper o_O ?

serge-sans-paille:
I'm happy with current state, final touches:
1. rebase on master branch and fix conflicts
2. amend commit to reflect mulhi -> mul_hi

serge-sans-paille:
thanks :bow:

PR #1345, opened 2026-05-10
Tentative support for avx512vl extensions to 256 bit registers

serge-sans-paille:
In addition to missing instructions (e.g. bas on int64_t etc) this mostly changes the mask representation from vector register to scalar, thus the big diff.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -32,6 +32,7 @@ jobs:
           - { compiler: 'clang', version: '18', flags: 'avx512' }
           - { compiler: 'clang', version: '18', flags: 'avx_128' }
           - { compiler: 'clang', version: '18', flags: 'avx2_128' }
+          - { compiler: 'clang', version: '18', flags: 'avx512vl_256' }
     steps:
     - name: Setup compiler
       if: ${{ matrix.sys.compiler == 'gcc' }}
@@ -93,6 +94,9 @@ jobs:
         fi
         if [[ '${{ matrix.sys.flags }}' == 'avx512' ]]; then
           CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=skylake-avx512"
+        fi
+        if [[ '${{ matrix.sys.flags }}' == 'avx512vl_256' ]]; then
+          CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=skylake-avx512 -DXSIMD_DEFAULT_ARCH=avx512vl_256"
         fi
         if [[ '${{ matrix.sys.flags }}' == 'avx512pf' ]]; then
           CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=knl"
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -81,13 +81,15 @@ jobs:
           CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=sandybridge"
         fi
         if [[ '${{ matrix.sys.flags }}' == 'avx_128' ]]; then
-          CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=sandybridge -DXSIMD_DEFAULT_ARCH=avx_128"
+          CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=sandybridge"
+          CXXFLAGS="$CXX_FLAGS -DXSIMD_DEFAULT_ARCH=avx_128"
         fi
         if [[ '${{ matrix.sys.flags }}' == 'avx2' ]]; then
           CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=haswell"
         fi
         if [[ '${{ matrix.sys.flags }}' == 'avx2_128' ]]; then
-          CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=haswell -DXSIMD_DEFAULT_ARCH=avx2_128"
+          CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=haswell"
+          CXXFLAGS="$CXX_FLAGS -DXSIMD_DEFAULT_ARCH=avx2_128"
         fi
         if [[ '${{ matrix.sys.flags }}' == 'sse3' ]]; then
           CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=nocona"
@@ -96,7 +98,8 @@ jobs:
           CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=skylake-avx512"
         fi
         if [[ '${{ matrix.sys.flags }}' == 'avx512vl_256' ]]; then
-          CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=skylake-avx512 -DXSIMD_DEFAULT_ARCH=avx512vl_256"
+          CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=skylake-avx512"
+          CXXFLAGS="$CXX_FLAGS -DXSIMD_DEFAULT_ARCH=avx512vl_256"
         fi
         if [[ '${{ matrix.sys.flags }}' == 'avx512pf' ]]; then
           CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=knl"
--- a/include/xsimd/arch/xsimd_avx512vl_256.hpp
+++ b/include/xsimd/arch/xsimd_avx512vl_256.hpp
@@ -1,0 +1,728 @@
+/***************************************************************************
+ * Copyright (c) Johan Mabille, Sylvain Corlay, Wolf Vollprecht and         *
+ * Martin Renou                                                             *
+ * Copyright (c) QuantStack                                                 *
+ * Copyright (c) Serge Guelton                                              *
+ * Copyright (c) Marco Barbone                                              *
+ *                                                                          *
+ * Distributed under the terms of the BSD 3-Clause License.                 *
+ *                                                                          *
+ * The full license is in the file LICENSE, distributed with this software. *
+ ****************************************************************************/
+
+#ifndef XSIMD_AVX512VL_256_HPP
+#define XSIMD_AVX512VL_256_HPP
+
+#include "../types/xsimd_avx512vl_register.hpp"
+#include "../types/xsimd_batch_constant.hpp"
+
+#include <type_traits>
+
+namespace xsimd
+{
+    namespace kernel
+    {
+        using namespace types;
+
+        namespace detail
+        {
+            template <class A, class T, int Cmp>
+            XSIMD_INLINE batch_bool<T, A> compare_int_avx512vl_256(batch<T, A> const& self, batch<T, A> const& other) noexcept
+            {
+                using register_type = typename batch_bool<T, A>::register_type;
+                if (std::is_signed<T>::value)
+                {
+                    XSIMD_IF_CONSTEXPR(sizeof(T) == 1)
+                    {
+                        // shifting to take sign into account
+                        uint64_t mask_low0 = _mm256_cmp_epi32_mask((batch<int32_t, A>(self.data) & batch<int32_t, A>(0x000000FF)) << 24,
+                                                                   (batch<int32_t, A>(other.data) & batch<int32_t, A>(0x000000FF)) << 24,
+                                                                   Cmp);
+                        uint64_t mask_low1 = _mm256_cmp_epi32_mask((batch<int32_t, A>(self.data) & batch<int32_t, A>(0x0000FF00)) << 16,
+                                                                   (batch<int32_t, A>(other.data) & batch<int32_t, A>(0x0000FF00)) << 16,
+                                                                   Cmp);
+                        uint64_t mask_high0 = _mm256_cmp_epi32_mask((batch<int32_t, A>(self.data) & batch<int32_t, A>(0x00FF0000)) << 8,
+                                                                    (batch<int32_t, A>(other.data) & batch<int32_t, A>(0x00FF0000)) << 8,
+                                                                    Cmp);
+                        uint64_t mask_high1 = _mm256_cmp_epi32_mask((batch<int32_t, A>(self.data) & batch<int32_t, A>(0xFF000000)),
+                                                                    (batch<int32_t, A>(other.data) & batch<int32_t, A>(0xFF000000)),
+                                                                    Cmp);
+                        uint64_t mask = 0;
+                        for (unsigned i = 0; i < 8; ++i)
+                        {
+                            mask |= (mask_low0 & (uint64_t(1) << i)) << (3 * i + 0);
+                            mask |= (mask_low1 & (uint64_t(1) << i)) << (3 * i + 1);
+                            mask |= (mask_high0 & (uint64_t(1) << i)) << (3 * i + 2);
+                            mask |= (mask_high1 & (uint64_t(1) << i)) << (3 * i + 3);
+                        }
+                        return (register_type)mask;
+                    }
+                    else XSIMD_IF_CONSTEXPR(sizeof(T) == 2)
+                    {
+                        // shifting to take sign into account
+                        uint16_t mask_low = _mm256_cmp_epi32_mask((batch<int32_t, A>(self.data) & batch<int32_t, A>(0x0000FFFF)) << 16,
+                                                                  (batch<int32_t, A>(other.data) & batch<int32_t, A>(0x0000FFFF)) << 16,
+                                                                  Cmp);
+                        uint16_t mask_high = _mm256_cmp_epi32_mask((batch<int32_t, A>(self.data) & batch<int32_t, A>(0xFFFF0000)),
+                                                                   (batch<int32_t, A>(other.data) & batch<int32_t, A>(0xFFFF0000)),
+                                                                   Cmp);
+                        return static_cast<register_type>(morton(mask_low, mask_high));
+                    }
+                    else XSIMD_IF_CONSTEXPR(sizeof(T) == 4)
+                    {
+                        return (register_type)_mm256_cmp_epi32_mask(self, other, Cmp);
+                    }
+                    else XSIMD_IF_CONSTEXPR(sizeof(T) == 8)
+                    {
+                        return (register_type)_mm256_cmp_epi64_mask(self, other, Cmp);
+                    }
+                }
+                else
+                {
+                    XSIMD_IF_CONSTEXPR(sizeof(T) == 1)
+                    {
+                        uint64_t mask_low0 = _mm256_cmp_epu32_mask((batch<uint32_t, A>(self.data) & batch<uint32_t, A>(0x000000FF)), (batch<uint32_t, A>(other.data) & batch<uint32_t, A>(0x000000FF)), Cmp);
+                        uint64_t mask_low1 = _mm256_cmp_epu32_mask((batch<uint32_t, A>(self.data) & batch<uint32_t, A>(0x0000FF00)), (batch<uint32_t, A>(other.data) & batch<uint32_t, A>(0x0000FF00)), Cmp);
+                        uint64_t mask_high0 = _mm256_cmp_epu32_mask((batch<uint32_t, A>(self.data) & batch<uint32_t, A>(0x00FF0000)), (batch<uint32_t, A>(other.data) & batch<uint32_t, A>(0x00FF0000)), Cmp);
+                        uint64_t mask_high1 = _mm256_cmp_epu32_mask((batch<uint32_t, A>(self.data) & batch<uint32_t, A>(0xFF000000)), (batch<uint32_t, A>(other.data) & batch<uint32_t, A>(0xFF000000)), Cmp);
+                        uint64_t mask = 0;
+                        for (unsigned i = 0; i < 8; ++i)
+                        {
+                            mask |= (mask_low0 & (uint64_t(1) << i)) << (3 * i + 0);
+                            mask |= (mask_low1 & (uint64_t(1) << i)) << (3 * i + 1);
+                            mask |= (mask_high0 & (uint64_t(1) << i)) << (3 * i + 2);
+                            mask |= (mask_high1 & (uint64_t(1) << i)) << (3 * i + 3);
+                        }
+                        return (register_type)mask;
+                    }
+                    else XSIMD_IF_CONSTEXPR(sizeof(T) == 2)
+                    {
+                        uint16_t mask_low = _mm256_cmp_epu32_mask((batch<uint32_t, A>(self.data) & batch<uint32_t, A>(0x0000FFFF)), (batch<uint32_t, A>(other.data) & batch<uint32_t, A>(0x0000FFFF)), Cmp);
+                        uint16_t mask_high = _mm256_cmp_epu32_mask((batch<uint32_t, A>(self.data) & batch<uint32_t, A>(0xFFFF0000)), (batch<uint32_t, A>(other.data) & batch<uint32_t, A>(0xFFFF0000)), Cmp);
+                        return static_cast<register_type>(morton(mask_low, mask_high));
+                    }
+                    else XSIMD_IF_CONSTEXPR(sizeof(T) == 4)
+                    {
+                        return (register_type)_mm256_cmp_epu32_mask(self, other, Cmp);
+                    }
+                    else XSIMD_IF_CONSTEXPR(sizeof(T) == 8)
+                    {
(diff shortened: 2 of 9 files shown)


DiamonDinoia:
Hi @serge-sans-paille,

Are we sure that we only need avx512f for this? It seems to me that instructions like https://diamondinoia.com/simdref/#_mm256_cmp_epi32_mask requires avx512f + VL. 

Let me know where I am wrong. 

Cheers,
Marco

serge-sans-paille:
You're right, all of this requires avx512f+avx512vl. It turns out most build have both (see https://en.wikipedia.org/wiki/AVX-512#CPUs_with_AVX-512) but we currently don't have anything to model avx512vl, which should be the parent of avx512f_256.

DiamonDinoia:
I guess it might be called avx512vl_256 at that point as we will also have avx512vl_128. What do you think?

I agree, most CPU have 512+extensions.

serge-sans-paille:
> I guess it might be called avx512vl_256 at that point as we will also have avx512vl_128. What do you think?
> 
> I agree, most CPU have 512+extensions.

It looks like avx512vl does not have any 512bit instruction :-) but it still has avx512 in its name and it implies avx512f, so I agree with you.

DiamonDinoia:
Actually, my suggestion might be wrong: https://diamondinoia.com/simdref/#_mm_maskz_andnot_pd

There are instructions that require DQ + VL for example.

serge-sans-paille:
based on the graph above, looks like DQ => VL, so we should still be fine?

DiamonDinoia:
It seems like it, also basically all architectures listed in the chart have DQ,VL,BW together. So in any case is should be fine.

serge-sans-paille:
@DiamonDinoia I've move the minimal creation of avx512vl to #1350 , once merged I'll rebase this PR

DiamonDinoia:
Sure! Ping me when you need a review. It might be spotty when I'm on holiday

serge-sans-paille:
> Sure! Ping me when you need a review. It might be spotty when I'm on holiday

ping ;-)

include/xsimd/arch/xsimd_avx512vl_256.hpp
```diff
@@ -0,0 +1,721 @@
...
+        template <class A>
+        XSIMD_INLINE batch_bool<double, A> neq(batch<double, A> const& self, batch<double, A> const& other, requires_arch<avx512vl_256>) noexcept
+        {
+            return (typename batch_bool<double, A>::register_type)_mm256_cmp_pd_mask(self, other, _CMP_NEQ_OQ);
+        }
+
+        template <class A, class T, class = std::enable_if_t<std::is_integral<T>::value>>
+        XSIMD_INLINE batch_bool<T, A> neq(batch<T, A> const& self, batch<T, A> const& other, requires_arch<avx512vl_256>) noexcept
+        {
+            return ~(self == other);
```
DiamonDinoia: I think this flips junk into bits above batch_bool::size.


We might need to:
```cpp
constexpr auto active_mask = register_type(register_type(-1) >> (sizeof(register_type) * 4));
return ~(self == other) & active_mask;

```

include/xsimd/arch/xsimd_avx512vl_256.hpp
```diff
@@ -0,0 +1,721 @@
...
+        template <class A, class T, class = std::enable_if_t<std::is_integral<T>::value>>
+        XSIMD_INLINE batch_bool<T, A> eq(batch<T, A> const& self, batch<T, A> const& other, requires_arch<avx512vl_256>) noexcept
+        {
+            return detail::compare_int_avx512vl_256<A, T, _MM_CMPINT_EQ>(self, other);
+        }
+        template <class A, class T>
+        XSIMD_INLINE batch_bool<T, A> eq(batch_bool<T, A> const& self, batch_bool<T, A> const& other, requires_arch<avx512vl_256>) noexcept
+        {
+            using register_type = typename batch_bool<T, A>::register_type;
+            return register_type(~self.data ^ other.data);
```
DiamonDinoia, line 512: Same as neq here.

include/xsimd/arch/xsimd_avx512vl_256.hpp
```diff
@@ -0,0 +1,721 @@
...
+            using register_type = typename batch_bool<T, A>::register_type;
+            return register_type(self.data & ~other.data);
+        }
+
+        // bitwise_not
+        template <class A, class T>
+        XSIMD_INLINE batch_bool<T, A> bitwise_not(batch_bool<T, A> const& self, requires_arch<avx512vl_256>) noexcept
+        {
+            using register_type = typename batch_bool<T, A>::register_type;
+            return register_type(~self.data);
```
DiamonDinoia, line 687: here also we have a problem with the content of the high bits.

include/xsimd/arch/xsimd_avx512vl_256.hpp
```diff
@@ -0,0 +1,721 @@
...
+            return _mm256_mask_loadu_epi32(_mm256_setzero_si256(), imm_mask, mem);
+        }
+        template <class A, bool... Values, class Mode>
+        XSIMD_INLINE batch<uint32_t, A> load_masked(uint32_t const* mem, batch_bool_constant<uint32_t, A, Values...> mask, convert<uint32_t>, Mode, requires_arch<avx512vl_256>) noexcept
+        {
+            constexpr auto imm_mask = mask.mask();
+            return _mm256_mask_loadu_epi32(_mm256_setzero_si256(), imm_mask, mem);
+        }
+
+        // store masked
```
DiamonDinoia, line 204: We are ignoring the alignment here. We should use _mm256_mask_loadu_epi32 and so on for not 32-byte unaligned batches.

include/xsimd/arch/xsimd_avx512vl_256.hpp
```diff
@@ -0,0 +1,721 @@
...
+        // abs
+        template <class A>
+        XSIMD_INLINE batch<int64_t, A> abs(batch<int64_t, A> const& self, requires_arch<avx512vl_256>) noexcept
+        {
+            return _mm256_abs_epi64(self);
+        }
+
+        // load masked
+        template <class A, bool... Values, class Mode>
+        XSIMD_INLINE batch<int32_t, A> load_masked(int32_t const* mem, batch_bool_constant<int32_t, A, Values...> mask, convert<int32_t>, Mode, requires_arch<avx512vl_256>) noexcept
```
DiamonDinoia, line 192: here instead we are always hardcoding the unaligned which is safe but suboptimal/

include/xsimd/arch/xsimd_avx512vl_256.hpp
```diff
@@ -0,0 +1,721 @@
...
+        }
+
+        // store masked
+        template <class A, bool... Values, class Mode>
+        XSIMD_INLINE void store_masked(uint32_t* mem, batch<uint32_t, A> const& src, batch_bool_constant<uint32_t, A, Values...> mask, Mode, requires_arch<avx512vl_256>) noexcept
+        {
+            _mm256_mask_store_epi32(mem, mask.mask(), src);
+        }
+        template <class A, bool... Values, class Mode>
+        XSIMD_INLINE void store_masked(int32_t* mem, batch<int32_t, A> const& src, batch_bool_constant<uint32_t, A, Values...> mask, Mode, requires_arch<avx512vl_256>) noexcept
```
DiamonDinoia, line 211: the type of mem, batch<> and batch_bool_constant is not consistent here, should batch_bool_constant be int32_t throughout?

.github/workflows/linux.yml
```diff
@@ -94,6 +95,9 @@ jobs:
         if [[ '${{ matrix.sys.flags }}' == 'avx512' ]]; then
           CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=skylake-avx512"
         fi
+        if [[ '${{ matrix.sys.flags }}' == 'avx512vl_256' ]]; then
```
DiamonDinoia, line 100: The matrix passes `-DXSIMD_DEFAULT_ARCH=avx512f_256` but the build system only references that variable inside the *emulated* path (`test/CMakeLists.txt:50`). For a non-emulated build CMake emits:

```
CMake Warning: Manually-specified variables were not used by the project:
    XSIMD_DEFAULT_ARCH
```

include/xsimd/config/xsimd_cpu_features_x86.hpp
```diff
@@ -895,6 +895,8 @@ namespace xsimd
 
         inline bool avx512vl() const noexcept { return avx512_enabled() && leaf7().all_bits_set<x86_cpuid_leaf7::ebx::avx512vl>(); }
 
+        inline bool avx512vl_256() const noexcept { return avx_enabled() && leaf7().all_bits_set<x86_cpuid_leaf7::ebx::avx512vl>(); }
```
DiamonDinoia: we should check OPMASK XSAVE. 

`avx_enabled()` validates XCR0[1:2] (XMM+YMM). EVEX-encoded AVX-512 instructions use the opmask register state (XCR0[5]) **regardless of register width**. Without OS-managed OPMASK XSAVE, `k0–k7` are not preserved across context switches so we get a silent corruption.

include/xsimd/arch/xsimd_avx512vl_256.hpp
```diff
@@ -0,0 +1,721 @@
...
+        XSIMD_INLINE batch<T, A> from_bool(batch_bool<T, A> const& self, requires_arch<avx512vl_256>) noexcept
+        {
+            return select(self, batch<T, A>(1), batch<T, A>(0));
+        }
+
+        // from_mask
+        template <class T, class A>
+        XSIMD_INLINE batch_bool<T, A> from_mask(batch_bool<T, A> const&, uint64_t mask, requires_arch<avx512vl_256>) noexcept
+        {
+            return static_cast<typename batch_bool<T, A>::register_type>(mask);
```
DiamonDinoia: I think we should and this with 

```cpp
((uint64_t(1) << batch_bool<T,A>::size) - 1)
```

we should truncate the junk bits.

include/xsimd/arch/xsimd_avx512vl_256.hpp
```diff
@@ -0,0 +1,721 @@
...
+        }
+
+        // store
+        template <class T, class A>
+        XSIMD_INLINE void store(batch_bool<T, A> const& self, bool* mem, requires_arch<avx512vl_256>) noexcept
+        {
+            using register_type = typename batch_bool<T, A>::register_type;
+            constexpr auto size = batch_bool<T, A>::size;
+            for (std::size_t i = 0; i < size; ++i)
+                mem[i] = self.data & (register_type(1) << i);
```
DiamonDinoia: This way we store 0 or 2^i which is safe because c++ treats 0 as false, everything else as true. But a user doing something like arr[mem[i]] with arr of size 2 might do an out of bound access that is difficult to track down. I would add & 1 to make sure we store 0s and 1s.

include/xsimd/arch/xsimd_avx512vl_256.hpp
```diff
@@ -0,0 +1,721 @@
...
+        template <size_t count, class A, class T, class = std::enable_if_t<std::is_integral<T>::value>>
+        XSIMD_INLINE batch<T, A> rotl(batch<T, A> const& self, requires_arch<avx512vl_256>) noexcept
+        {
+            constexpr auto bits = std::numeric_limits<T>::digits + std::numeric_limits<T>::is_signed;
+            static_assert(count < bits, "Count must be less than the number of bits in T");
+            XSIMD_IF_CONSTEXPR(sizeof(T) == 4)
+            {
+                return _mm256_rol_epi32(self, count);
+            }
+            XSIMD_IF_CONSTEXPR(sizeof(T) == 8)
```
DiamonDinoia: here we do not use else but we do below. 
I do not like using else but it is better to do so as with -O0 the unused-size intrinsic is still instantiated and may emit

DiamonDinoia (changes requested):
Hi @serge-sans-paille,

Just some suggestions and maybe we could add to the tests:

```cpp
TEST_CASE_TEMPLATE_DEFINE("batch_bool mask hygiene", B, batch_bool_hygiene_id)
{
    using value_type = typename B::value_type;
    using batch_type = xsimd::batch<value_type, typename B::arch_type>;

    SUBCASE("any(a != a) is false")        
    {
        batch_type a(value_type(1));
        CHECK_FALSE(xsimd::any(a != a));
    }

    SUBCASE("any(~(a == a)) is false")        
    {
        batch_type a(value_type(1));
        CHECK_FALSE(xsimd::any(~(a == a)));
    }

    SUBCASE("eq(false_mask, false_mask) is all-true")  
    {
        auto m0 = (batch_type(value_type(1)) != batch_type(value_type(1)));
        CHECK_UNARY(xsimd::all(m0 == m0));
    }

    SUBCASE("from_mask ignores bits above lane count")  
    {
        constexpr std::size_t N = B::size;
        uint64_t valid_all_true = (N == 64) ? ~uint64_t(0)
                                            : ((uint64_t(1) << N) - 1);
        uint64_t junk            = valid_all_true | (uint64_t(1) << 63);
        B clean = B::from_mask(valid_all_true);
        B dirty = B::from_mask(junk);
        CHECK_EQ(clean.mask(), dirty.mask());
        CHECK_UNARY(xsimd::all(clean == dirty));
    }

    SUBCASE("batch_bool stored to bool[] is canonical 0/1") 
    {
        batch_type a(value_type(1)), b(value_type(2));
        auto m = (a == b);                    // all false
        alignas(64) bool buf[B::size + 1] = {true, true};  // sentinel
        xsimd::store_aligned(buf, m);
        for (std::size_t i = 0; i < B::size; ++i)
        {
            // bit-level check: must be exactly 0, not just falsy
            CHECK_EQ(*reinterpret_cast<uint8_t const*>(&buf[i]), uint8_t(0));
        }
    }
}
TEST_CASE_TEMPLATE_APPLY(batch_bool_hygiene_id, batch_bool_types);
```cpp
TEST_CASE_TEMPLATE_DEFINE("store_masked respects Mode", B, store_masked_mode_id)
{
    using T = typename B::value_type;
    using A = typename B::arch_type;
    constexpr std::size_t N = B::size;

    // Unaligned-mode + unaligned pointer: must not fault.       
    alignas(64) T big[2 * N + 1] = {};
    T* unaligned_p = big + 1;                 // sizeof(T)-aligned only

    struct AllTrue { static constexpr bool get(std::size_t, std::size_t) { return true; } };
    auto cst = xsimd::make_batch_bool_constant<T, AllTrue, A>();

    B v(T(7));
    xsimd::kernel::store_masked(unaligned_p, v, cst,
                                xsimd::unaligned_mode{},
                                xsimd::kernel::requires_arch<A>{});
    for (std::size_t i = 0; i < N; ++i)
        CHECK_EQ(unaligned_p[i], T(7));

    // Overload resolution: store_masked with same-typed mask must compile.
    // If C3 regresses, this becomes a compile error.                
    auto signed_cst = xsimd::make_batch_bool_constant<T, AllTrue, A>();
    alignas(64) T aligned_buf[N] = {};
    xsimd::kernel::store_masked(aligned_buf, v, signed_cst,
                                xsimd::aligned_mode{},
                                xsimd::kernel::requires_arch<A>{});
    for (std::size_t i = 0; i < N; ++i)
        CHECK_EQ(aligned_buf[i], T(7));
}
TEST_CASE_TEMPLATE_APPLY(store_masked_mode_id, batch_types_for_masked_store);
```

include/xsimd/arch/xsimd_avx512vl_256.hpp
serge-sans-paille: yes for the others, but for this one everything is done at batch level so we're good (provided operator== is correct)

serge-sans-paille: Actually as we always ignore the upper bits, I think we're good

include/xsimd/arch/xsimd_avx512vl_256.hpp
serge-sans-paille: what's the issue with upper bits being garbage?

serge-sans-paille: I've changed ` uint64_t mask(batch_bool<T, A> const& self, requires_arch<avx512vl_256>) noexcept` to apply the mask, because that's where the garbage bits may leak.

include/xsimd/arch/xsimd_avx512vl_256.hpp
serge-sans-paille: I've kept it as is, optimal is not a primary goal here.

.github/workflows/linux.yml
serge-sans-paille: That's actually a big problem, esp. as avx_128 and avx2_128  testing suffer from the same issue. Big thanks!

include/xsimd/arch/xsimd_avx512vl_256.hpp
DiamonDinoia: I was thinking about if there are "horizontal" operations like and, or and so on. I'm not sure there is a vl variant that ignores these bits when using them. Otherwise when using the full avx512 variant we have to 0 those bits manually.

serge-sans-paille:
All green!

include/xsimd/arch/xsimd_avx512vl_256.hpp
```diff
@@ -0,0 +1,723 @@
...
+
+        template <class A, bool... Values, class Mode>
+        XSIMD_INLINE void store_masked(int64_t* mem, batch<int64_t, A> const& src, batch_bool_constant<uint64_t, A, Values...> mask, Mode, requires_arch<avx512vl_256>) noexcept
+        {
+            _mm256_mask_storeu_epi64(mem, mask.mask(), src);
+        }
+        template <class A, bool... Values, class Mode>
+        XSIMD_INLINE void store_masked(float* mem, batch<float, A> const& src, batch_bool_constant<uint32_t, A, Values...> mask, Mode, requires_arch<avx512vl_256>) noexcept
+        {
+            _mm256_mask_store_ps(mem, mask.mask(), src);
```
DiamonDinoia: this requires alignment to 32 bytes so we have to specify the Mode

include/xsimd/arch/xsimd_avx512vl_256.hpp
```diff
@@ -0,0 +1,723 @@
...
+        }
+
+        template <class A, bool... Values, class Mode>
+        XSIMD_INLINE void store_masked(double* mem, batch<double, A> const& src, batch_bool_constant<uint64_t, A, Values...> mask, Mode, requires_arch<avx512vl_256>) noexcept
+        {
+            _mm256_mask_store_pd(mem, mask.mask(), src);
```
DiamonDinoia: same here this requires alignment.

DiamonDinoia (commented):
Hi,

Can you check that the following code produces the expected output? Because on my machine I get that a == a is true and a != is also true in some cases for example.

Also are we handling the store/load alignment correctly? I seem to notice (unless overload resolution dispatches to some other types) that we always use unaligned for integer types and always aligned for float/double?

```
// clang++ -march=x64-64-v4 -I/home/marco/repos/xsimd/include -DXSIMD_DEFAULT_ARCH=xsimd::avx512vl_256 /tmp/xsimd_stray_bits.cpp -o /tmp/xsimd_stray_bits
#include <cstdio>
#include <cstdint>
#include "xsimd/xsimd.hpp"


#include <iostream>
#include <typeinfo>
#include <cxxabi.h>
#include <memory>

// Helper function to demangle typeid names
std::string demangle(const char* mangled) {
    int status = 0;
    std::unique_ptr<char, decltype(&free)> result(
        abi::__cxa_demangle(mangled, nullptr, nullptr, &status),
        &free
    );
    return (status == 0) ? result.get() : mangled;
}


static int fail_count = 0;

#define EXPECT(label, expr, expected)                                          \
    do {                                                                       \
        const bool got = (expr);                                               \
        const bool ok  = (got == (expected));                                  \
        std::printf("    %-58s expected %-5s got %-5s  %s\n",                  \
                    (label),                                                   \
                    (expected) ? "true" : "false",                             \
                    got ? "true" : "false",                                    \
                    ok ? "PASS" : "FAIL");                                     \
        if (!ok) ++fail_count;                                                 \
    } while (0)

template <class T>
int check() {
    using A = xsimd::avx512vl_256;
    using B = xsimd::batch<T, A>;
    std::cout << "\n=== " << demangle(typeid(B).name())
              << "  (arch=" << demangle(typeid(A).name())
              << ", lanes=" << B::size << ") ===\n";

    B a(T(1));
    std::cout << "  a = " << a << "\n";

    // a equals itself
    EXPECT("a should equal itself                                 [all(a == a)]",
           xsimd::all(a == a), true);
    EXPECT("a should not differ from itself                       [any(a != a)]",
           xsimd::any(a != a), false);

    // combine two logically empty masks
    auto m1 = (a != a);
    auto m2 = (a != a);
    EXPECT("AND of two empty masks should be empty                [any(m1 & m2)]",
           xsimd::any(m1 & m2), false);
    EXPECT("OR  of two empty masks should be empty                [any(m1 | m2)]",
           xsimd::any(m1 | m2), false);
    EXPECT("XOR of identical empty masks should be empty          [any(m1 ^ m2)]",
           xsimd::any(m1 ^ m2), false);

    // complement of an empty mask is a full mask
    auto m_not = ~m1;
    EXPECT("complement of an empty mask should be full            [all(~m1)]",
           xsimd::all(m_not), true);
    EXPECT("complement of an empty mask should be nonempty        [any(~m1)]",
           xsimd::any(m_not), true);

    // a bool batch equals itself
    EXPECT("a bool mask should equal itself                       [all(m_not == m_not)]",
           xsimd::all(m_not == m_not), true);

    return 0;
}

int main() {
    int fails = 0;
    fails += check<int8_t>();
    fails += check<int16_t>();
    fails += check<int32_t>();
    fails += check<int64_t>();
    fails += check<float>();
    fails += check<double>();
    std::printf("%d failures\n", fails);
    return fails != 0 ? 1 : 0;
}
```

include/xsimd/arch/xsimd_avx512vl_256.hpp
serge-sans-paille: I went the easy way and forced unaligned store here and below.

serge-sans-paille:
bug found, the all / any implementation was incorrect, fixed and test case added (your example was strange, the `fails` accumulator was always zero)

serge-sans-paille:
And all green again :-)

include/xsimd/arch/xsimd_avx512vl_256.hpp
```diff
@@ -0,0 +1,725 @@
...
+        }
+
+        template <class A, bool... Values, class Mode>
+        XSIMD_INLINE void store_masked(uint64_t* mem, batch<uint64_t, A> const& src, batch_bool_constant<uint64_t, A, Values...> mask, Mode, requires_arch<avx512vl_256>) noexcept
+        {
+            _mm256_mask_storeu_epi64(mem, mask.mask(), src);
+        }
+
+        template <class A, bool... Values, class Mode>
+        XSIMD_INLINE void store_masked(int64_t* mem, batch<int64_t, A> const& src, batch_bool_constant<uint64_t, A, Values...> mask, Mode, requires_arch<avx512vl_256>) noexcept
```
DiamonDinoia, line 223: should the type of the batch_bool_constant be int64_t?

include/xsimd/config/xsimd_cpu_features_x86.hpp
```diff
@@ -895,6 +895,8 @@ namespace xsimd
 
         inline bool avx512vl() const noexcept { return avx512_enabled() && leaf7().all_bits_set<x86_cpuid_leaf7::ebx::avx512vl>(); }
 
+        inline bool avx512vl_256() const noexcept { return avx_enabled() && osxsave() && leaf7().all_bits_set<x86_cpuid_leaf7::ebx::avx512vl>(); }
```
DiamonDinoia: are we checking XCR0[5] (OPMASK state) ?

include/xsimd/arch/xsimd_avx512vl_256.hpp
```diff
@@ -0,0 +1,725 @@
...
+        XSIMD_INLINE batch_bool<double, A> isnan(batch<double, A> const& self, requires_arch<avx512vl_256>) noexcept
+        {
+            return (typename batch_bool<double, A>::register_type)_mm256_cmp_pd_mask(self, self, _CMP_UNORD_Q);
+        }
+
+        // rotl
+        template <class A, class T, class = std::enable_if_t<std::is_integral<T>::value>>
+        XSIMD_INLINE batch<T, A> rotl(batch<T, A> const& self, batch<T, A> const& other, requires_arch<avx512vl_256>) noexcept
+        {
+            XSIMD_IF_CONSTEXPR(sizeof(T) == 4)
```
DiamonDinoia, line 394: here we do not have else XSIMD_IF_CONSTEXPR while we do on lines 418 and 422

DiamonDinoia (approved):
Looks good! The code works locally now without issues. I just have some suggestions which are not crucial. One is more a question to make sure things are okay.

include/xsimd/arch/xsimd_avx512vl_256.hpp
serge-sans-paille: I don't think so, the mask are always unsigned

include/xsimd/config/xsimd_cpu_features_x86.hpp
serge-sans-paille: we were not. should be good now, thanks for spotting this.

DiamonDinoia:
Great work! Thanks!

DiamonDinoia (approved):

serge-sans-paille:
Thank *you*, the review was great and very helpful

PR #1348, opened 2026-05-13
Add CMakePresets for target micro arch

AntoinePrv:
I've taken the direction of explicit flags such as `-mavx -mno-avx2`.
This is IMHO less error prone and more accurate that using architecture name such as `haswell`.
The main difference is that this does not add other feature flags or change the `-mtune` model.
For a test setting accuracy is more important IMHO.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -10,46 +10,50 @@ defaults:
 jobs:
   build:
     runs-on: ubuntu-latest
-    name: '${{ matrix.sys.compiler }} ${{ matrix.sys.version }} - ${{ matrix.sys.flags }}'
+    name: '${{ matrix.sys.compiler }} ${{ matrix.sys.version }} - ${{ matrix.sys.preset }} - ${{ matrix.sys.flags }}'
     strategy:
       matrix:
         sys:
-          - { compiler: 'gcc',   version: '12', flags: 'force_no_instr_set' }
-          - { compiler: 'gcc',   version: '13', flags: 'enable_xtl_complex' }
-          - { compiler: 'gcc',   version: '14', flags: 'avx' }
-          - { compiler: 'gcc',   version: '14', flags: 'avx2' }
-          - { compiler: 'gcc',   version: '13', flags: 'avx512' }
-          - { compiler: 'gcc',   version: '10', flags: 'avx512' }
-          - { compiler: 'gcc',   version: '12', flags: 'i386' }
-          - { compiler: 'gcc',   version: '13', flags: 'avx512pf' }
-          - { compiler: 'gcc',   version: '13', flags: 'avx512vbmi' }
-          - { compiler: 'gcc',   version: '14', flags: 'avx512vbmi2' }
-          - { compiler: 'gcc',   version: '13', flags: 'avx512vnni' }
-          - { compiler: 'clang', version: '16', flags: 'force_no_instr_set' }
-          - { compiler: 'clang', version: '16', flags: 'enable_xtl_complex' }
-          - { compiler: 'clang', version: '17', flags: 'avx' }
-          - { compiler: 'clang', version: '17', flags: 'sse3' }
-          - { compiler: 'clang', version: '18', flags: 'avx512' }
-          - { compiler: 'clang', version: '18', flags: 'avx_128' }
-          - { compiler: 'clang', version: '18', flags: 'avx2_128' }
-          - { compiler: 'clang', version: '18', flags: 'avx512vl_128' }
-          - { compiler: 'clang', version: '18', flags: 'avx512vl_256' }
+          - { compiler: 'gcc',   version: '12', flags: 'force_no_instr_set', preset: 'native' }
+          - { compiler: 'gcc',   version: '13', flags: 'enable_xtl_complex', preset: 'native' }
+          - { compiler: 'gcc',   version: '14', flags: '',                   preset: 'avx' }
+          - { compiler: 'gcc',   version: '14', flags: '',                   preset: 'avx2' }
+          - { compiler: 'gcc',   version: '13', flags: '',                   preset: 'avx512f' }
+          - { compiler: 'gcc',   version: '10', flags: '',                   preset: 'avx512f' }
+          - { compiler: 'gcc',   version: '12', flags: 'i386',               preset: 'native' }
+          - { compiler: 'gcc',   version: '13', flags: '',                   preset: 'avx512pf' }
+          - { compiler: 'gcc',   version: '13', flags: '',                   preset: 'avx512vbmi' }
+          - { compiler: 'gcc',   version: '14', flags: '',                   preset: 'avx512vbmi2' }
+          - { compiler: 'gcc',   version: '13', flags: '',                   preset: 'avx512vnni_avx512bw' }
+          - { compiler: 'clang', version: '16', flags: 'force_no_instr_set', preset: 'native' }
+          - { compiler: 'clang', version: '16', flags: 'enable_xtl_complex', preset: 'native' }
+          - { compiler: 'clang', version: '17', flags: '',                   preset: 'avx' }
+          - { compiler: 'clang', version: '17', flags: '',                   preset: 'sse3' }
+          - { compiler: 'clang', version: '18', flags: '',                   preset: 'avx512f' }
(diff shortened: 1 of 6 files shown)


.github/workflows/linux.yml
```diff
@@ -10,28 +10,28 @@ defaults:
...
+          - { compiler: 'gcc',   version: '13', flags: '',                   preset: 'avx512pf' }
+          - { compiler: 'gcc',   version: '13', flags: '',                   preset: 'avx512vbmi' }
+          - { compiler: 'gcc',   version: '14', flags: '',                   preset: 'avx512vbmi2' }
+          - { compiler: 'gcc',   version: '13', flags: '',                   preset: 'avx512vnni_avx512bw' }
+          - { compiler: 'clang', version: '16', flags: 'force_no_instr_set', preset: 'native' }
+          - { compiler: 'clang', version: '16', flags: 'enable_xtl_complex', preset: 'native' }
+          - { compiler: 'clang', version: '17', flags: '',                   preset: 'avx' }
+          - { compiler: 'clang', version: '17', flags: '',                   preset: 'sse3' }
+          - { compiler: 'clang', version: '18', flags: '',                   preset: 'avx512' }
+          - { compiler: 'clang', version: '18', flags: '',                   preset: 'avx128' }
```
serge-sans-paille: missing underscore here and below?

serge-sans-paille:
I really like your approach and will eagerly merge it once it validates \o/

AntoinePrv:
I've only kept the micro architecture target in `CMakePresets.txt` because combining with (debug/release) / (xtl on/off)... results in a combinatorial explosion of presets for which there is currently no support.
Another shortcoming is that we cannot dispatch here based on compiler for MSVC flags. We can do it based on OS but it is not quite the same.

I have ongoing work to actually do the same as these presets at the CMake level, with a function that can be made available to users to help in the tooling for dynamic dispatch (our current solution in Arrow is very verbose).
In this case, we'd need to also define a safe `-march` baseline. The reason is the code in these translation units might also include non SIMD code (this is sometimes the case in Arrow). In this case, with very advanced instruction sets, we're leaving perf on the table by having a `x86-64` baseline. But what should be a reasonable baseline for dynamic dispatching to for example `avx2`?
- `haswell` (first `avx2`) also has `fma3` and `bmi2`
- `-march=haswell -mno-fma3 -mno-bmi2` if that is a thing?
- Or go further back? `sandybridge` (first `avx`)? `nehalem` (first `sse4.2`)

AntoinePrv:
@serge-sans-paille this is in a ready state, but I am not fully happy with it.

Getting into AVX512, and AVX512-256, the combinatorial explosion of possibilities start to show again.
Inheritance of flags from other settings is also not possible.

This reinforce my belief that I should keep on with the work to do it in CMake (that could also be installed for our users to improve our dynamic dispatch tooling), and also homogenized with the test `TARGET_ARCH` var.

This PR is not completely worthless though. For example we now have the possibility to really test with `avx512f`, which was not the case before because no Intel arch is limited to the `f` feature only.

What do you think? Should we give this some mileage before I get the time to work on a CMake solution?

.github/workflows/linux.yml
```diff
@@ -77,82 +84,47 @@ jobs:
...
         fi
-        if [[ '${{ matrix.sys.flags }}' == 'avx' ]]; then
-          CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=sandybridge"
-        fi
-        if [[ '${{ matrix.sys.flags }}' == 'avx_128' ]]; then
-          CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=sandybridge"
-          CXXFLAGS="$CXX_FLAGS -DXSIMD_DEFAULT_ARCH=avx_128"
-        fi
-        if [[ '${{ matrix.sys.flags }}' == 'avx2' ]]; then
-          CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=haswell"
```
serge-sans-paille, line 89: what a nice cleanup. The current merge conflict is probably related to the latest addition of avx512vl support, sorry about that

.github/workflows/linux.yml
```diff
@@ -77,82 +84,47 @@ jobs:
...
-        fi
-        if [[ '${{ matrix.sys.flags }}' == 'avx512vbmi2' ]]; then
-          CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=icelake-server"
-        fi
-        if [[ '${{ matrix.sys.flags }}' == 'avx512vnni' ]]; then
-          CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=knm"
-        fi
         if [[ '${{ matrix.sys.flags }}' == 'i386' ]]; then
-          CXX_FLAGS="$CXX_FLAGS -m32"
+          export CXXFLAGS="$CXXFLAGS -m32"
```
serge-sans-paille, line 86: !!!

.github/workflows/linux.yml
```diff
@@ -77,82 +84,47 @@ jobs:
...
+              -D CMAKE_C_COMPILER=$CC \
+              -D CMAKE_CXX_COMPILER=$CXX \
+              -D CMAKE_CXX_FLAGS='$CXX_FLAGS' \
+              -D TARGET_ARCH="x86-64" \
+              "${CMAKE_EXTRA_ARGS}" \
               -G Ninja
+
     - name: Build
-      run: cmake --build _build
+      run: cmake --build _build --parallel
```
serge-sans-paille, line 106: we probably have the same issue elsewhere, I'll handle hat

CMakePresets.json
```diff
@@ -0,0 +1,208 @@
...
+        {
+            "name": "avx2_128",
+            "cacheVariables": {
+                "CMAKE_CXX_FLAGS": "$env{CXXFLAGS} -march=x86-64-v2 -mno-sse4a -mavx2 -mno-avx512f -DXSIMD_DEFAULT_ARCH=avx2_128"
+            }
+        },
+        {
+            "name": "avx512f",
+            "cacheVariables": {
+                "CMAKE_CXX_FLAGS": "$env{CXXFLAGS} -march=x86-64-v2 -mno-sse4a -mavx -mavx2 -mavx512f -mno-avx512cd -mno-avx512dq -mno-avx512bw -mno-avx512er -mno-avx512pf -mno-avx512ifma -mno-avx512vbmi -mno-avx512vbmi2 -mno-avx512vnni"
```
serge-sans-paille: I like this strict checking. And it looks like this did not uncover too many errors, yay

CMakePresets.json
```diff
@@ -0,0 +1,208 @@
...
+        {
+            "name": "avx_128",
+            "cacheVariables": {
+                "CMAKE_CXX_FLAGS": "$env{CXXFLAGS} -march=x86-64-v2 -mno-sse4a -mavx -mno-avx2 -DXSIMD_DEFAULT_ARCH=avx_128"
+            }
+        },
+        {
+            "name": "avx2",
+            "cacheVariables": {
+                "CMAKE_CXX_FLAGS": "$env{CXXFLAGS} -march=x86-64-v2 -mno-sse4a -mavx2 -mno-avx512f"
```
serge-sans-paille: we sometime have fallback from avx2 instructions to sse instructions. How can this work??

serge-sans-paille: I do understand the need to prune higher instruction sets, but not the need to prune lower ones, please explain.

serge-sans-paille (changes requested):
All looks good, *except* the question on pruning lower architectures which raises a big unknown to me.

CMakePresets.json
AntoinePrv: Do you mean `-mno-sse4a` ? This can be removed, I added when trying to debug some `-march=native` that was added by the absence of `TARGET_ARCH`.

Though it is not a problem here: `sse4a` is an AMD extension that was never implemented on Intel (and that is why it was failing in SDE).

.github/workflows/linux.yml
AntoinePrv: Yes, there is a weird mismatch in master. Both `CXX_FLAGS` and `CXXFLAGS` where set but only `CXX_FLAGS` was explicitly passed to CMake. `CXXFLAGS` is picked up automatically but it was not exported.

PR #1351, opened 2026-05-19
Resolve xtl in user cmake

### LINKED ISSUE #1338: inconsistency in configuration of XTL complex number support
xsimd 14.2.0 introduced a change in the way complex number (xtl) support is configured. The problem arises when xsimd in general is configured and installed with complex number support (building the xsimd installation with `-DENABLE_XTL_COMPLEX=ON`), while a specific application uses xsimd without using the complex number support (so the application is not using `-DENABLE_XTL_COMPLEX=On`).

The problem arises due to PR #1279 , which "hard-codes" the older `cxx_std_14` with or without `ENABLE_XTL_COMPLEX`. Previously the CXX standard was to set to either the newer or the older standard depending on whether  `ENABLE_XTL_COMPLEX` is supported or not (xtl requires the newer standard to provide complex number support ).

An example is running the xsimd tests standalone against the install xsimd (i.e. configuring with config files from /usr/share/cmake/xsimd/).
If `ENABLE_XTL_COMPLEX` is not specified,  the tests fail  since  /usr/share/cmake/xsimd/xsimdTargets.cmake sets both `XSIMD_ENABLE_XTL_COMPLEX=1` (and links xtl) but also sets `cxx_std_14`, evidently without first finding xtl (in  xsimdConfig.cmake) which would set `cxx_std_17`:
```
[  3%] Building CXX object CMakeFiles/test_xsimd.dir/test_api.cpp.o
/usr/bin/c++ -DXSIMD_ENABLE_XTL_COMPLEX=1  -Wunused-parameter -Wextra -Wreorder -std=c++14 -g -fPIC -march=native -O3 -DNDEBUG -MD -MT CMakeFiles/test_xsimd.dir/test_api.cpp.o -MF CMakeFiles/test_xsimd.dir/test_api.cpp.o.d -o CMakeFiles/test_xsimd.dir/test_api.cpp.o -c /projects/mathlibs/build/xsimd/test/test_api.cpp
In file included from /usr/include/xtl/xcomplex.hpp:32,
                 from /usr/include/xsimd/arch/xsimd_scalar.hpp:26,
                 from /usr/include/xsimd/xsimd.hpp:18,
                 from /projects/mathlibs/build/xsimd/test/test_api.cpp:12:
/usr/include/xtl/xtype_traits.hpp:306:40: error: ‘conjunction’ is not a member of ‘std’
  306 |     constexpr bool xtl_requires = std::conjunction<C...>::value;
      |                                        ^~~~~~~~~~~
/usr/include/xtl/xtype_traits.hpp:306:40: note: ‘std::conjunction’ is only available from C++17 onwards
```

If the application is instead configured with `-DENABLE_XTL_COMPLEX=On`, then /usr/share/cmake/xsimd/xsimdConfig.cmake finds xtl, which adds `cxx_std_17`, and the application build succeeds (the tests pass).

My testing is in the context of building the debian package of xsimd.  The intention is for complex number support to be available (supported by /usr/share/cmake/xsimd/xsimdConfig.cmake ) but only used by applications when explicitly requested (when an application configuration sets `-DENABLE_XTL_COMPLEX=On`).    Is this a correct interpretation of xsimd's complex number support, that it should be configurable by applications like this?

Should PR #1279 be reverted (and updated for `cxx_std_14` vs `cxx_std_17`)?  This would have the impact of aligning the xsimd installation with the requirements of installed complex number support. That is, if xsimd is configured with complex number support but an application is not using it, then the application will get the newer CXX standard (`cxx_std_17`) regardless.

Alternatively, should more logic control be added to xsimdConfig.cmake.in (or to the installed xsimdTargets.cmake) so that the CXX standard set in the installed cmake config files is made conditional on whether an application sets `-DENABLE_XTL_COMPLEX=On` or not?  This would mean conditional setting of `set_target_properties` in `xsimdTargets.cmake`.

AntoinePrv:
Instead of having the packager (Debian, Conda recipe) decide on xtl support, let the final user decide on wheter to use it.

The current behaviour is maintained and overridable until xsim 15.
In xsimd 15, the configure time option has no effect for installed targets.

Close #1338

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,9 @@ cmake_minimum_required(VERSION 3.13)
 
 project(xsimd)
 
+OPTION(ENABLE_XTL_COMPLEX "enables support for xcomplex defined in xtl" OFF)
+OPTION(BUILD_TESTS "xsimd test suite" OFF)
+
 # Versioning
 # ==========
 
@@ -37,14 +40,18 @@ target_include_directories(xsimd INTERFACE
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
     $<INSTALL_INTERFACE:include>)
 
-OPTION(ENABLE_XTL_COMPLEX "enables support for xcomplex defined in xtl" OFF)
-OPTION(BUILD_TESTS "xsimd test suite" OFF)
-
 target_compile_features(xsimd INTERFACE cxx_std_14)
+
+# Only add xtl build option to the build tree, that is, if xsimd being locally
+# developed or is vendored.
+# Otherwise (if an install is performed), this will be handled in the user
+# cmake script (xsimdConfig.cmake).
 if(ENABLE_XTL_COMPLEX)
     find_package(xtl 0.8.0 REQUIRED)
-    target_compile_definitions(xsimd INTERFACE XSIMD_ENABLE_XTL_COMPLEX=1)
-    target_link_libraries(xsimd INTERFACE xtl)
+    target_link_libraries(xsimd INTERFACE $<BUILD_INTERFACE:xtl>)
+    target_compile_definitions(xsimd INTERFACE
+        $<BUILD_INTERFACE:XSIMD_ENABLE_XTL_COMPLEX=1>
+    )
 endif()
 
 if(BUILD_TESTS)
--- a/xsimdConfig.cmake.in
+++ b/xsimdConfig.cmake.in
@@ -18,12 +18,37 @@
 @PACKAGE_INIT@
 
 if(NOT TARGET @PROJECT_NAME@)
-    set(@PROJECT_NAME@_ENABLE_XTL_COMPLEX @ENABLE_XTL_COMPLEX@)
-    if(@PROJECT_NAME@_ENABLE_XTL_COMPLEX)
-        include(CMakeFindDependencyMacro)
-        find_dependency(xtl REQUIRED)
+    include("${CMAKE_CURRENT_LIST_DIR}/@PROJECT_NAME@Targets.cmake")
+    get_target_property(
+        @PROJECT_NAME@_INCLUDE_DIRS
+        @PROJECT_NAME@ INTERFACE_INCLUDE_DIRECTORIES
+    )
+
+    # xsimd <= 14 behaviour.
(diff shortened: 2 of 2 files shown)


AntoinePrv:
@drew-parsons does that work for you?

JohanMabille (approved):

serge-sans-paille (approved):
LGTM too, thanks

drew-parsons:
I'm not certain how to judge this PR.  It explicitly preserves the current behaviour of xsimd 14, i.e. reproduces the "inconsistent" behaviour that I reported, adding `-DXSIMD_ENABLE_XTL_COMPLEX=1` with or without ENABLE_XTL_COMPLEX. I guess it will become more clear once xsimd 15 is released whether this patch works the way that I was thinking.

edit: wait, no, I think it's working fine for me.  I didn't have the tests in my demo test organised as clearly as it could be.  The "2 tests" it reported were for type float vs complex, not test_load_store vs test_load_store_xtl.  doctest didn't report the SUBCASES clearly.  Placing test_load_store_xtl in a separate TEST_CASE_TEMPLATE rather than a SUBCASE shows the patch is working.  Compile flag `-DXSIMD_ENABLE_XTL_COMPLEX=1` gets added by default in my configuration, but can be deactivated by configuring `cmake -DXSIMD_ENABLE_XTL_COMPLEX=Off ..`. I'll update my bug report with the updated demo test.

So, looks like this patch is working fine with xsimd 14. 

If I understood right, the ENABLE_XTL_COMPLEX handling for tests at https://github.com/xtensor-stack/xsimd/blob/510b4f189f66576784899b4c3f4f11781a70d4ed/test/CMakeLists.txt#L62 could be removed or refactored in the future with xsimd 15.

drew-parsons:
Incidentally, I added an extra test run to the debian CI tests, "test xsimd-test-float" (and test "xsimd-test-clang-float"), which runs the tests with this patch with an explicit `-DXSIMD_ENABLE_XTL_COMPLEX=Off`. You can see in the test logs at https://ci.debian.net/packages/x/xsimd/testing/amd64/71383346/#S8
that it causes the `-DXSIMD_ENABLE_XTL_COMPLEX=1` to be dropped from compile flags, as intended from this PR.

PR #1352, opened 2026-05-19
Tentative support for avx512vl extensions to 128 bit registers

serge-sans-paille:
Very similar to the implementation for avx512vl_256, with a few less instruction supported though.

As a side effect, fix an OOB acces in tobitset for avx512_256


JohanMabille (approved):

DiamonDinoia (approved):
Looks good! It pass all the tests on my machine even some corner cases I made up.

PR #1354, opened 2026-05-21
Add efficient_bmi2 cpu feature

AntoinePrv:


include/xsimd/config/xsimd_cpu_features_x86.hpp
```diff
@@ -841,6 +860,45 @@ namespace xsimd
...
+         * @see https://en.wikipedia.org/wiki/CPUID#EAX=1:_Processor_Info_and_Feature_Bits
+         */
+        inline detail::x86_reg32_t cpu_model() const noexcept
+        {
+            using eax = x86_cpuid_leaf1::eax;
+            auto model = leaf1().get_range<eax::model_start, eax::model_end>();
+            auto ext_model = leaf1().get_range<eax::ext_model_start, eax::ext_model_end>();
+            auto family_id = leaf1().get_range<eax::family_id_start, eax::family_id_end>();
+            return (family_id == 15 || family_id == 6) ? ((ext_model << 4) + model) : model;
+        }
```
serge-sans-paille, line 900: it would be great to add some testing for those.
I wonder if we could compare to the output of `__builtin_cpu_is` and `__builtin_cpu_supports` as a sanity check?

serge-sans-paille: (can be in a follow-up)

serge-sans-paille:
LGTM

serge-sans-paille:
btw: thanks for quoting the wikipedia page, it helped a lot for the review

AntoinePrv:
I think a good approach would be an optional executable that spit it all out as text.
Then we can easily compare with cpuinfo etc. Would also be useful for our user if they ever need to debug if we detect feature xxx.

PR #1355, opened 2026-05-21
Fix avx512vp cpuid

AntoinePrv:
This will properly apply the chain
`leaf1:osxsave`  -> `xcr0:[xy]mm`
With proper default if `osxsave` is not set.

Previously only this checks for saving `zmm` registers (`avx512_enabled`) so this would get deactivted every time 512 bits registers get deactivated (which would be one motivation for using smaller registers).

@serge-sans-paille I think you might have gotten the first implementation wrong, no?

### DIFF
--- a/include/xsimd/config/xsimd_cpu_features_x86.hpp
+++ b/include/xsimd/config/xsimd_cpu_features_x86.hpp
@@ -857,7 +857,12 @@ namespace xsimd
 
         inline bool avx() const noexcept { return avx_enabled() && leaf1().all_bits_set<x86_cpuid_leaf1::ecx::avx>(); }
 
-        inline bool avx_128() const noexcept { return sse_enabled() && leaf1().all_bits_set<x86_cpuid_leaf1::ecx::avx>(); }
+        inline bool avx_128() const noexcept
+        {
+            // Avx 128 bit instructions use the same xmm registers from SSE so checking if those
+            // are enabled is sufficient.
+            return sse_enabled() && leaf1().all_bits_set<x86_cpuid_leaf1::ecx::avx>();
+        }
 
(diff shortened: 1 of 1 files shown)


include/xsimd/config/xsimd_cpu_features_x86.hpp
```diff
@@ -857,7 +857,12 @@ namespace xsimd
 
         inline bool avx() const noexcept { return avx_enabled() && leaf1().all_bits_set<x86_cpuid_leaf1::ecx::avx>(); }
 
-        inline bool avx_128() const noexcept { return sse_enabled() && leaf1().all_bits_set<x86_cpuid_leaf1::ecx::avx>(); }
+        inline bool avx_128() const noexcept
+        {
+            // Avx 128 bit instructions use the same xmm registers from SSE so checking if those
+            // are enabled is sufficient.
+            return sse_enabled() && leaf1().all_bits_set<x86_cpuid_leaf1::ecx::avx>();
+        }
```
AntoinePrv, line 865: Just adding a comment

include/xsimd/config/xsimd_cpu_features_x86.hpp
```diff
@@ -869,7 +874,12 @@ namespace xsimd
 
         inline bool avx2() const noexcept { return avx_enabled() && leaf7().all_bits_set<x86_cpuid_leaf7::ebx::avx2>(); }
 
-        inline bool avx2_128() const noexcept { return sse_enabled() && leaf7().all_bits_set<x86_cpuid_leaf7::ebx::avx2>(); }
+        inline bool avx2_128() const noexcept
+        {
+            // Avx 128 bit instructions use the same xmm registers from SSE so checking if those
+            // are enabled is sufficient.
+            return sse_enabled() && leaf7().all_bits_set<x86_cpuid_leaf7::ebx::avx2>();
+        }
```
AntoinePrv, line 882: Same, this is just one extra comment

serge-sans-paille:
Thanks!

PR #1357, opened 2026-06-02
Fix avx512vl_{128,256} correctness bugs + CI default-arch typo

DiamonDinoia:
Fixes bugs on the k-mask `avx512vl_{128,256}` arches, plus the CI typo hiding them:

- **CI**: `$CXX_FLAGS` → `$CXXFLAGS` in `linux.yml`; the `avx512vl_256` job now builds with its intended default arch.
- **swizzle**: 16-bit dynamic `swizzle` recursed infinitely (build failure); added a `uint16_t` base forwarding to avx2.
- **neq**: `_CMP_NEQ_OQ` → `_CMP_NEQ_UQ` so `neq(NaN, x)` is true.
- **decr_if/incr_if**: avx's `self ± batch(mask.data)` is wrong for k-mask bools; delegate to the select-based `common` path.

Covered by existing tests (`test_decr_if/incr_if/neq/neq_nan`, uint16 `swizzle`); now pass on both VL arches under SDE.

PR #1358, opened 2026-06-02
perf(avx512vl): native EVEX int64 shift, signed rotr, compress/expand

DiamonDinoia:
`avx512vl_{128,256}` inherit AVX2/SSE codegen for un-overridden kernels. Adds three pure-VL overrides (no DQ/CD, no VBMI2), each one EVEX op:

- **int64 `>>`** (signed): `vpsraq`/`vpsravq` (was ~5–10-instr emulation); unsigned/32-bit keep the inherited path.
- **signed `rotr`**: `vprorvq`/`vprolq` (dropped the `is_unsigned` guard to mirror `rotl`).
- **compress/expand**: `VPCOMPRESSD/Q`, `VCOMPRESSPS/PD` and the `VPEXPAND*` forms for 32/64-bit int + float/double.

8/16-bit `compress`/`expand` (`VPCOMPRESSB`/`VPCOMPRESSW`) require **AVX512_VBMI2** and are **intentionally deferred** — they fall through to the `common{}` fallback here and will be added later alongside new `avx512vbmi2_{128,256}` arch types.

Depends on #1357 (its swizzle fix unblocks the `avx512vl_256` test build).

### DIFF
--- a/include/xsimd/arch/xsimd_avx512vl_128.hpp
+++ b/include/xsimd/arch/xsimd_avx512vl_128.hpp
@@ -378,6 +378,48 @@ namespace xsimd
             return (typename batch_bool<double, A>::register_type)_mm_cmp_pd_mask(self, self, _CMP_UNORD_Q);
         }
 
+        // bitwise_rshift — signed int64 uses the native EVEX arithmetic shift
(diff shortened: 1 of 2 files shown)


serge-sans-paille:
The changes looks good \o/

PR #1363, opened 2026-06-04
feat: Add batch::to_native

AntoinePrv:
Using `batch::register_type` can sometimes be involved if it is heavily templated or if it is deducted from an `auto const&`. This makes casting to a native type when needed easier.

It should not be necessary to explicitly cast most of the time, but as we know with various compiler and intrinsicts implementations, we do what we can.


serge-sans-paille:
I'm not against the idea, but I pretty much dislike the name which is not explicit enough to me :-) 
Suggestions: `native_register`, `raw_register`, `get_register`, `read_register`, `native`, `to_native`...
I have a slight preference for the one including the `native` part

AntoinePrv:
Done and added a some doc on "Breaking out of xsimd".

JohanMabille (approved):

JohanMabille:
Let's get this one in!

PR #1367, opened 2026-06-11
feat: fused mul_hilo for 64-bit batches (shared 32x32->64 partials)

DiamonDinoia:
mul_hilo<uint64_t> previously fell through to the generic common path { mul_hi(x, y), x * y }


serge-sans-paille:
Thanks!

PR #1369, opened 2026-06-29
fix: route avx512vl_128 FMA ops through fma3<avx2_128>

DiamonDinoia:
avx512vl_128 derived from avx2_128, whose dispatch chain reaches no FMA3 kernel provider, so fnma/fnms fell back to the generic neg(x*y)+/-z form (an extra vxorpd ahead of vfmadd) even on a real AVX-512 build. Its 256-bit sibling avx512vl_256 already derives from fma3<avx2> and was unaffected.

Apply the same fma3<> inheritance pattern as avxvnni and avx512vl_256: introduce fma3<avx2_128> (a strict superset of avx2_128 that adds the 128-bit FMA3 kernels.

Validated under Intel SDE -skx: avx512vl_128 fnma/fnms/fma/fms/fmas emit 128-bit vf* (no vxorpd) and match scalar std::fma for float and double; avxvnni and fma3<avx2> codegen unchanged; all FMA-capable x86 arches compile across sse/avx2/avxvnni/avx512 flag sets.

### DIFF
--- a/include/xsimd/arch/xsimd_fma3_avx2_128.hpp
+++ b/include/xsimd/arch/xsimd_fma3_avx2_128.hpp
@@ -1,0 +1,47 @@
+/***************************************************************************
+ * Copyright (c) Johan Mabille, Sylvain Corlay, Wolf Vollprecht and         *
+ * Martin Renou                                                             *
+ * Copyright (c) QuantStack                                                 *
+ * Copyright (c) Serge Guelton                                              *
+ * Copyright (c) Marco Barbone                                              *
+ *                                                                          *
+ * Distributed under the terms of the BSD 3-Clause License.                 *
+ *                                                                          *
+ * The full license is in the file LICENSE, distributed with this software. *
+ ****************************************************************************/
+
+#ifndef XSIMD_FMA3_AVX2_128_HPP
+#define XSIMD_FMA3_AVX2_128_HPP
+
+#include "../types/xsimd_fma3_avx2_128_register.hpp"
+
+// Allow inclusion of xsimd_fma3_sse.hpp
+#ifdef XSIMD_FMA3_SSE_HPP
+#undef XSIMD_FMA3_SSE_HPP
+#define XSIMD_FORCE_FMA3_SSE_HPP
+#endif
+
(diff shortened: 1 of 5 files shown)


include/xsimd/arch/xsimd_fma3_avx2_128.hpp
```diff
@@ -0,0 +1,47 @@
...
+// Allow inclusion of xsimd_fma3_sse.hpp
+#ifdef XSIMD_FMA3_SSE_HPP
+#undef XSIMD_FMA3_SSE_HPP
+#define XSIMD_FORCE_FMA3_SSE_HPP
+#endif
+
+// Disallow inclusion of ./xsimd_fma3_sse_register.hpp
+#ifndef XSIMD_FMA3_SSE_REGISTER_HPP
+#define XSIMD_FMA3_SSE_REGISTER_HPP
+#define XSIMD_FORCE_FMA3_SSE_REGISTER_HPP
```
serge-sans-paille, line 27: I don't see the point of this `#define` ?

include/xsimd/arch/xsimd_fma3_avx2_128.hpp
```diff
@@ -0,0 +1,47 @@
...
+#endif
+
+// Include ./xsimd_fma3_sse.hpp but s/sse4_2/avx2_128
+#define sse4_2 avx2_128
```
serge-sans-paille, line 31: This is a very dirty trick. I think I'd rather have some code duplication that this.

DiamonDinoia: I basically copied it from here: https://github.com/xtensor-stack/xsimd/blob/master/include/xsimd/arch/xsimd_fma3_avx2.hpp

DiamonDinoia:
@serge-sans-paille what do you suggest here? We go for duplication? But, in that case we need to fix also the other case. Or we keep it as is for now and change this in a follow up?

include/xsimd/arch/xsimd_fma3_avx2_128.hpp
serge-sans-paille: of which I was the author :-/ let's roll.

PR #1370, opened 2026-07-06
perf(avx512f): single vpermt2 for zip_lo/zip_hi

DiamonDinoia:
The AVX-512 zip_lo/zip_hi lowered the interleave as per-128-bit unpacklo/unpackhi followed by a 3x insertf32x4 / extractf32x4 lane-fixup pile (~7 shuffle-port uops) needed to undo AVX-512's in-lane element ordering. A single 2-source cross-lane permute (vpermt2d/vpermt2q for integral, vpermt2ps/vpermt2pd for float/double) produces the same result in one port-5 uop.


serge-sans-paille:
This is *much* better. Thanks for looking at this.

serge-sans-paille (approved):

PR #1372, opened 2026-07-06
Add an explicit SSE2 path for batch_bool_constant masks where the three

DiamonDinoia:
Add an explicit SSE2 path for batch_bool_constant masks where the three
low lanes are active and lane 3 is zero ({T,T,T,F} = countr_one()==3).

load_masked<float>:  movsd (lanes 0,1) + movss (lane 2) + shufps into
                     [lo0, lo1, ss0, 0].
store_masked<float>: storel_pi (lanes 0,1) + movehl+store_ss (lane 2).

Without this both paths fell through to the common-arch scalar-buffer
fallback (zero-init std::array + element-wise scatter + aligned reload) (when compiler optimization was < -O3)

### DIFF
--- a/include/xsimd/arch/xsimd_sse2.hpp
+++ b/include/xsimd/arch/xsimd_sse2.hpp
@@ -1147,6 +1147,11 @@ namespace xsimd
             {
                 return _mm_loadh_pi(_mm_setzero_ps(), reinterpret_cast<__m64 const*>(mem + 2));
             }
+            else XSIMD_IF_CONSTEXPR(mask.countr_one() == 3)
+            {
+                __m128 const lo2 = _mm_castsi128_ps(_mm_loadl_epi64(reinterpret_cast<__m128i const*>(mem)));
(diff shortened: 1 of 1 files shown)


include/xsimd/arch/xsimd_sse2.hpp
```diff
@@ -1147,6 +1147,11 @@ namespace xsimd
             {
                 return _mm_loadh_pi(_mm_setzero_ps(), reinterpret_cast<__m64 const*>(mem + 2));
             }
+            else XSIMD_IF_CONSTEXPR(mask.countr_one() == 3)
+            {
+                __m128 const lo2 = _mm_loadl_pi(_mm_setzero_ps(), reinterpret_cast<__m64 const*>(mem));
```
serge-sans-paille: looks like `_mm_loadl_epi64` would be slightly better.

serge-sans-paille (changes requested):

DiamonDinoia:
Good idea!

serge-sans-paille (approved):

PR #1374, opened 2026-07-07
fix(neon): under-aligned store in generic swizzle kernel

### LINKED ISSUE #1260: SIGABRT on 32bit armhf in [batch]<xsimd::batch<unsigned char>>
Building xsimd 14.0.0 on debian fails on 32-bit armhf:
```
test 1
    Start 1: test_xsimd

1: Test command: /build/reproducible-path/xsimd-14.0.0/obj-arm-linux-gnueabihf/test/test_xsimd
1: Working Directory: /build/reproducible-path/xsimd-14.0.0/obj-arm-linux-gnueabihf/test
1: Test timeout computed to be: 10000000
1: test_xsimd: ./include/xsimd/types/xsimd_batch.hpp:596: void xsimd::batch<T, A>::store_aligned(U*) const [with U = unsigned char; T = unsigned char; A = xsimd::neon]: Assertion `((reinterpret_cast<uintptr_t>(mem) % A::alignment()) == 0) && "store location is not properly aligned"' failed.
1: [doctest] doctest version is "2.4.12"
1: [doctest] run with "--help" for options
1: ===============================================================================
1: ./test/test_batch.cpp:955:
1: TEST CASE:  [batch]<xsimd::batch<unsigned char>>
1:   horizontal_operations
1: 
1: ./test/test_batch.cpp:955: FATAL ERROR: test case CRASHED: SIGABRT - Abort (abnormal termination) signal
1: 
1: ===============================================================================
1: ./test/test_batch.cpp:955:
1: TEST CASE:  [batch]<xsimd::batch<unsigned char>>
1: 
1: DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
1:   horizontal_operations
1: 
1: ===============================================================================
1: [doctest] test cases:  13 |  12 passed | 1 failed | 274 skipped
1: [doctest] assertions: 706 | 706 passed | 0 failed |
1: [doctest] Status: FAILURE!
1/1 Test #1: test_xsimd .......................Subprocess aborted***Exception:   0.01 sec
test_xsimd: ./include/xsimd/types/xsimd_batch.hpp:596: void xsimd::batch<T, A>::store_aligned(U*) const [with U = unsigned char; T = unsigned char; A = xsimd::neon]: Assertion `((reinterpret_cast<uintptr_t>(mem) % A::alignment()) == 0) && "store location is not properly aligned"' failed.
```
Full log at https://buildd.debian.org/status/fetch.php?pkg=xsimd&arch=armhf&ver=14.0.0-4&stamp=1771339739&raw=0

This is with no specific machine flags provided. Previously we specified `-mfloat-abi=hard -mfpu=vfpv3` because of #889 , but that now fails with xsimd 14.0.0, triggering an error `unrecognizable insn`, see https://buildd.debian.org/status/fetch.php?pkg=xsimd&arch=armhf&ver=14.0.0-2&stamp=1771004640&raw=0

This is with PR #1177 reverted, because of the ARM64 bug #1232 (I'm not sure that that directly affects 32-bit arm)

vsaraikin:
Fixes #1260.

The generic neon swizzle kernel stores the batch into a local `std::array<T, size>`, which is only aligned to `alignof(T)` — 1 byte for the 8-bit batches that take this path (armv7 neon has no `vqtbl1q`; 8/16-bit swizzles fall through to the generic kernel). `store_aligned` requires the full batch alignment, so whenever the compiler places the array at a non-16-byte-aligned stack offset the alignment assertion fires: `reduce_max → kernel::swizzle(neon) → store_aligned` abort — exactly the Debian armhf backtrace in the issue.

This is not armhf-specific: the same abort reproduces on arm64 macOS with `batch<unsigned char, xsimd::neon>` compiled at `-O2` with assertions enabled:

```cpp
using batch_type = xsimd::batch<unsigned char, xsimd::neon>;
alignas(16) unsigned char raw[16] = { 0, 1, 2, 3, /* … */ };
auto b = batch_type::load_aligned(raw);
xsimd::reduce_max(b); // Assertion failed: store location is not properly aligned
```

The fix declares the buffer `alignas(A::alignment())`, matching the existing idiom in `xsimd_neon64.hpp` (`alignas(A::alignment()) double data[] = …`).

Adds a NEON-guarded regression test exercising the generic swizzle kernel and `reduce_max` for all four 8/16-bit batch types on the `neon` arch. One honest caveat: whether the assertion fires on the *old* code depends on the frame layout the compiler happens to pick (it fires in the issue's Debian build and in the `-O2` repro above, but not in a local `-O0` doctest build), so the test documents and exercises the path rather than deterministically aborting on regression.

Full test suite passes on arm64 macOS (391 test cases / 5928 assertions).

### DIFF
--- a/test/test_batch_manip.cpp
+++ b/test/test_batch_manip.cpp
@@ -330,23 +330,43 @@ TEST_CASE_TEMPLATE("[swizzle]", B, BATCH_SWIZZLE_TYPES)
     SUBCASE("rotate_right") { swizzle_test<B>().rotate_right(); }
 }
 
-#if XSIMD_WITH_NEON
 // Regression test for https://github.com/xtensor-stack/xsimd/issues/1260: the generic
-// neon swizzle kernel stored to an std::array aligned only to alignof(T), tripping the
-// alignment assertion in store_aligned for the 8- and 16-bit batches that use it
-TEST_CASE_TEMPLATE("[swizzle generic neon kernel]", B, xsimd::batch<uint8_t, xsimd::neon>, xsimd::batch<int8_t, xsimd::neon>, xsimd::batch<uint16_t, xsimd::neon>, xsimd::batch<int16_t, xsimd::neon>)
+// swizzle kernel stored to an std::array aligned only to alignof(T), tripping the
+// alignment assertion in store_aligned for the 8- and 16-bit batches that use it.
+// Run on every supported architecture, not just the default one, so the generic
+// kernels of the narrower archs are exercised too
+struct check_narrow_swizzle_and_reduce
 {
-    XSIMD_SWIZZLE_PATTERN_CASE(Reversor);
-    XSIMD_SWIZZLE_PATTERN_CASE(EvenThenOdd);
-    XSIMD_SWIZZLE_PATTERN_CASE(RotateRight1);
-    SUBCASE("reduce_max")
+    template <class Arch>
+    void operator()(Arch) const
     {
+        if (!Arch::available())
+        {
+            return;
+        }
+        check_for<uint8_t, Arch>();
+        check_for<int8_t, Arch>();
+        check_for<uint16_t, Arch>();
+        check_for<int16_t, Arch>();
+    }
+
+    template <class T, class Arch>
+    void check_for() const
+    {
+        using B = xsimd::batch<T, Arch>;
+        swizzle_test<B>().template run<xsimd::Reversor>();
+        swizzle_test<B>().template run<xsimd::EvenThenOdd>();
+        swizzle_test<B>().template run<xsimd::RotateRight1>();
         auto lhs = swizzle_test<B>::make_lhs();
         auto b = B::load_unaligned(lhs.data());
         CHECK_EQ(xsimd::reduce_max(b), lhs[B::size - 1]);
     }
+};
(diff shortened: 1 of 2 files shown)


test/test_batch_manip.cpp
```diff
@@ -330,6 +330,24 @@ TEST_CASE_TEMPLATE("[swizzle]", B, BATCH_SWIZZLE_TYPES)
     SUBCASE("rotate_right") { swizzle_test<B>().rotate_right(); }
 }
 
+#if XSIMD_WITH_NEON
```
DiamonDinoia: no need to run this only in the neon case you can run this as a unit test on all architectures.

serge-sans-paille:
LGTM once @DiamonDinoia comment has been addressed.

test/test_batch_manip.cpp
vsaraikin: Done — dropped the NEON guard; the test now iterates `xsimd::supported_architectures` (gated on `Arch::available()`), so the narrow swizzle kernels and `reduce_max` run for every enabled arch on each CI target.

serge-sans-paille:
Thanks :bow:

PR #1378, opened 2026-07-10
Avx256 constant mask plain moves

DiamonDinoia:
Half-confined constant-mask load_masked/store_masked shapes now emit plain moves instead of vpmaskmov/vmaskmov, which never store-forward on Intel and are microcoded on AMD.

### DIFF
--- a/include/xsimd/arch/xsimd_avx.hpp
+++ b/include/xsimd/arch/xsimd_avx.hpp
@@ -985,6 +985,19 @@ namespace xsimd
             {
                 return _mm256_insertf128_pd(_mm256_setzero_pd(), hi, 1);
             }
+
+            // 128-bit batch into the lower half, upper half zero (no instruction)
+            template <class A, class SrcA>
+            XSIMD_INLINE batch<float, A> zero_extend_lo(batch<float, SrcA> const& lo) noexcept
+            {
+                return _mm256_zextps128_ps256(lo);
+            }
+
+            template <class A, class SrcA>
+            XSIMD_INLINE batch<double, A> zero_extend_lo(batch<double, SrcA> const& lo) noexcept
+            {
+                return _mm256_zextpd128_pd256(lo);
+            }
         }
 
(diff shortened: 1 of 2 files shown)


DiamonDinoia:
@JohanMabille I think we can release after this. Just double checking. 

This can be further simplified now that we have avx2_128. I am editing as we speak

serge-sans-paille:
LGTM. We really need to extend the test suite with a way to ensure that a given high-level construct lowers to a single instruction. I already have basic support for that.

include/xsimd/arch/xsimd_avx.hpp
```diff
@@ -1021,16 +1034,27 @@ namespace xsimd
...
+            XSIMD_IF_CONSTEXPR(mask.prefix() == half_size)
             {
-                constexpr auto mlo = ::xsimd::detail::lower_half<half_arch>(batch_bool_constant<int_t, A, Values...> {});
-                const auto lo = load_masked(reinterpret_cast<int_t const*>(mem), mlo, convert<int_t> {}, Mode {}, half_arch {});
-                return bitwise_cast<T>(batch<int_t, A>(_mm256_zextsi128_si256(lo)));
+                return detail::zero_extend_lo<A>(half_batch::load(mem, Mode {}));
+            }
+            // lower 128-bit half: stay in the value domain so the half kernel can
+            // lower pure-prefix shapes to plain narrow moves (movss/movlps/movsd)
+            else XSIMD_IF_CONSTEXPR(mask.countl_zero() >= half_size)
```
serge-sans-paille, line 1048: nit: we could assert that `mask.countl_zero() > half_size`, right?

serge-sans-paille (approved):

include/xsimd/arch/xsimd_avx.hpp
DiamonDinoia: I will improve in a follow up PR that I am almost done drafting.

Issue #1267, opened 2026-03-04
Overload/SFINAE misses are hard to detect


AntoinePrv:
When, by mistake, one overload is not selected, the code usually falls back to `common` via inheritance.
Tests are silently passing but the behavior is not what is intended.

https://github.com/xtensor-stack/xsimd/blob/c3a8d37db2777c65c889e49143af36f32746d3da/include/xsimd/types/xsimd_common_arch.hpp#L27

One recent example is in #1266 

Possible solutions:
- Have some mode where (with a macro or something similar) one can disable all common fallbacks (or print a compile time warning).
  This enables power users to have a first diagnosis of their xsimd code without diving in assembly.
- HARD: Test for specific instructions in compile code.

AntoinePrv:
I came up with a more generic idea to extend the macro solution above. The idea is to trace "provenance" of the path of `batch` through xsimd.

There are many way in which this could be done: compile or run time, only in tests activated by a macros, using modern C++ versions behind a macro *etc*.

One way is to modify `batch` like

```cpp
template<typename T, typename A, typename P = User /* Provenance, unused outside of testing */>
class batch;
```

With all implicit conversions from `batch<T, A, P1>` to `batch<T, A, P2> so that the current syntax still holds.
Provenance could be a tag type, a tuple of tag types, or  could be gated behind C++20 and us an NTTP `std::array` of enums.

Then, in common, all functions would return `batch<T, A, append_t<P, provenance::common>>`. Other parts of the code that make no trivial dispatch could also append other things.

Finally in tests, we should be able to make assertions

```cpp
auto x = xsimd::bitwise_rshift(a, b);
static_assert(xsimd::provenance_never_comon(x));
```

This could also be done at runtime, by adding an `std::vector` attribute (behind a macro).

To make this not intrusive, we would derive utilities so that assertions, appends etc. disappear without needing to wrap them in `#ifdef` every time. Also provide default so that we can start small (say only common) and not have to change the whole codebase.

AntoinePrv:
Taking this opportunity to add another topic relevant to using overloads: forward declarations.

Today it is hard to reuse other functions (say `neg`, `all`...) inside the kernel because they may not be declared above.
[...]

PR #1383, opened 2026-07-16
Moved to C++17

JohanMabille:
Antoine found the issue I was thinking about: https://github.com/xtensor-stack/xsimd/issues/1267


AntoinePrv (approved):

DiamonDinoia:
Sure, I'll have a look, we should change the docs here too then, I'll double check everything

DiamonDinoia:
Hi I would update the docs here:
```
test/architectures/CMakeLists.txt:21   cxx_std_14 → cxx_std_17
README.md:46                            C++14 → C++17
docs/source/index.rst:72                C++14 → C++17
```
.github/workflows/cxx-versions.yml:17,43 still tests CMAKE_CXX_STANDARD=14 so this is obsolete now? 
benchmark/CMakeLists.txt:38–48 still has the old C++14 fallback probe

I think we should fix these in this commit, then we an make changes.

DiamonDinoia (changes requested):
Sorry, I added a comment before so could you read the above?

DiamonDinoia (approved):
All good!

PR #1386, opened 2026-07-21
endian.h not present on AIX

mehendarkarprajwal:
endian.h header file is not present on AIX, hence skipping its inclusion on AIX


serge-sans-paille:
looks like this header could be just removed instead of being guarded. Could you give this a try?

mehendarkarprajwal:
I removed the header file inclusion line and verified both on Power Linux and Power AIX, looks like no macros or symbols from `"endian.h"` are being referenced in this hpp file

JohanMabille (approved):

serge-sans-paille:
Thanks @mehendarkarprajwal

PR #1387, opened 2026-07-22
Fix gcc-15 'must be an 8-bit immediate' in avx_128 swizzle

DiamonDinoia:
g++-15 rejects a `constexpr` helper call passed inline to `_mm_permute_ps`/`_mm_permute_pd`:

```
error: the last argument must be an 8-bit immediate
```

It refuses to constant-fold a `constexpr` *function call* directly at a builtin immediate operand unless the value is first bound to a `constexpr` object. Every other `detail::mod_shuffle` / `detail::shuffle` call site in the codebase already does this; these two `swizzle` overloads in `xsimd_avx_128.hpp` were the only inline exceptions.


serge-sans-paille:
thanks!

PR #1391, opened 2026-07-28
Add runtime batch_bool mask overloads for complex batch load/store

DiamonDinoia:
The scalar batch<T, A> already accepted both batch_bool_constant (compile-time) and batch_bool (runtime) masks for load/store, but batch<std::complex<T>, A> was missing the runtime overloads.  This made it impossible to use a dynamic mask with complex batches.

Simple naive implementation for now just API parity.

### DIFF
--- a/include/xsimd/types/xsimd_batch.hpp
+++ b/include/xsimd/types/xsimd_batch.hpp
@@ -553,6 +553,8 @@ namespace xsimd
         // Compile-time mask overloads
         template <class U, bool... Values, class Mode = aligned_mode>
         [[nodiscard]] static XSIMD_INLINE batch load(U const* mem, batch_bool_constant<value_type, A, Values...> mask, Mode = {}) noexcept;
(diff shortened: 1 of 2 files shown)


JohanMabille (approved):

serge-sans-paille:
Just as  a side note: I'm unhappy with this PR. I think it should have been written in terms of dynamic masks on each real / imaginary part to benefit from existing dynamic mask API, or am I missing something @DiamonDinoia ?

DiamonDinoia:
you are correct. I just defaulted on common will work. I can make a follow  up PR but I was thinking of revamping the whole complex airthmetic for next release. I was making a roadmap but I did not have itme.

serge-sans-paille:
I can work on a followup before hand if you don't mind

DiamonDinoia:
Sure! Go for it! Any contribution or improvement is always welcome!

PR #1393, opened 2026-08-04
refactor: replace swizzle template recursion with C++17 constexpr 

DiamonDinoia:


include/xsimd/arch/common/xsimd_common_swizzle.hpp
```diff
@@ -27,146 +26,46 @@ namespace xsimd
...
-            XSIMD_INLINE constexpr bool is_cross_lane() noexcept
-            {
-                static_assert(sizeof...(Vs) >= 1, "Need at least one lane");
-                return cross_impl<0, sizeof...(Vs), sizeof...(Vs) / 2, Vs...>::value;
+            // both halves read the same indices, all taken from the Hi ? high : low half
+            template <typename T, bool Hi, T... Vs>
+            XSIMD_INLINE constexpr bool is_dup_from_half() noexcept
+            {
+                constexpr std::size_t half = sizeof...(Vs) / 2;
+                constexpr T lo = Hi ? static_cast<T>(half) : T(0);
```
serge-sans-paille: i think this would be much easier to understand if implemented with the `Hi == true` logic speerated from the `Hi == false` logic.

```
{
    if constexpr(Hi) { ...}
    else { ...}
}
```

esp. for this condition `if (v[i] < lo || v[i] >= hi || v[i + half] != v[i])` which looks different from `i<H but V>=H) or (i>=H but V<H)`

AntoinePrv (approved):

include/xsimd/arch/common/xsimd_common_swizzle.hpp
DiamonDinoia: I think this way might be clearer and have less duplication. If you prefer your way, I can make the changes.

PR #1400, opened 2026-08-12
fix: make the avx512vl_128 and avx512vl_256 arches build and pass

DiamonDinoia:
1. xsimd_common_fwd.hpp — forward-declare incr_if/decr_if. The VL sub-arches call the common versions unqualified from headers included before xsimd_common.hpp; dependent lookup can't find them. 16 compile errors on vl_128.
2. xsimd_avx512f.hpp — tobitset<4> shift 24→28, tobitset<2> 8→14 (correct shift is 8N-N). Only reachable for batch_bool of size 4 and 2, i.e. only those two arches.
4. xsimd_avx.hpp — sadd/ssub build the sign mask with other < 0 instead of reinterpreting a shifted vector as batch_bool, which breaks on k-register booleans. Also drops sadd<int64_t> 24→14 and ssub<int64_t> 21→11 instructions on AVX2.

### DIFF
--- a/include/xsimd/arch/xsimd_avx.hpp
+++ b/include/xsimd/arch/xsimd_avx.hpp
@@ -1477,10 +1477,10 @@ namespace xsimd
         {
             if (std::is_signed_v<T>)
             {
-                auto mask = (other >> (8 * sizeof(T) - 1));
+                auto negative = other < batch<T, A>(T(0));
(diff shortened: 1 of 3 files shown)


AntoinePrv:
Seems we came to the same conclusion in #1412.

PR #1401, opened 2026-08-13
feat: add per-lane popcount on batch

DiamonDinoia:
Count the bits set in each element of an integer batch. The common kernel is the SWAR fold; NEON uses CNT with pairwise widening adds, SVE uses svcnt_x, and WASM uses i8x16.popcnt with pairwise widening extends, falling back to the common kernel for 64-bit elements.

I will add kernels in the future too when I have more of this. I need the bitwise operation for a Morton transform library I am writing. 

Creating a new file because I plan to add more functions in the future: 
countl_zero/countr_zero, bit_reverse, multishift, bit_deposit/bit_extract,  bit_permute + bit_permute_constant, bit_matmul, GFNI kernels. 

Reviewed by: Claude Opus 5 <noreply@anthropic.com>


serge-sans-paille:
With the appropriate credits, https://github.com/WojciechMula/sse-popcount seems to provide relevant implementation for arch-specifics

DiamonDinoia:
Good point. I also should have listed that the algorithm I implemented I found it on wikipedia.

serge-sans-paille:
See also https://github.com/kimwalisch/libpopcnt/tree/master based on the paper above for an AVX2 implementation you could add to this PR (with proper credit of course)

include/xsimd/arch/xsimd_neon.hpp
```diff
@@ -3680,6 +3680,37 @@ namespace xsimd
...
+            // adds, after Wojciech Muła's popcnt_neon_vcnt,
+            // https://github.com/WojciechMula/sse-popcount
+            uint8x16_t counts = vcntq_u8(bitwise_cast<uint8_t>(self).data);
+            if constexpr (sizeof(T) == 1)
+                return bitwise_cast<T>(batch<uint8_t, A>(counts));
+            else if constexpr (sizeof(T) == 2)
+                return bitwise_cast<T>(batch<uint16_t, A>(vpaddlq_u8(counts)));
+            else
+            {
+#if defined(__ARM_FEATURE_DOTPROD)
```
DiamonDinoia: @serge-sans-paille you know about arm more than I do. Is this a good idea? should this be a new architecture entirely? If so I will purge it from here and add it in a separate PR.

serge-sans-paille: yeah, it should be a proper architecture :-/

include/xsimd/arch/common/xsimd_common_bitwise.hpp
```diff
@@ -0,0 +1,70 @@
...
+{
+
+    namespace kernel
+    {
+
+        using namespace types;
+
+        namespace detail
+        {
+            // bit i is set when i / s is even: 0x55.. for s == 1, 0x33.. for
```
serge-sans-paille: I'm not a big fan of neither the name, nor the explanation, nor the implementation.
What about the more generic: 

```
// Create an integer of type U with Pattern P repeated as much as possible within U
template<class U, class P>
constexpr U repeat_pattern() {
    static_assert(P!=0);
    constexpr size_t pattern_size = sizeof(U) * CHAR_BITS - std::count_lzero(P); // FIXME: need to provide our implementation
    static_assert(sizeof(U) * CHAR_BIT % pattern_size == 0);
    constexpr size_t rep_count = sizeof(U) * CHAR_BIT % pattern_size;
    U result = P;
    for(size_t i = 1; i < rep_count; ++i) {
        result |= P << ( pattern_size * i );
    }
}
```

serge-sans-paille (commented):
Could you please add (and test) a scalar version for this, that uses the version you provide in xsimd_common_bitwise.hpp?

include/xsimd/arch/common/xsimd_common_bitwise.hpp
```diff
@@ -0,0 +1,70 @@
...
+            b_type const m2(detail::alternating_mask<U>(2));
+            x = (x & m2) + ((x >> 2) & m2);
+            x = (x + (x >> 4)) & b_type(detail::alternating_mask<U>(4));
+            if constexpr (bits >= 16)
+                x = (x + (x >> 8)) & b_type(detail::alternating_mask<U>(8));
+            if constexpr (bits >= 32)
+                x = (x + (x >> 16)) & b_type(detail::alternating_mask<U>(16));
+            if constexpr (bits >= 64)
+                x = (x + (x >> 32)) & b_type(detail::alternating_mask<U>(32));
+            return bitwise_cast<T>(x);
```
serge-sans-paille: you could / should check if `__builtin_popcount` is available and if so, use it.

include/xsimd/arch/xsimd_neon.hpp
DiamonDinoia: Okay I will do a follow up PR with the missing architectures.

include/xsimd/arch/common/xsimd_common_bitwise.hpp
DiamonDinoia: Good point. I was wondering how to simplify this

include/xsimd/arch/common/xsimd_common_bitwise.hpp
DiamonDinoia: We fallback on it in the scalar case. Am I missing something? This is the fallback in case that builtin is not available. I am thinking of msvc.

PR #1402, opened 2026-08-14
Add LoongArch64 support

Chris-wh233:
The LoongArch architecture (LoongArch) is an Instruction Set Architecture (ISA) that has a RISC style.
Documentations:
ISA:
https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html
ABI:
https://loongson.github.io/LoongArch-Documentation/LoongArch-ELF-ABI-EN.html
More docs can be found at:
https://loongson.github.io/LoongArch-Documentation/README-EN.html

Tested natively on a LoongArch64 machine.
### LSX
Build flags:  -mlsx
Result:
[doctest] test cases:  409 |  409 passed | 0 failed | 0 skipped
[doctest] assertions: 5976 | 5976 passed | 0 failed
[doctest] Status: SUCCESS!

### LASX
Build flags:  -mlasx
[doctest] test cases:  409 |  409 passed | 0 failed | 0 skipped
[doctest] assertions: 7912 | 7912 passed | 0 failed
[doctest] Status: SUCCESS!


serge-sans-paille:
Thanks for the PR. Before going forward: if no CI is setup for this, we cannot maintain that contribution. Can you make the extra step?

Once CI is setup, I'll have a look at the actual implementation ;-)

Chris-wh233:
Thank you for your reply. I have submitted CI; please review it.

serge-sans-paille:
While I'm reviewing it, can you give us some context? Why are you adding loongArch64 support xsimd? Are you planning to use it for some project? And more important to us, do you feel you can provide some support in maintaining that arch with xsimd?
Thanks :bow:

serge-sans-paille:
From a high level perspective: it seems to me we should have two different arch, lsx and lasx, maybe with lasx depending from lsx. Much like avx-family depends from sse-family.
It also means two different testing setup etc.

Chris-wh233:
Thanks for the feedback! lsx and lasx are already two distinct arch types with two separate testing setups. But the LASX ⊃ LSX superset relationship was not reflected in the type hierarchy. I've addressed this in the latest commit by making lasx derive from lsx.

Chris-wh233:
This came up while porting Milvus to LoongArch64, since Milvus depends on xsimd. The current Milvus release doesn't need a newer xsimd (it's still pinned to 9.0.1), but without upstream support the dependency chain stays blocked on LoongArch even after future upgrades. So the main goal is to unblock that dependency chain — for Milvus, and more broadly for any project that wants to run on LoongArch.

On maintenance: yes, I can provide support, on a best-effort basis — this isn't my full-time focus, so my response time may vary.

.github/workflows/cross-loongarch.yml
```diff
@@ -0,0 +1,65 @@
...
+    - name: Build
+      run: cmake --build _build
+    - name: Set CPU feature test expectations
+      run: |
+        echo "XSIMD_TEST_CPU_ASSUME_SSE4_2=0" >> "$GITHUB_ENV"
+        echo "XSIMD_TEST_CPU_ASSUME_NEON64=0" >> "$GITHUB_ENV"
+        echo "XSIMD_TEST_CPU_ASSUME_SVE=0" >> "$GITHUB_ENV"
+        echo "XSIMD_TEST_CPU_ASSUME_RVV=0" >> "$GITHUB_ENV"
+        echo "XSIMD_TEST_CPU_ASSUME_LSX=1" >> "$GITHUB_ENV"
+        echo "XSIMD_TEST_CPU_ASSUME_LASX=1" >> "$GITHUB_ENV"
```
serge-sans-paille: I would expect two different runs with different qemu invocation and different feature tests, one for LSX and one for LASX?

docs/Doxyfile
```diff
@@ -24,6 +24,7 @@ INPUT             = ../include/xsimd/types/xsimd_api.hpp \
                     ../include/xsimd/types/xsimd_neon64_register.hpp \
                     ../include/xsimd/types/xsimd_neon_register.hpp \
                     ../include/xsimd/types/xsimd_rvv_register.hpp \
+                    ../include/xsimd/types/xsimd_lsx_register.hpp \
```
serge-sans-paille, line 28: there should be an lsx register file and an lasx register file

include/xsimd/arch/xsimd_isa.hpp
```diff
@@ -154,6 +154,10 @@
 #include "./xsimd_rvv.hpp"
 #endif
 
+#if XSIMD_WITH_LSX || XSIMD_WITH_LASX
+#include "./xsimd_lsx.hpp"
```
serge-sans-paille, line 158: and a split between the different implementation here. the idea is that lsx support could be on even when lasx is not supported

include/xsimd/arch/xsimd_lsx.hpp
```diff
@@ -0,0 +1,573 @@
...
+            XSIMD_INLINE typename batch<T, A>::register_type loongarch_from_bits(loongarch_unsigned_register_t<T, A> const& value) noexcept
+            {
+                return loongarch_bit_cast<typename batch<T, A>::register_type>(value);
+            }
+
+            template <class A>
+            XSIMD_INLINE batch<float, A> fast_cast(batch<std::int32_t, A> const& self, batch<float, A> const&, requires_arch<loongarch>) noexcept
+            {
+                typename batch<float, A>::register_type result {};
+                for (std::size_t i = 0; i < batch<float, A>::size; ++i)
```
serge-sans-paille: this seems terribly slow, definitively not a *fast cast*

include/xsimd/arch/xsimd_lsx.hpp
```diff
@@ -0,0 +1,573 @@
...
+                {
+                    result[i] = static_cast<float>(self.data[i]);
+                }
+                return result;
+            }
+
+            template <class A>
+            XSIMD_INLINE batch<std::int32_t, A> fast_cast(batch<float, A> const& self, batch<std::int32_t, A> const&, requires_arch<loongarch>) noexcept
```
serge-sans-paille: same here

include/xsimd/arch/xsimd_lsx.hpp
```diff
@@ -0,0 +1,573 @@
...
+                    result[i] = static_cast<std::int32_t>(self.data[i]);
+                }
+                return result;
+            }
+        }
+
+        // abs
+        template <class A, class T, class = std::enable_if_t<std::is_floating_point_v<T>>>
+        XSIMD_INLINE batch<T, A> abs(batch<T, A> const& self, requires_arch<loongarch>) noexcept
+        {
```
serge-sans-paille: there should be a better way!

serge-sans-paille (changes requested):
I've only started the review, but it appears to me that many operations are implemented as scalar operations and not vector operations. this is not what we expect here :-/

The documentations seems scarce, but https://jia.je/unofficial-loongarch-intrinsics-guide/migrating_sse/ and https://jia.je/unofficial-loongarch-intrinsics-guide/migrating_avx/ should help filling the gap.

Does this make sense?

Chris-wh233:
@serge-sans-paille Thank you for the feedback! I have reworked the LoongArch backend accordingly:
1. Split the LSX and LASX support into separate backends.
2. Replaced the scalar-loop implementations in the original version with LSX/LASX intrinsics.

Chris-wh233:
This is the official documentation of the toolchain conventions of the LoongArch Architecture:
https://github.com/loongson/la-toolchain-conventions

include/xsimd/config/xsimd_config.hpp
```diff
@@ -512,6 +512,39 @@
...
+#else
+#define XSIMD_TARGET_LOONGARCH64 0
+#endif
+
+/**
+ * @ingroup xsimd_config_macro
+ *
+ * Set to 1 if the 128-bit Loongson SIMD Extension is available.
+ */
+#if XSIMD_TARGET_LOONGARCH64 && (defined(__loongarch_sx) || defined(__loongarch_asx))
```
serge-sans-paille: Is there a case where `__loongarch_asx` is defined but not `__loongarch_sx` ? If not, better keep the condition simple here.

include/xsimd/config/xsimd_cpu_features_loongarch.hpp
```diff
@@ -0,0 +1,88 @@
...
+        constexpr unsigned long loongarch_hwcap_lsx = 1ul << 4;
+#endif
+#ifdef HWCAP_LOONGARCH_LASX
+        constexpr unsigned long loongarch_hwcap_lasx = HWCAP_LOONGARCH_LASX;
+#else
+        // Possibly missing on older Linux distributions
+        constexpr unsigned long loongarch_hwcap_lasx = 1ul << 5;
+#endif
+        return hwcap().has_feature(loongarch_hwcap_lsx)
+            || hwcap().has_feature(loongarch_hwcap_lasx);
```
serge-sans-paille: same here, I wonder if only checking for `lsx` would be enough?

include/xsimd/config/xsimd_cpu_features_loongarch.hpp
```diff
@@ -0,0 +1,88 @@
...
+#else
+        return XSIMD_WITH_LSX || XSIMD_WITH_LASX;
```
serge-sans-paille: same :-)

test/CMakeLists.txt
```diff
@@ -97,7 +97,7 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER_ID MATCHES "GNU"
         message(STATUS "CMAKE_CXX_LINK_EXECUTABLE: ${CMAKE_CXX_LINK_EXECUTABLE}")
     elseif(${CMAKE_SYSTEM_PROCESSOR} MATCHES "^ppc64"  OR ${CMAKE_SYSTEM_PROCESSOR} MATCHES "aarch64")
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mcpu=${TARGET_ARCH} -mtune=${TARGET_ARCH}")
-    elseif(${CMAKE_SYSTEM_PROCESSOR} MATCHES "riscv64")
+    elseif(${CMAKE_SYSTEM_PROCESSOR} MATCHES "riscv64|loongarch64")
         # Nothing specific
```
serge-sans-paille, line 103: nit: looks more idiomatic (wrt. surrounding code) to just

```
elseif(${CMAKE_SYSTEM_PROCESSOR} MATCHES "loongarch64")
    # Nothing specific
```

we're not at a point where we optimise our test cmakefiles :-)

test/test_batch_complex.cpp
```diff
@@ -199,6 +200,11 @@ struct batch_complex_test
             imag[i] = lhs[i].imag();
             tmp[i] = value_type(real[i]);
         }
+
+#if XSIMD_WITH_LSX || XSIMD_WITH_LASX || XSIMD_WITH_SVE || XSIMD_WITH_RVV
```
serge-sans-paille, line 204: shouldn't this always work? Why the arch restriction?

include/xsimd/arch/xsimd_lasx.hpp
```diff
@@ -0,0 +1,1012 @@
...
+            {
+                return __lasx_xvftintrz_lu_d(self.data);
+            }
+        }
+
+        // abs
+        template <class A>
+        XSIMD_INLINE batch<float, A> abs(batch<float, A> const& self, requires_arch<lasx>) noexcept
+        {
+            return detail::lasx_from_int<float, A>(__lasx_xvbitclri_w(detail::lasx_to_int(self), 31));
```
serge-sans-paille: looks like every scalar could have the same implementation?

```
        template <class A, class T>
        XSIMD_INLINE batch<T, A> abs(batch<T, A> const& self, requires_arch<lasx>) noexcept
        {
            return detail::lasx_from_int<T, A>(__lasx_xvbitclri_w(detail::lasx_to_int(self), CHAR_SIZE * sizeof(T) - 1));
        }
```

include/xsimd/arch/xsimd_lasx.hpp
```diff
@@ -0,0 +1,1012 @@
...
+        XSIMD_INLINE batch<double, A> add(batch<double, A> const& self, batch<double, A> const& other, requires_arch<lasx>) noexcept
+        {
+            return __lasx_xvfadd_d(self.data, other.data);
+        }
+
+        // all/any
+        template <class A, class T>
+        XSIMD_INLINE bool all(batch_bool<T, A> const& self, requires_arch<lasx>) noexcept
+        {
+            constexpr std::uint64_t all_bits = (std::uint64_t(1) << batch_bool<T, A>::size) - 1;
```
serge-sans-paille: see `xsimd::utils::make_low_mask``

include/xsimd/arch/xsimd_lasx.hpp
```diff
@@ -0,0 +1,1012 @@
...
+        // select
+        template <class A, class T>
+        XSIMD_INLINE batch<T, A> select(batch_bool<T, A> const& cond, batch<T, A> const& true_br, batch<T, A> const& false_br, requires_arch<lasx>) noexcept
+        {
+            return detail::lasx_from_int<T, A>(__lasx_xvbitsel_v(detail::lasx_to_int(false_br), detail::lasx_to_int(true_br), cond.data));
+        }
+
+        template <class A, class T, bool... Values>
+        XSIMD_INLINE batch<T, A> select(batch_bool_constant<T, A, Values...> const&, batch<T, A> const& true_br, batch<T, A> const& false_br, requires_arch<lasx>) noexcept
+        {
```
serge-sans-paille, line 827: Future work (and common with other architectures), there may be common patterns we want to detect at compile time there, although compilers are getting quite good at it.

serge-sans-paille (changes requested):
Looks quite good to me. i've made a few minot remarks, and I'm trusting validation for the architecture details.

Thanks a lot for going through this 🙇

test/test_batch_complex.cpp
Chris-wh233: I guarded the test because only LSX, LASX, SVE, and RVV currently provide generic complex set implementations supporting both float and double. Removing the guard causes compilation failures on x86 and Arm configurations.
Would you prefer to keep this regression test guarded for the backends that currently support the constructor, or add a common fallback that separates the complex values into real and imaginary batches, allowing the test to run unconditionally?

include/xsimd/config/xsimd_config.hpp
Chris-wh233: LASX implies LSX, and enabling LASX also defines __loongarch_sx. I will simplify it in subsequent submissions

include/xsimd/arch/xsimd_lasx.hpp
Chris-wh233: They can share a templated implementation, but the template must still select the appropriate intrinsic(_w / _d) based on the element width.

test/test_batch_complex.cpp
serge-sans-paille: let's add this for future work. I *think* we could make this an arch property ("supports double") and if constexpr on it.

serge-sans-paille (approved):
This looks good to me.

Can you (a) rebase on `master` branch, (b) make sure clang-format test passes and (c) squash history and edit commit logs appropriately?

serge-sans-paille:
Thanks a lot for @Chris-wh233 for contributing this :bow:

PR #1404, opened 2026-08-27
docs: update supported compiler versions to match the C++17 requirement

DiamonDinoia:
The listed versions (MSVC 2015 update 2, g++ 4.9, clang 4.0/3.7) predate C++17 and cannot compile the library. Minimum versions with full C++17 language support per the official compiler documentation.


Alex-PLACET:
I propose to reference those we have in the CI:
gcc 10
clang 16
MSVC 2022
If we plan to support older compilers, we should add them in the CI

DiamonDinoia:
I have no problem with that. I personally dislike maintaining very old compilers but I leave htis decisions to you all.

JohanMabille:
It's indeed complicated to maintain compilers that we don't test in the CI, so I agree with Alexis, let's require those from the CI as minimal versions

DiamonDinoia:
We can always say something like xsimd should work with any c++17 compiler,
however we ufficially support (the ones tested in CI).

serge-sans-paille:
@DiamonDinoia : I leave it to you to do the doc update, based on this thread ;-)

serge-sans-paille:
thanks

PR #1405, opened 2026-08-28
Add release github workflow

Alex-PLACET:


serge-sans-paille:
I'm not personally a big fan of automated release, I like the simple 'tag and done' approach. Why do we need more?

Alex-PLACET:
I just propose to automate it, which will be a guarantee to be sure to update header and changelog as expected for each release.

serge-sans-paille:
> I just propose to automate it, which will be a guarantee to be sure to update header and changelog as expected for each release.

What about writing a script, independent from github, that checks that a release correctly formatted? I assume we could trigger that check as an extra rule for branch named as `release/x.y.z` and that would give a nice extrab check that packager can also run on their machine.
 
I'm esp. not a big fan of automatically generated Changelog, because people can just read the git log if that's what they need.
On the other hand a nicely crafted changelog like https://github.com/xtensor-stack/xsimd/blob/master/Changelog.rst#1420 provides high level summary of the git log, and provides a net value to downstream users or packagers.

PR #1406, opened 2026-09-01
Experiment with intrinsict generation

AntoinePrv:
Second experiment to generate only overloads of intrinsics from a (semi) official data file (`overload/neon.hpp`).
The code in `neon.hpp`/`neon64.hpp` is hand modified, but greatly reduced.

This is work in progress, showing what could be made before we commit to this approach.

From a discussion with @JohanMabille:
Generating code is fine, but we may want to then check-in the code and keep maintaining it by hand since the generator would otherwise grow in complexity.
The generator can be kept for reference but would mostly be used to transition purposes.


JohanMabille (commented):
I really like the idea! For now, the only remark I have would be to rename `overload` into `wrapper` or something more related to what this folder contains.

PR #1407, opened 2026-09-02
perf: double fast_cast

DiamonDinoia:
Common double -> int64_t cast.  Adding 1.5 * 2^52 puts an integer |v| < 2^51 exactly in the mantissa so subtracting the bit patterns reads it back out.

3.743 -> 0.551 ns/element on a 128 KB L2-resident loop (Xeon w5-3435X, gcc 13.3
-O3, min of 300 reps).

### DIFF
--- a/include/xsimd/arch/common/xsimd_common_details.hpp
+++ b/include/xsimd/arch/common/xsimd_common_details.hpp
@@ -251,6 +251,26 @@ namespace xsimd
                 return bitwise_cast<int64_t>(self);
             }
 
+            // Common double -> int64_t cast.  Adding 1.5 * 2^52 puts an integer
+            // |v| < 2^51 exactly in the mantissa, so subtracting the bit patterns
+            // reads it back out.  trunc first, then split into halves that small.
+            // An architecture reaching this overload needs a non-common trunc:
+            // the common one goes through to_int and would recurse into here.
+            template <class A>
+            XSIMD_INLINE batch<int64_t, A> fast_cast(batch<double, A> const& x, batch<int64_t, A> const&, requires_arch<common>) noexcept
+            {
+                using batch_type = batch<double, A>;
+                batch_type magic(0x1.8p52);
+                batch<int64_t, A> magic_i = bitwise_cast<int64_t>(magic);
+                batch_type t = trunc(x);
+                batch_type hi = trunc(t * batch_type(0x1p-32)); // |hi| <= 2^31
+                batch_type lo = t - hi * batch_type(0x1p32); // exact, |lo| < 2^32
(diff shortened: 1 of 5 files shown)


serge-sans-paille:
I did a few experiments and the following failed straight away:


```
include <xmmintrin.h>
#include <stdio.h>
#include <math.h>
__m128d vtrunc(__m128d self) {

            // The common implementation goes through batch_cast<int64_t>, which on sse2
            // is itself built on trunc.  Adding 2^52 rounds |self| to an integer whatever
            // the rounding mode is; one conditional decrement turns that into floor.
            __m128d signmask = _mm_set1_pd(-0.);
            __m128d t2n = _mm_set1_pd(4503599627370496.); //  2^52
            __m128d s = _mm_and_pd(self, signmask);
            __m128d v = _mm_andnot_pd(signmask, self); // |self|
            __m128d d0 = _mm_add_pd(v, t2n);
            __asm__ volatile("" : "+x"(d0));
            __m128d d = _mm_sub_pd(d0, t2n);
            d = _mm_sub_pd(d, _mm_and_pd(_mm_cmpgt_pd(d, v), _mm_set1_pd(1.))); // floor(|self|)
            d = _mm_andnot_pd(signmask, d); // roundTowardNegative makes 2^52 - 2^52 negative
            __m128d small = _mm_cmplt_pd(v, t2n);
            return _mm_or_pd(s, _mm_or_pd(_mm_and_pd(small, d), _mm_andnot_pd(small, v)));
}

#include <stdlib.h>
int main() {
  long n = random();
  double r = 0.000071; //drand48() * n;
  __m128d d = {r};
  __m128d tmp = vtrunc(d);
  printf("%lf => %lf %lf\n", r, d, tmp[0]);
}
```

nan, -nan, inf and -inf are working correctly

DiamonDinoia:
Oh, thanks for spotting 2^52! I completely forgot to test it

DiamonDinoia:
Hi @serge-sans-paille, I do not seem to be able to reproduce. Am I missing something? https://godbolt.org/z/EK84xx5qM

serge-sans-paille:
> Hi @serge-sans-paille, I do not seem to be able to reproduce. Am I missing something? https://godbolt.org/z/EK84xx5qM

indeed, my test was incorrect. I fixed it and let it run for a while, it did not uncover any issues. sorry for the extra delay :bow:

serge-sans-paille:
The RISC6V failure looks unrelated (it also happens on https://github.com/xtensor-stack/xsimd/pull/1410 for instance). We should have a look, but I don't think your patch is the culprit there.

DiamonDinoia:
Yes, it is the version of qemu not being pinned the new patch release seem
to have a problem.

PR #1408, opened 2026-09-04
Enable the ARM64 target on the ARM64EC ABI.

degasus:
ARM64EC enables both _M_AMD64 and _M_ARM64EC. We already exclude ARM64EC from the SSE2 detection, but it was missing for the X64 target detection (by me in #1152, apparently on purpose).

However we're now fine to enable the ARM64 target. At least my kernels compile fine (as good as on the native arm64) on MSVC.

Related: https://techcommunity.microsoft.com/blog/windowsosplatform/getting-to-know-arm64ec-defines-and-intrinsic-functions/2957235

Tested on a local Windows arm64 machine with the arm64ec ABI:
now:
```
[doctest] doctest version is "2.4.9"
[doctest] run with "--help" for options
===============================================================================
[doctest] test cases:  406 |  406 passed | 0 failed | 0 skipped
[doctest] assertions: 5999 | 5999 passed | 0 failed |
[doctest] Status: SUCCESS!
```

before enabling XSIMD_TARGET_ARM64:
```
[doctest] doctest version is "2.4.9"
[doctest] run with "--help" for options
===============================================================================
[doctest] test cases:  94 |  94 passed | 0 failed | 0 skipped
[doctest] assertions: 803 | 803 passed | 0 failed |
[doctest] Status: SUCCESS!
```


serge-sans-paille:
LGTM. This does mean we should add that architecture to our CI target, can you give it a try?

degasus:
> LGTM. This does mean we should add that architecture to our CI target, can you give it a try?

Done. CMake's recormendation seems to be to provide `/arm64EC` as envvar, else you need to set also linking flags. But we overwrite the linker flags, I had to fix this as well.

The strangest issue I had with the test is the git's msys2 envvar handling: When calling cmake (native) from bash (msys2), it converts `/arm64EC` to `C:/Program files/GIT/arm64EC`... A single space works around this issue now...

.github/workflows/windows.yml
```diff
@@ -154,6 +159,9 @@ jobs:
       uses: actions/checkout@v6
     - name: Setup
       run: cmake -B _build -DBUILD_TESTS=ON -DDOWNLOAD_DOCTEST=ON -DBUILD_BENCHMARK=ON -DBUILD_EXAMPLES=ON -DCMAKE_BUILD_TYPE=Release -G Ninja
+      env:
+        CFLAGS: ${{ matrix.abi.cflags }}
+        CXXFLAGS: ${{ matrix.abi.cflags }}
     - name: Build
       run: cmake --build _build
     - name: Testing xsimd
```
serge-sans-paille, line 167: This is not testing the x86 emulation part, but I think it's ok, not our business to check that.

serge-sans-paille (approved):
Looks good. Can you rebase on master brnach, this should make CI green again.

.github/workflows/windows.yml
degasus: IMO it would be our business to test the x64 intrinsics in arm64ec mode, but I've disabled them on purpose. On arm64ec, I think we should only allow to generate the native neon intrinsics instead.

PR #1409, opened 2026-09-04
Fix comilation of neon/shifts_all_positive on msvc.

degasus:
This fixes a regression since b8f7ebef1a5197cb2df4856d133d2c8b236988e4

However this seems to only affects debug builds...


AntoinePrv (approved):

AntoinePrv:
Failures are unrelated and happening in main

PR #1410, opened 2026-09-04
Fix neon64/swizzle on MSVC.

degasus:
vreinterpretq is a no-op here, not even converting the type. As we provide the batch into vreinterpretq, we keep its strong typing and on return, CL tells me that it cannot convert to the output type.

Please don't treat this as high quality code - more as a bug report - even if it fix compiling my kernels.
After looking at this code for roughly 30 minutes, I assume a real fix might be to encapsulate all vreinterpretq calls into something strongly typed, even if this sounds annoying.

I actually wonder, is there a test which checks if swizzle compiles?


AntoinePrv:
> After looking at this code for roughly 30 minutes, I assume a real fix might be to encapsulate all vreinterpretq calls into something strongly typed, even if this sounds annoying.

This is in line with what I have been doing in #1406 (check the `overload/neon.hpp`), although it is not meant to be merged at the moment.

@degasus CI is a mess on #1406 but it did run for me locally. Feel free to check it out if you have time.

PR #1411, opened 2026-09-08
[ci] Frail attempt to fix riscv qemu run

serge-sans-paille:


serge-sans-paille:
@DiamonDinoia as the author of 8b8168722d1ab425186efd2cddb1e29aba8b9c02, can you review this patch? Thanks!

DiamonDinoia (approved):
LGTM, less code is better. It seems gcc emits an instruction that kills qemu? While clang does no?

Interesting.

PR #1412, opened 2026-09-14
Fix linux CI and AVX512-128/256

AntoinePrv:


.github/workflows/linux.yml
```diff
@@ -161,3 +162,33 @@ jobs:
           export XSIMD_TEST_CPU_ASSUME_MANUFACTURER="intel,amd"
           ./test_xsimd
         fi
+
+  build-extra:
```
AntoinePrv, line 167: I had to split out the benchmark and architecture tests because their CMake did not play well with `XSIMD_DEFAULT_ARCH` (needed to force lower sizes).

AntoinePrv:
@serge-sans-paille these architectures were not properly tested in CI.
Would you mind reviewing the fixes?

.github/workflows/linux.yml
```diff
@@ -119,7 +119,7 @@ jobs:
           CMAKE_EXTRA_ARGS="$CMAKE_EXTRA_ARGS -DTARGET_ARCH=knm"
         fi
         if [[ '${{ matrix.sys.flags }}' == 'i386' ]]; then
-          CXX_FLAGS="$CXX_FLAGS -m32"
```
serge-sans-paille, line 122: that's so error prone, thanks for fixing this.

include/xsimd/arch/xsimd_avx.hpp
```diff
@@ -1726,12 +1726,12 @@ namespace xsimd
                                           { return ssub(batch<T, sse4_2>(s), batch<T, sse4_2>(o)); },
                                           self, other);
             }
-            else if (std::is_signed_v<T>)
+            else if constexpr (std::is_signed_v<T>)
             {
-                auto mask = (other >> (8 * sizeof(T) - 1));
+                auto other_is_negative = other < batch<T, A>(T(0));
```
serge-sans-paille, line 1731: This is much more idiomatic

include/xsimd/arch/xsimd_avx512f.hpp
```diff
@@ -1561,7 +1561,7 @@ namespace xsimd
 
                     const uint32_t magic = (0x80 + 0x4000 + 0x200000 + 0x10000000);
 
-                    unsigned char res = ((data * magic) >> 24) & 0xFF;
+                    unsigned char res = (data * magic) >> 28;
```
serge-sans-paille, line 1564: I feel ashamed. But also very proud of that trick ;-)

test/CMakeLists.txt
```diff
@@ -218,7 +218,10 @@ if (XSIMD_ENABLE_WERROR)
 endif()
 
 add_subdirectory(doc)
-add_subdirectory(architectures)
+OPTION(BUILD_ARCHITECTURE_TESTS "xsimd compilation tests for various -march targets" OFF)
```
serge-sans-paille: Why didn't you turn this ON by default?

serge-sans-paille (approved):
A few questions, but nice work, thanks a lot!

test/CMakeLists.txt
AntoinePrv: We can turn this on by default. The current issue is that it does not play well with `XSIMD_DEFAULT_ARCH` (needed to build a`vx512_128` etc).
I plan to submit another PR for making it easier to select the architectures so perhaps I'll find a better way there.
Making it the default in the meantime

include/xsimd/arch/xsimd_avx512f.hpp
AntoinePrv: I tend to write all magic numbers as result of `constexpr` functions. Mostly because I cannot reverse engineer other people magic numbers 😅

PR #1413, opened 2026-09-15
CMake architecture selection utility

AntoinePrv:
Add CMake file with `xsimd_target_set_arch` to easily select compile to the matching xsimd achitecture.
It's a free function that can be reused in downstream projects, reducing CMake boilerplate.

It's also a replacement to #1348, `CMakePresets.json` being unflexible and suffering from combinatorial explosion.
eg. `-DTARGET_ARCH=avx-128` is enough to make sure we don't build on a higher architecture locally, and to set the default architecture the smaller size avx-128.


JohanMabille (approved):

serge-sans-paille (approved):
Thanks a lot for the hard work. this is a big step forward in making our CI and cmake files easier to maintain.

Please squash all commits and we're good!

